PR #1, opened 2022-01-20
Alpha 1: We have an Ingress!

mmanciop:
Early-stage MVP. No unit tests or integration tests. Manually tested with scaling up and down the Traefik Juju application and multiple upstreams (using the prometheus-k8s charm as built from [this PR](https://github.com/canonical/prometheus-k8s-operator/pull/190) that supports the `ingress` relation as supported by this PR and the [istio-pilot PR](https://github.com/canonical/istio-operators/pull/43)).

Supports having Traefik monitored by the [prometheus-k8s](https://charmhub.io/prometheus-k8s) charm, because of course it does ;-)


src/charm.py
```diff
@@ -106,8 +106,9 @@ def _on_traefik_pebble_ready(self, event):
                     }
                 },
                 # We always start the Prometheus endpoint for simplicity
-                # TODO: Generate this file in the dynamic configuration folder when the metrics-endpoint
-                # relation is established?
+                # TODO: Generate th
+                # is file in the dynamic configuration folder when the
```
jnsgruk, line 110:
```suggestion
                # TODO: Generate this file in the dynamic configuration folder when the
```

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+from ops.pebble import PathError
 
 logger = logging.getLogger(__name__)
 
 
-class OperatorTemplateCharm(CharmBase):
-    """Charm the service."""
+_TRAEFIK_CONTAINER_NAME = "traefik"
+_TRAEFIK_LAYER_NAME = "traefik"
+_TRAEFIK_SERVICE_NAME = "traefik"
```
jnsgruk, line 26: Is there any reason not to coalesce these, at least while they're all the same?

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+        container = self.unit.get_container(_TRAEFIK_CONTAINER_NAME)
+
+        # TODO Handle case the relation events are deferred and this hook has not
+        # run yet when the pebble event queueing triggers the previously deferred
+        # events.
+        if not container.can_connect():
+            self.unit.status = WaitingStatus(
+                f"container '{_TRAEFIK_CONTAINER_NAME}' not yet ready"
+            )
+            return
```
jnsgruk, line 77: `event.defer()` before this? To be honest, you can probably drop this check in general -- this event is fired precisely *because* Juju can connect to it... leaving this in isn't *wrong* mind you

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+
+        self.framework.observe(self.on["ingress"].relation_created, self._handle_ingress_change)
+        self.framework.observe(self.on["ingress"].relation_joined, self._handle_ingress_change)
+        self.framework.observe(self.on["ingress"].relation_changed, self._handle_ingress_change)
+        self.framework.observe(self.on["ingress"].relation_departed, self._handle_ingress_change)
+        self.framework.observe(self.on["ingress"].relation_broken, self._handle_ingress_broken)
+
+    def _on_traefik_pebble_ready(self, event):
+        # Ensure the required basic configurations and folders exist
+        container = self.unit.get_container(_TRAEFIK_CONTAINER_NAME)
```
jnsgruk, line 68:
```suggestion
        container = event.workload
```

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+        # routes.
+        for ingress_relation_configuration_file in container.list_files(
+            path=_TRAEFIK_INGRESS_CONFIGURATIONS_DIRECTORY,
+            pattern="juju_*.yaml"
+        ):
+            container.remove_path(ingress_relation_configuration_file.path)
+            logger.debug(f"Deleted orphaned ingress configuration file: {ingress_relation_configuration_file.path}")
+
+        # TODO Disable static config BS with telemetry and check new version
+        basic_configurations = yaml.dump(
```
jnsgruk, line 95:
```suggestion
        static_configuration = yaml.dump(
```

Might be nice to rename this so it's consistent with Traefik's docs

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+
+            supported_versions = yaml.safe_load(supported_versions_raw)
+
+            if "v3" not in supported_versions:
+                logger.error(
+                    f"The {other_app_name} application does not support the ingress relation v3 "
+                    f"(found: '{supported_versions}'); aborting data negotiation"
+                )
+                # TODO Put the charm in Blocked status?
+                return
```
jnsgruk, line 173: What happens with all these early returns? Feels like there should be some sensible return value resulting in a status being set?

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+            ingress_relation_configuration["http"]["services"][traefik_service_name] = {
+                "loadBalancer": {"servers": [{"url": f"http://{unit_ingress_address}:{port}"}]}
+            }
+
+            unit_urls[unit.name] = f"http://{gateway_address}:{self._port}/{route_prefix}"
+
+        ingress_relation_configuration_path = f"{_TRAEFIK_INGRESS_CONFIGURATIONS_DIRECTORY}/{self._ingress_config_file_name(relation)}"
+        container.push(ingress_relation_configuration_path, yaml.dump(ingress_relation_configuration))
+
+        logger.debug(f"Updated ingress configuration file: {ingress_relation_configuration_path}")
```
jnsgruk, line 280: Nite: best not to build logger strings (applies elsewhere)

```suggestion
        logger.debug("Updated ingress configuration file: '%s'", ingress_relation_configuration_path)
```

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
+
+    def _handle_ingress_broken(self, event: RelationEvent):
+        if not isinstance(self.unit.status, ActiveStatus):
+            logger.debug("Charm not active yet, deferring event")
+            event.defer()
+            return
+
+        relation = event.relation
+
+        container = self.unit.get_container(_TRAEFIK_CONTAINER_NAME)
```
jnsgruk, line 296: This line features a lot throughout this code, you could consider doing

```
self._container = self.unit.get_container(_TRAEFIK_CONTAINER_NAME)
```

In the `__init__`

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
 
-    def _on_config_changed(self, _):
-        """Just an example to show how to deal with changed configuration.
+        current_pebble_layer = container.get_plan().to_dict()
 
-        TEMPLATE-TODO: change this example to suit your needs.
-        If you don't need to handle config, you can remove this method,
-        the hook created in __init__.py for it, the corresponding test,
-        and the config.py file.
+        if current_pebble_layer != updated_pebble_layer:
```
jnsgruk, line 387:
```suggestion
        if current_pebble_layer["services"] != updated_pebble_layer["services"]:
```
?

src/charm.py
```diff
@@ -4,101 +4,424 @@
...
-        """
-        fail = event.params["fail"]
-        if fail:
-            event.fail(fail)
+            self.unit.status = MaintenanceStatus(f"starting the '{_TRAEFIK_SERVICE_NAME}' service")
+
+            container.start(_TRAEFIK_SERVICE_NAME)
+
+            if is_restart:
+                logger.info(f"'{_TRAEFIK_SERVICE_NAME}' service restarted")
```
jnsgruk, line 413: Don't build logger strings :)

tests/unit/test_charm.py
```diff
@@ -4,64 +4,14 @@
...
 
-from charm import OperatorTemplateCharm
+from charm import TraefikIngressCharm
 
 
-class TestCharm(unittest.TestCase):
+class TestTraefikIngressCharm(unittest.TestCase):
     def setUp(self):
-        self.harness = Harness(OperatorTemplateCharm)
+        self.harness = Harness(TraefikIngressCharm)
```
jnsgruk, line 15: I can only assume there is more to come here :sweat_smile:

tests/unit/test_kubernetes_service.py
```diff
@@ -0,0 +1,325 @@
+# Copyright 2021 Canonical Ltd.
```
jnsgruk, line 1: I'm not sure I'd include this test file -- it's in the upstream source for the charm that hosts the library, and won't be updated automatically with `charmcraft fetch-lib`

jnsgruk (changes requested):
Nice work overall. Some minor comments

simskij:
> supports the ingress relation 

I was under the impression we decided to use a separate interface and relation name for this.
However, I might very well have mixed things up.

mmanciop:
> > supports the ingress relation
> 
> I was under the impression we decided to use a separate interface and relation name for this. However, I might very well have mixed things up.

Yes, we eventually will. But I would like that to be properly specc’d, which was also agreed, and until then, no new relation interface :-)

Also, I did not want to have also to change the matching Prometheus PR with so many things in flight.

simskij:
Fair enough, good thinking!

metadata.yaml
```diff
@@ -3,23 +3,40 @@
...
-# TEMPLATE-TODO: each container defined above must specify an oci-image resource
 resources:
-  httpbin-image:
+  traefik-image:
     type: oci-image
-    description: OCI image for httpbin (kennethreitz/httpbin)
+    description: OCI image for traefik (https://hub.docker.com/_/traefik/)
     # Included for simplicity in integration tests
-    upstream-source: kennethreitz/httpbin
+    upstream-source: docker.io/library/traefik:latest
```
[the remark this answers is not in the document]
mmanciop: Commented out in b37fbfecb379a15e0f272f6af2f24f1e8cfe5671

requirements.txt
```diff
@@ -1 +1,4 @@
 ops >= 1.2.0
```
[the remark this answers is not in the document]
mmanciop: There isn't really anything we need in 1.3.0 atm, and we are not pinning the `ops` version. Why do you find this important @jnsgruk ?

src/charm.py
mmanciop: Done in 3cff0be

src/charm.py
mmanciop: mmm, now that you mention it @jnsgruk , shouldn't we compare `current_pebble_layer["services"].to_dict()` and `updated_pebble_layer["services"].to_dict()` instead?

src/charm.py
mmanciop: Please ;-) DRY here trumps a negligible amount of performance.

tests/unit/test_charm.py
mmanciop: I never said I would provide tests in alpha ;-)
The TODO promises a spec of which integration tests are needed.

tests/unit/test_kubernetes_service.py
mmanciop: Done in aa98606

CONTRIBUTING.md
```diff
@@ -58,6 +74,6 @@ juju add-model dev
 # Enable DEBUG logging
 juju model-config logging-config="<root>=INFO;unit=DEBUG"
 # Deploy the charm
-juju deploy ./template-operator_ubuntu-20.04-amd64.charm \
-    --resource httpbin-image=kennethreitz/httpbin \
+juju deploy ./traefik-k8s-operator_ubuntu-20.04-amd64.charm traefik-ingress \
+    --resource traefik-image=docker.io/library/traefik:latest \
```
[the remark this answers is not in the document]
mmanciop: Done in b385f53f3140033be5f273cc5b276531857f050d

tests/unit/test_charm.py
mmanciop: Specifically, I am not writing brittle unit tests when we know we will change the relation interface and likely library significant with @johnsca work ongoing to stabilize the `istio-pilot` lib.

requirements.txt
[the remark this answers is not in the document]
jnsgruk: Not urgent, just that `container.can_connect` and some of the files API fixes are in `1.3.0`

src/charm.py
jnsgruk: Erm, you should be good I think

mmanciop: Aaaactually, we do not have any variability in terms of CLI args. I think we are good to reduce a lot of complexity here. Brb.

mmanciop: Done in 47c1ef3ff371d561c59c2ae5accda21efac1deb5

PR #2, opened 2022-01-28
Replace ingress relation with ingress-unit

johnsca:
Brings the `ingress-unit` relation into the existing PR to replace the `ingress` relation. Tested against a modified version of the prometheus-k8s PR.


johnsca:
Closing in favor of #3 since #1 was merged.

PR #4, opened 2022-02-15
Add GitHub workflows for PR verification and edge-publication to Charmhub

mmanciop:


mmanciop:
Before publishing this PR we need to sort out #5

PR #8, opened 2022-02-15
Introduce a build argument in the Dockerfile for Traefik's version

mmanciop:
Introduce the `traefik_version` build argument, with default set to `2.5.6`, to parametrize which Traefik release from GitHub is downloaded over the container image build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
 FROM ubuntu:20.04 AS download
 
+ARG traefik_version="2.5.6"
+
 RUN apt-get update; apt-get install -y curl ca-certificates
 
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):
Not wild about the gratuitous line breaks in the Dockerfile, but this change LGTM 😉

mmanciop:
> Not wild about the gratuitous line breaks in the Dockerfile, but this change LGTM 😉

I quite literally do not know what you are talking about :-)

mmanciop:
We are instead going to use the image from https://github.com/jnsgruk/traefik-oci-image

PR #12, opened 2022-02-18
Implement routing_mode option

mmanciop:
Introduce a `routing_mode` charm option, structured as an enumeration, that allows you to select how Traefik will generate routes:

* `path`: Traefik will use its externally-visible url and create a route for the requester that will be structure like:

   `<external_url>/<requester_model_name>-<requester_application_name>-<requester-unit-index>`

   For example, an ingress-per-unit provider with `http://foo` external URL, will provide to the unit `my-unit/2` in the `my-model` model the following URL:

   `http://foo/my-model-my-unit-2`

* `subdomain`: Traefik will use its externally-visible url, and create a route for the requester that will be structure like:

  `<protocol>://<requester_model_name>-<requester_application_name>-<requester-unit-index>.<external_hostname>:<port>/`

  For example, an ingress-per-unit provider with `http://foo:8080` external URL, will provide to the unit `my-unit/2` in the `my-model` model the following URL:

  `http://my-model-my-unit-2.foo:8080`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,28 @@ microk8s enable metallb 192.168.0.10-192.168.0.100 # You likely wanna change the
 juju deploy ./traefik-k8s_ubuntu-20.04-amd64.charm traefik-ingress --trust --resource traefik-image=docker.io/jnsgruk/traefik:2.6.1
 ```
 
+## Configurations
+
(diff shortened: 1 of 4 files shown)


jnsgruk (approved):
LGTM. I don't really have any comments at this stage 😊

Would be good to get some integration tests in at some point but alll in good time

mmanciop:
I am very much looking forward to integration tests when https://github.com/canonical/prometheus-k8s-operator/pull/231 lands

PR #14, opened 2022-02-22
Support a per-application ingress relation

### LINKED ISSUE #9: Implement ingress relation interface
Besides the support for proxying in a load-balanced fashion the K8s service of a Juju application, rather than single units.

mmanciop:
This PR adds support for a new `ingress` library, which works like the existing `ingress_per_unit` library, but rather that exposing single units of the proxied related application, it expects to be given a host that performs load-balancing across (e.g., the hostname of a local K8s service).

Fixes #9


mmanciop:
@johnsca I think I have having troubles with SDI, halp!

jnsgruk:
@mmanciop some refactoring done, and now pointing at the SDI branch with my proposed fix, which I think leaves the "constructor only" initialisation in tact, and the unit tests passing. We still have some verification to do though, for sure.

Issue #24, opened 2022-03-08
Remove the `serialized_data_interface` dependency


jnsgruk:
At present, this charm depends on the [`serialized-data-interface`](https://github.com/canonical/serialized-data-interface) (SDI) library to perform version negotiation and schema validation on relation data.

While this has some attractive properties, the implementation is such that it obscures the way that relations are actually formed, and how charms really communicate through the use of 'relation data bags'. Concepts such as `requests` and `responses` do not really exist in the context of relations, and including such language will be confusing to the ecosystem.

Juju will soon include the native ability to perform some of the function that SDI aims to provide, and this charm is likely to proliferate beyond just internal Canonical products, and as such we should aim to implement what we need using constructs natively available in `ops`.

My suggestion is that we fix `ingress-per-unit` first, and then `ingress` - in line with the Observability team's current priorities.

jnsgruk:
Key points for discussion:

- Version checking or not? We can still validate relation data bag contents and surface reasonable errors without
  - If possible, I'd like to move away from explicit versioning. It is not in the Juju core, and it is likely to cause us issues later
- Validation of relation *role* and relation *limits*
- Validation of schema for relation data bag contents

mmanciop:
There is also the relation `meta` validation: relation interface, relation role and relation limit. See https://github.com/canonical/prometheus-k8s-operator/blob/98ec8278fda13bf23091ffab18d00b23ebceeada/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L372 and related code for an example of we perform this validation in other charm libs.

mmanciop:
About "Version checking or not" oh yes! Each of these issues, on their own, effectively warrant a new version to work properly:

* #6
* #7
* #18

PR #28, opened 2022-03-14
remove SDI dependency

PietroPasotti:
This is a work in progress.
I created a new v1/ingress_per_unit provider that does not use SDI.
Next: work on the requirer and prettify the code.

Fixes https://github.com/canonical/traefik-k8s-operator/issues/24


tests/unit/test_lib_per_unit_provides.py
```diff
@@ -4,7 +4,7 @@
 from textwrap import dedent
 from unittest.mock import Mock
 
-from charms.traefik_k8s.v0.ingress_per_unit import IngressPerUnitProvider
+from charms.traefik_k8s.v1.ingress_per_unit import IngressPerUnitProvider
```
mmanciop, line 7: `v1` is intended to mark a library stable. I would rather stick to `v0` for now if we have sufficient backwards compat.

mmanciop (changes requested):

tests/unit/test_lib_per_unit_provides.py
PietroPasotti: sure it's just my development setup, so I can have a charm using v0 requirer talk to a charm using v1 provider and check that they work :)

mmanciop: Nice :-)

PietroPasotti:
@jnsgruk @mmanciop initial surgery is hereby done.
I verified that the charm works with prometheus, more testing with more units would be desirable but prometheus-k8s bugs out on scaling...


TBD:
what to do with version negotiation?
For now I mocked the version data. Both sides share a hardcoded ['v1'] for versions, and the negotiation is marked as succeeded if both sides have shared some data (whatever). The schema-based validation is still in place. The data format and encoding is unchanged.

I encountered some issues when fixing the tests. The tests now pass, but I'm not sure if I should have fixed my code instead.
Working now to ensure that the charm unittests also pass.

## Issue 1:
image
At line 46, the old tests assert that the provider is available. 
In my code, the provider is however at that stage not available, and it cannot be since the provider doesn't have leadership, so it hasn't had the chance to share its version data and the handshake cannot have been completed yet. So I think that the test was wrong, and that now it's right.

The old tests only set leadership at line 54, after which provider and requirer both flip to available as the requirer is now able to share its version and the handshake is done.

To make more clear what's happening I split the one big test into several smaller tests with function-bound fixtures. 


## Issue 2:

The requires-side tests check that at each stage the number of ingress-changed events fired on the requirer increases by a certain amount; however this number does not increase (at all) in my tests. For now I fixed the tests by setting the number to what I actually get, but I am still looking into this issue.

jnsgruk:
Glad to see this progressing! Quick work! :smile: 

On versioning - my view is to *remove it for now*. We're already validating the schema, and can surface errors appropriately if either side is not adhering to that contract. Any versioning we do is sort of artificial at the moment

My observation about the tests is that we may be able to remove a lot of the `MockIUPProvider` style classes - I don't recall how much value they add without SDI - but I'm happy to be corrected. Will try to do a more thorough review later.

PietroPasotti:
> My observation about the tests is that we may be able to remove a lot of the `MockIUPProvider` style classes - I don't recall how much value they add without SDI - but I'm happy to be corrected. Will try to do a more thorough review later.

I find them pretty useful because of the remote_context wrapping they do: they allow to simulate pretty nicely the back-and-forth of both sides of the relation. I guess this makes this kind of tests integration-ish.

mmanciop:
I think I broke this as follows:

* Deployed `traefik-k8s` into a model where there was already `cose-lite` from `edge` (see Juju status below)
* `juju relate traefik prometheus:ingress` while Traefik was in `Waiting` state (as it has no `external_hostname` nor an ingress-ip through metallb)
* Got very confused as `prometheus` never got the ingress URL :D

Killing the `prometheus-0` pod did not help. I am under the impression that the `IngressPerUnitProvider` should have listened to `relation-joined`, but it isn't. Logs [here](https://github.com/canonical/traefik-k8s-operator/files/8342870/BORKED.log).

PietroPasotti:
@mmanciop have you by any chance tried to reproduce with v0.ingress-per-unit? Because the SDI EndpointWrapper is also not listening to relation-joined.

mmanciop:
> @mmanciop have you by any chance tried to reproduce with v0.ingress-per-unit? Because the SDI EndpointWrapper is also not listening to relation-joined.

No, I have not tried. But when we were debugging, we agreed that it did not matter whether the current version has this bug or not because we wouldn't fix it, given that it is deep into SDI.

PietroPasotti:
Something very fishy is going on. 
juju show-unit sees traefik-k8s application-data (version)
image

but prometheus is unable to read it
image
and the relation fails with UnversionedRelation error.

PietroPasotti:
> Something very fishy is going on. juju show-unit sees traefik-k8s application-data (version) image
> 
> but prometheus is unable to read it image and the relation fails with UnversionedRelation error.

Reason for this, turns out, is that 4-day weeks should become a standard so I won't have to think on Fridays. The new IPU version stripped the _supported_versions tag which was precondition to the negotiation first, and the data sharing second. So the relation failed on the requirer side. Re-packing prometheus with the new IPU version solved this issue.

mmanciop:
> > Something very fishy is going on. juju show-unit sees traefik-k8s application-data (version) image
> > but prometheus is unable to read it image and the relation fails with UnversionedRelation error.
> 
> Reason for this, turns out, is that 4-day weeks should become a standard so I won't have to think on Fridays. The new IPU version stripped the _supported_versions tag which was precondition to the negotiation first, and the data sharing second. So the relation failed on the requirer side. Re-packing prometheus with the new IPU version solved this issue.

This type of issues is _precisely_ why I was so adamant that we should not get rid of the `_supportedVersions` stuff. It leaves requires of older versions in an hopelessly borked, livelocked state. I do find it hardly acceptable to have to merge to a bunch of other charms a version of the library that is incompatible with `traefik-k8s` on `edge` in order to make them compatible with this PR.

My take is that the best way to go about it is to change this PR to still set the `_supportedVersions` (and thus add it to the scheme), make a new `edge` release of `traefik-k8s` that works with the `prometheus-k8s` in the wild (and other charms currently WIP for ingress integration in COS!), make the overall thing work, then update those charms to newer ingress libs.

PR #29, opened 2022-03-15
Add stripPrefix middleware for path based routing

rbarry82:
Some applications (notably Prometheus in this case) may not
correctly resolve endpoints if the prefix is passed through to
the application, as the "expected" endpoint per-unit is simply
`/` (for Prometheus `web.external-url`, including a prefix will
result in the Prometheus webroot itself also being at that URL).

Some flexibility will be required here, since other charms
(Grafana in particular) expect to see at least part of the prefix
if that is where they are hosted, or they will redirect users in
an infinite loop.

This is a first pass to unbreak Prometheus, and we can options from
there (whether it is dropping `web.external-url` from Prometheus,
keeping it as a constructor argument in Traefik, or other)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -294,6 +294,16 @@ class TraefikIngressCharm(CharmBase):
                 return None
             return gateway_address
 
+    def _generate_middleware_config(self, prefix: str) -> dict:
+        """Generate a stripPrefix middleware for path based routing."""
+        return {
+            "middlewares": {
+                "juju-sidecar-noprefix": {
+                    "stripPrefix": {"prefixes": [f"/{prefix}"], "forceSlash": False}
+                }
+            }
+        }
+
     def _generate_per_unit_config(self, request, gateway_address, unit) -> Tuple[dict, str]:
         """Generate a config dict for a given unit for IngressPerUnit."""
         config = {"http": {"routers": {}, "services": {}}}
@@ -315,6 +325,11 @@ class TraefikIngressCharm(CharmBase):
             "service": traefik_service_name,
             "entryPoints": ["web"],
         }
+
+        if self._routing_mode == _RoutingMode.path:
+            config["http"]["routers"][traefik_router_name].update(
+                self._generate_middleware_config(prefix)
+            )
 
         config["http"]["services"][traefik_service_name] = {
             "loadBalancer": {
@@ -355,6 +370,11 @@ class TraefikIngressCharm(CharmBase):
                 },
             }
         }
+
+        if self._routing_mode == _RoutingMode.path:
+            config["http"]["routers"][traefik_router_name].update(
+                self._generate_middleware_config(prefix)
+            )
 
         return config, app_url
 
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -55,6 +55,14 @@ class TestTraefikIngressCharm(unittest.TestCase):
                         "entryPoints": ["web"],
                         "rule": "PathPrefix(`/test-model-ingress-per-unit-remote-0`)",
                         "service": "juju-test-model-ingress-per-unit-remote-0-service",
+                        "middlewares": {
+                            "juju-sidecar-noprefix": {
+                                "stripPrefix": {
+                                    "prefixes": ["/test-model-ingress-per-unit-remote-0"],
+                                    "forceSlash": False,
+                                }
+                            }
+                        },
                     }
                 },
                 "services": {
@@ -101,6 +109,14 @@ class TestTraefikIngressCharm(unittest.TestCase):
                         "entryPoints": ["web"],
                         "rule": "PathPrefix(`/test-model-ingress-remote`)",
                         "service": "juju-test-model-ingress-remote-service",
+                        "middlewares": {
+                            "juju-sidecar-noprefix": {
+                                "stripPrefix": {
+                                    "prefixes": ["/test-model-ingress-remote"],
+                                    "forceSlash": False,
+                                }
+                            }
+                        },
                     }
                 },
                 "services": {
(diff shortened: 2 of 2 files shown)


mmanciop:
Actually, this is more an issue with the `Grafana Source` library not supporting the specification of a subpath, but rather only the `port`. I am generally very, very skeptical of path rewriting in proxies, as it leads inevitably to baffling UI bugs. For now, -2.

rbarry82:
Why are we talking about Grafana? This doesn't have anything to do with Grafana in any way.

This was specifically because the integration tests for the `cos-lite-bundle` were checking whether Prometheus was available on a subpath through ingress -- they were -- then following up by trying to connect to `http://cluster-local-prom-ip:9090/` with no subpath.

Using `web.external-url` with Prometheus also pushes Prometheus into configuring subpath routing on its embedded webserver, so the web root is no longer `/`, but `/some-sub-path`, so every application which communicates with it (including in-namespace/model charms) must be aware of this, and each unit will have a different endpoint.

That's _ok_. We agreed that this was acceptable after this PR. Let's not throw the baby out with the bathwater, though. Personal opinions about path rewriting/stripping proxies aside (and it's easy enough to point to applications which send directs which are not aware of proxy subpathing, so `rewrite-url` in nginx, `stripPrefix` in Traefik, or other options are basically required), it's worth having a discussion about whether or not we want to support this as an _option_ in the Traefik charm, even if we don't need it to pass Prometheus integration tests (because we rewrote them).

mmanciop:
> Why are we talking about Grafana? This doesn't have anything to do with Grafana in any way.

Because with Traefik in `rounting_mode = path`, which is the _default_, the URL of the Prometheus API changes from

```
http://<pod_ip>:9090/
```

to:

```
http://<pod_ip>:9090/<model_name>-<unit_name.replace('/','-')>
```

(Nevermind the hostname, the pod ip of Prometheus works too within the cluster ofc.)

This interplay between `ingress_per_unit` and `grafana_source` is literally the reason for [this discussion](https://chat.charmhub.io/charmhub/pl/jqjbsttg9idrmjua9p4zi43o8e).

> This was specifically because the integration tests for the `cos-lite-bundle` were checking whether Prometheus was available on a subpath through ingress -- they were -- then following up by trying to connect to `http://cluster-local-prom-ip:9090/` with no subpath.

> Using `web.external-url` with Prometheus also pushes Prometheus into configuring subpath routing on its embedded webserver, so the web root is no longer `/`, but `/some-sub-path`, so every application which communicates with it (including in-namespace/model charms) must be aware of this, and each unit will have a different endpoint.

> That's _ok_. We agreed that this was acceptable after this PR. Let's not throw the baby out with the bathwater, though. Personal opinions about path rewriting/stripping proxies aside (and it's easy enough to point to applications which send directs which are not aware of proxy subpathing, so `rewrite-url` in nginx, `stripPrefix` in Traefik, or other options are basically required), it's worth having a discussion about whether or not we want to support this as an _option_ in the Traefik charm, even if we don't need it to pass Prometheus integration tests (because we rewrote them).

> It may lead to baffling UI bugs in some cases, but it may be required in others. `rewriite-path`, `stripPrefix`, and friends do not exist for no reason. It may equally lead to baffling UI bugs in other cases where it is _not_ used, because some application will keep sending redirects for logins (or others) back to `http://example.com/login` when you hit `http://example.com/juju-foo-charm/`, leading to endless redirect loops. Some applications support flags which fix this behavior. Not all.

I do not dispute the general value of a `stripPrefix`, but:

1. I am adamant we must avoid trying to reinvent the Traefik DSL, but through Juju config options on `traefik-k8s`. The "just another routing behavior switch" is the most slippery of slopes.
2. I am arguing that this should not be the default behavior of the `rounting_mode = path` because it will _break_ applications.

I am open to making it the opt-in behavior for SOME applications. . Given the fact that today Juju does not have means of configuring single relations (or relation endpoints in this case), it is something that requires a not-yet-implemented _configuration charm_, using the same pattern we adopted for [prometheus-scrape-config](https://github.com/canonical/prometheus-scrape-config-k8s-operator). A braindump about it is provided [here](https://docs.google.com/document/d/1FceQEs4eJdPIGjnSKrQ82OfkNxAbEe9GeoenphsONLc/edit).

rbarry82:
Sure, but no.

Yes, `grafana_source` needs changes to account for that interplay, but that was not the reason for _this patch_. Nothing about `grafana_source` is relevant to this patch at all in its current state, and it is, simply, not a good reason to -2 a patch which isn't related. The reason for this patch was the failing integration tests which were attempting to contact Prometheus at `/`, not `/some-unit-prefix`, and the lack of changes to the Prometheus URI in an integration test including ingress (and `web.external_url`) were an indicator that Prometheus _also_ listening on a subpath with a non-ingress IP/route were either not intentional or not remembered.

In particular, "relating Prometheus to ingress also requires changes to `remote_write` and `grafana_source` and any other library which expects to be able to 'find' the Prometheus API" certainly did not seem intentional. The state of `grafana_source` is more or less unrelated to this patch, since it will adapt to meet one scheme or another, though.

Put very simply, this patch was "your integration tests are broken, if this was the reason why, this patch is a fix". The last paragraph of the commit message is pretty clear about this needing further discussion, and I am not advocating for reinventing the Traefik DSL _or_ making it the default. It was  "first pass", as mentioned.

Whether it's opt-in via constructor arguments or some other mechanism isn't important to me, or even that it's supported right now. I didn't mark it with `[DRAFT]` because the commit message seemed clear enough. If we _do_ want to support it either now or eventually, without re-inventing the Traefik DSL in Juju config, the unit tests here and general logic provide a starting point.

We can defer this for now or reject it forever, but reject it on the basis of "we don't want `traefik-k8s` to support this" or "we want a different mechanism to configure this" or "we want to make a different configuration builder for traefik", not "this is an issue with `grafana_source`", since that doesn't have any relation here, and correlation is not causation.

PR #35, opened 2022-04-04
Remove SDI dependency from ingress_per_unit

PietroPasotti:
Removed SDI dependency from v0/ingress_per_unit Provider/Requirer

Fixes https://github.com/canonical/traefik-k8s-operator/issues/24

### DIFF
--- a/lib/charms/traefik_k8s/v0/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_unit.py
@@ -14,9 +14,7 @@ To get started using the library, you just need to fetch the library using `char
 charm's `requirements.txt`.**
 
 ```shell
-cd some-charm
 charmcraft fetch-lib charms.traefik_k8s.v0.ingress_per_unit
-echo -e "serialized_data_interface\n" >> requirements.txt
 ```
 
 ```yaml
@@ -48,34 +46,23 @@ class SomeCharm(CharmBase):
         logger.info("This unit's ingress URL: %s", self.ingress_per_unit.url)
 ```
 """
-
 import logging
-from typing import Optional
+import warnings
+from typing import Dict, Optional, Tuple, TypeVar, Union
 
-from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent, RelationRole
-from ops.framework import EventSource
-from ops.model import Relation, Unit
-
-try:
-    from serialized_data_interface import EndpointWrapper
-    from serialized_data_interface.errors import RelationDataError
-    from serialized_data_interface.events import EndpointWrapperEvents
-except ImportError:
-    import os
-
-    library_name = os.path.basename(__file__)
-    raise ModuleNotFoundError(
-        "To use the '{}' library, you must include "
-        "the '{}' package in your dependencies".format(library_name, "serialized_data_interface")
-    ) from None  # Suppress original ImportError
-
-try:
-    # introduced in 3.9
-    from functools import cache  # type: ignore
-except ImportError:
-    from functools import lru_cache
-
-    cache = lru_cache(maxsize=None)
+import jsonschema
+import yaml
+from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent
+from ops.framework import EventSource, Object, ObjectEvents
+from ops.model import (
+    ActiveStatus,
+    Application,
+    BlockedStatus,
+    Relation,
+    StatusBase,
+    Unit,
+    WaitingStatus,
+)
 
 # The unique Charmhub library identifier, never change it
 LIBID = "7ef06111da2945ed84f4f5d4eb5b353a"  # can't register a library until the charm is in the store 9_9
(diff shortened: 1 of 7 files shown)


lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+        ingress_data: dict = {}
+        # we start by looking at the provider's app databag
+        if this_unit.is_leader():
+            # only leaders can read their app's data
+            data = relation.data[this_app].get("data")
+            deserialized = {}
+            if data:
+                deserialized = _deserialize_data(data)
+                if validate:
+                    _validate_data(deserialized, INGRESS_PROVIDES_APP_SCHEMA)
```
sed-i, line 460: Who is responsible for catching the exception? If not here then should probably add a `Raises:` statement to the public `get_request` docstring.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+            ingress_data[this_app] = deserialized
+        else:
+            # non-leader units cannot read/write the app databag
+            ingress_data[this_app] = {}
+
+        # then look at the requirer's (thus remote) unit databags
+        remote_units = [
+            obj for obj in relation.data if isinstance(obj, Unit) and obj is not this_unit
+        ]
```
sed-i, line 469: Consider a `@property` for remote units. Possibly even in ops..? :D

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+
+        for remote_unit in remote_units:
+            remote_data = relation.data[remote_unit].get("data")
+            remote_deserialized = {}
+            if remote_data:
+                remote_deserialized = _deserialize_data(remote_data)
+                if validate:
+                    _validate_data(remote_deserialized, INGRESS_REQUIRES_UNIT_SCHEMA)
```
sed-i, line 477: Would it make more sense to have the validation return a bool instead of raising?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+            ingress_data[remote_unit] = remote_deserialized
+
+        return ingress_data
+
+    def publish_ingress_data(
+        self, relation: Relation, data: typing.Dict[typing.Union[Unit, Application], dict]
```
sed-i, line 483: Import Dict, Union? That's what we do in o11y charms.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+            if entity not in (this_app, this_unit):
+                # check that we're not attempting to push new data to remote
+                # units or apps
+                if data[entity] != old_data.get(entity):
+                    raise RelationPermissionError(relation, this_unit)
+                continue
+
+            # validation step 2: only leaders can write app data
+            if entity is this_app and not this_unit.is_leader():
+                raise RelationPermissionError(relation, this_unit)
```
sed-i, line 503: Can this case be covered by a regular leader guard at the beginning of the function?
And why raise in that case?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -186,13 +431,102 @@ def is_failed(self, relation: Relation = None):
...
+
+                # if all is well, write the data
+                relation.data[this_app]["data"] = _serialize_data(this_app_data)
+
+            # repeat for unit
+            elif entity is this_unit:
+                this_unit_data = data[this_unit]
+                _validate_data(this_unit_data, INGRESS_PROVIDES_APP_SCHEMA)
+
+                relation.data[this_unit]["data"] = _serialize_data(this_unit_data)
```
sed-i, line 520:
```suggestion
            if entity is this_app or entity is this_unit:
                data = data[entity]
                _validate_data(data, INGRESS_PROVIDES_APP_SCHEMA)

                # if all is well, write the data
                relation.data[entity]["data"] = _serialize_data(data)
```

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -246,30 +581,71 @@ def app_name(self):
...
     def units(self):
-        """The remote units."""
-        return sorted(self._relation.units, key=lambda unit: unit.name)
+        """The remote units that have posted requests, i.e., those that have provided the name."""
+        ready_units = filter(self.is_unit_ready, self._relation.units)
+        return sorted(ready_units, key=lambda unit: unit.name)
+
+    def is_unit_ready(self, unit: Unit) -> bool:
+        """Returns whether the given unit has provided its name, model, host and port data."""
+        self._check_unit_belongs_to_relation(unit)
```
sed-i, line 592: Would it make more sense to have this return a bool instead of raising?
If not, then should probably add a `Raises:` statement.
But shouldn't `is_ready`  return a bool without raising regardless?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -278,12 +654,16 @@ def _get_data_from_first_unit(self, key: str):
...
+            return None
 
-                if key in unit_data:
-                    return unit_data[key]
+        unit_data = self._data[unit]
+
+        if key in unit_data:
+            return unit_data[key]
 
         return None
```
sed-i, line 668:
```suggestion
        return unit_data.get(key, None)
```

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -89,65 +86,161 @@ def _handle_ingress_per_unit(self, event):
...
-                "required": ["ingress"],
-            }
-        },
-    }
+# ======================= #
+#      LIBRARY GLOBS      #
+# ======================= #
+
+INTERFACE = "ingress_per_unit"
+ENDPOINT = INTERFACE.replace("_", "-")
```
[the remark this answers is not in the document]
mmanciop: Interface is interface. Endpoint here is "Relation name". @PietroPasotti we should use `DEFAULT_RELATION_NAME` as in https://github.com/canonical/prometheus-k8s-operator/blob/a369c99cfdafeceaf94b1ec3ba1d961a11bb0cc6/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py#L42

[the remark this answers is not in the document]
PietroPasotti: Umm but in this case I think the usage is appropriate. 
image

In this case 'ingress_per_unit' is the interface, and 'ingress-per-unit' is... what would you call that otherwise?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -156,25 +249,177 @@ def __init__(self, charm: CharmBase, endpoint: str = None):
...
+
+    def _handle_relation(self, event):
+        if self.is_ready(event.relation):
+            self.on.ready.emit(event.relation)
+        elif self.is_available(event.relation):
+            self.on.available.emit(event.relation)
+        elif self.is_failed(event.relation):
+            self.on.failed.emit(event.relation)
+
+    @cache
```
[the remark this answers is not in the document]
PietroPasotti: It makes it also hard(er) to test these methods, so I'm all +1 for removing it. I kept it because it was in SDI, and within that context it made sense because SDI managed its own status and could guarantee that it would never change within a charm's execution lifetime. But now that we own it, and if we think it might change in the future... Fair enough. Also it's not that big an optimization to warrant all the extra potential bugginess and mental overhead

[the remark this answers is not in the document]
PietroPasotti: @mmanciop according to the same spirit, what'd you think about removing all caching in IPU and be done with it?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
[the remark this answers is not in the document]
mmanciop: `ingress-per-unit` would be the relation name. `ingress_per_unit` is the relation interface.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: @mmanciop what was the idea of making it raise? 
I see most of the methods which use _check_unit_belongs_to_relation return Optional[Trype], so wouldn't it make more sense to do:
```
if not self._unit_belongs_to_relation(unit):
    return None
```
or do we want the user to know that there's something deeply wrong with what they're trying to do?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -14,9 +14,7 @@
 charm's `requirements.txt`.**
 
 ```shell
-cd some-charm
 charmcraft fetch-lib charms.traefik_k8s.v0.ingress_per_unit
```
jnsgruk, line 18: Unless we break the library into two (provides/requires) will charmers not need to add `jsonschema` to their `requirements.txt`?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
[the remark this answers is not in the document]
jnsgruk: +1

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -156,25 +249,177 @@ def __init__(self, charm: CharmBase, endpoint: str = None):
...
 
-    def get_request(self, relation: Relation):
-        """Get the IngressRequest for the given Relation."""
-        return IngressRequest(self, relation)
+    @cache
+    def is_available(self, relation: Relation = None):
+        """Check whether the given relation is available.
+
+        Or any relation if not specified.
+        """
```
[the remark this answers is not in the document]
jnsgruk: I would be content to drop the returns piece in the docstring and rely upon type annotations unless the return type needs further explanation

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: Good point. How does that work for libraries? Do we have a way for libs to declare their dependencies?

PietroPasotti: - How do @jnsgruk  and @mmanciop feel about this? 
image

jnsgruk: Now that's interesting. I would probably use `log.warning`, but I quite like the idea that it still has a chance of working...

jnsgruk: either way, you still need to include in the docs to echo jsonschema into requirements.txt or similar :)

Other option is just to validate on the Traefik side, but that will be less reliable.

PietroPasotti: yeah the catch here is that both the provider and the requirer have to install that dependency. Or indeed we could say that Traefik owns the correctness of the data, so the requirer is responsible for doing its own validation if it so desires; and we can provide a utility method or just the public schema for "this is what you should be expecting". Then the charm using the requirer side would have to also install that dependency.

mmanciop (approved):

Issue #39, opened 2022-04-06
Remove SDI-inherited testing infrastructure


PietroPasotti:
### Bug Description

When I made this change, that seemed perfectly innocuous:
![image](https://user-images.githubusercontent.com/6230162/161923156-ff7021b4-73dd-40e3-8a5a-d00fe3cc9ec1.png)

I bumped into an issue: `test_lib_helpers` was patching the IngressRequest object to such an extent that `_check_unit_belongs_to_relation(self.units[0])` was raising.
Bottom line is: all the mocking and patching we're doing in `test_lib_helpers` is hard to maintain and debug.

### To Reproduce

checkout purge-sdi-polish at 18c7b99b5098276e65b87186b8a57f6d2069c404
and run the utests.

### Environment

dev setup

### Relevant log output

```shell
n.a.
```

PR #42, opened 2022-04-07
Remove Request/Response mechanism in ingress-per-unit

PietroPasotti:
## Issue
The IngressRequest object is unnecessarily complicated and introduces a confusing layer of request/response mechanics that is hard to reason about.

## Solution
Get rid of that object, replace the existing request/response terminology for 'publish' on the one and the other side.

## Testing Instructions
Unittest

## NB
it depends on the purge-sdi-polish PR, base will change to main as soon as that PR is merged

## Release Notes
Simplified the ingress-per-unit api and internals.

Fixes https://github.com/canonical/traefik-k8s-operator/issues/43

### DIFF
--- a/lib/charms/traefik_k8s/v0/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_unit.py
@@ -47,9 +47,7 @@ class SomeCharm(CharmBase):
 ```
 """
 import logging
-import typing
-import warnings
-from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
+from typing import Dict, Optional, Tuple, TypeVar, Union
 
 import jsonschema
 import ops.model
@@ -129,6 +127,8 @@ except ImportError:
 
 
 class RequirerData(TypedDict):
+    """Model of the data a unit implementing the requirer will need to provide."""
+
     model: str
     name: str
     host: str
@@ -484,7 +484,11 @@ class IngressPerUnitProvider(_IngressPerUnitBase):
         try:
             _validate_data(data, INGRESS_PROVIDES_APP_SCHEMA)
         except DataValidationError as e:
-            log.error(f"unable to publish url to {unit_name}: " f"corrupted application databag")
+            log.error(
+                "unable to publish url to {}: corrupted application databag ({})".format(
+                    unit_name, e
+                )
+            )
             return
         data["ingress"][unit_name] = {"url": url}
         try:
--- a/lib/charms/traefik_k8s/v0/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_unit.py
@@ -478,7 +478,8 @@ class IngressPerUnitProvider(_IngressPerUnitBase):
         raw_data = relation.data[self.app].get("data", None)
         data = _deserialize_data(raw_data) if raw_data else {"ingress": {}}
 
-        # TODO: is this necessary?
+        # we ensure that the application databag has the shape we think it
+        # should have; to catch any inconsistencies early on.
         try:
             _validate_data(data, INGRESS_PROVIDES_APP_SCHEMA)
         except DataValidationError as e:
@@ -488,7 +489,14 @@ class IngressPerUnitProvider(_IngressPerUnitBase):
                 )
             )
             return
+
+        # we update the data with a new url
         data["ingress"][unit_name] = {"url": url}
+
+        # we validate the data **again**, to ensure that we respected the schema
+        # and did not accidentally corrupt our own databag.
+        _validate_data(data, INGRESS_PROVIDES_APP_SCHEMA)
+
         try:
             relation.data[self.app]["data"] = _serialize_data(data)
         except ops.model.RelationDataError:
--- a/lib/charms/traefik_k8s/v0/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_unit.py
@@ -47,6 +47,7 @@ class SomeCharm(CharmBase):
 ```
 """
 import logging
+import typing
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 
@@ -120,9 +121,16 @@ INGRESS_PROVIDES_APP_SCHEMA = {
 #          TYPES          #
 # ======================= #
 
+if typing.TYPE_CHECKING:
+    from typing import TypedDict
+    class RequirerData(TypedDict):
+        model: str
+        name: str
+        host: str
+        port: int
 
+RequirerUnitData = Dict[Unit, 'RequirerData']
 KeyValueMapping = Dict[str, str]
-RequirerUnitData = Dict[Unit, KeyValueMapping]
 ProviderApplicationData = Dict[str, KeyValueMapping]
 
 
@@ -443,6 +451,22 @@ class IngressPerUnitProvider(_IngressPerUnitBase):
         """Get the IngressRequest for the given Relation."""
         provider_app_data, requirer_unit_data = self._fetch_relation_data(relation)
         return IngressRequest(self, relation, provider_app_data, requirer_unit_data)
+
+    def is_unit_ready(self, relation: Relation, unit: Unit) -> bool:
+        """Whether the given unit has shared its side of the data."""
+        assert unit in relation.units, "attempting to get ready state for unit that does not belong to relation"
+        if relation.data.get(unit, {}).get('data'):
+            # TODO consider doing schema-based validation here
+            return True
+        return False
+
+    def get_data(self, relation: Relation, unit: Unit, validate:bool = False) -> RequirerData:
+        """Fetch the data shared by this unit via the relation (Requirer side)."""
+        # todo: should we validate here?
+        data = _deserialize_data(relation.data[unit]['data'])
+        if validate:
+            _validate_data(data, INGRESS_REQUIRES_UNIT_SCHEMA)
+        return data
 
     def _fetch_relation_data(
         self, relation: Relation, validate=False
@@ -805,7 +829,7 @@ class IngressPerUnitRequirer(_IngressPerUnitBase):
             data = _deserialize_data(raw)
             try:
                 _validate_data(data, INGRESS_REQUIRES_UNIT_SCHEMA)
-            except jsonschema.ValidationError:
+            except DataValidationError:
                 log.exception("Error validating relation data")
                 return True
 
--- a/lib/charms/traefik_k8s/v0/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_unit.py
@@ -636,7 +636,7 @@ class IngressPerUnitRequirer(_IngressPerUnitBase):
 
     def _publish_auto_data(self, relation: Relation):
         if self._auto_data and self.is_available(relation):
-            self._publish_ingress_data(*self._auto_data)
+            self._provide_ingress_requirements(*self._auto_data)
 
     @property
     def relation(self) -> Optional[Relation]:
@@ -698,7 +698,7 @@ class IngressPerUnitRequirer(_IngressPerUnitBase):
         # TODO Avoid spurious events, emit only when URL changes
         self.on.ingress_changed.emit(self.relation)
 
-    def _publish_ingress_data(self, host: Optional[str], port: int):
+    def _provide_ingress_requirements(self, host: Optional[str], port: int):
         """Publish the data that the provider needs to provide ingress."""
         if not host:
             binding = self.charm.model.get_binding(self.relation_name)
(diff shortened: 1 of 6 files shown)


lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -439,10 +455,58 @@ def is_failed(self, relation: Relation = None) -> bool:
 
         return False
 
-    def get_request(self, relation: Relation):
-        """Get the IngressRequest for the given Relation."""
-        provider_app_data, requirer_unit_data = self._fetch_relation_data(relation)
-        return IngressRequest(self, relation, provider_app_data, requirer_unit_data)
+    def is_unit_ready(self, relation: Relation, unit: Unit) -> bool:
+        """Whether the given unit has shared its side of the data."""
+        assert (
```
jnsgruk, line 460: Is this supposed to still be here? Seems like a strange use of `assert`, or at least a strange place to use it

jnsgruk (changes requested):
Mostly docstring nits, but the API is important :)

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -832,8 +728,8 @@ def _publish_ingress_data(self, host: Optional[str], port: int):
         }
         self.relation.data[self.unit]["data"] = _serialize_data(data)
 
-    def request(self, *, host: str = None, port: int):
-        """Request ingress to this unit.
+    def publish_ingress_data(self, *, host: str = None, port: int):
```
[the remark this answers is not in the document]
mmanciop: `update` ?

src/charm.py
```diff
@@ -277,36 +282,55 @@ def _process_ingress_relation(self, relation: Relation):
...
-        # FIXME Ideally, follower units could instead watch for the data in the
-        # ingress app data bag, but Juju does not allow non-leader units to read
-        # the application data bag on their side of the relation, so we may start
-        # routing for a remote unit before the leader unit of ingress has
-        # communicated the url.
-        if provider == self.ingress_per_app:
+        # TODO: once also IngressPerApp is SDI-free,
+        #  abstract the common logic here and remove this branch
+        if provider is self.ingress_per_app:
+            request = provider.get_request(relation)
```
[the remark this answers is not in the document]
PietroPasotti: that's the ingress_per_app branch, still need to remove SDI there

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: It's a sanity check, that condition should not occur in production (is_unit_ready is used in ways which guard against that ever happening), but I had a case during testing (because of https://github.com/canonical/traefik-k8s-operator/issues/39) where this precondition was violated. I think it makes sense to check it because the error is really hard to track down otherwise.

Will add a comment.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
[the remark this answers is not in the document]
PietroPasotti: I like `provide_ingress_requirements`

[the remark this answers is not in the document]
jnsgruk: Sure, either is fine. Let me know when this ready for another pass.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -419,10 +437,60 @@ def is_failed(self, relation: Relation = None) -> bool:
...
+        # during testing: cfr https://github.com/canonical/traefik-k8s-operator/issues/39
+        assert (
+            unit in relation.units
+        ), "attempting to get ready state for unit that does not belong to relation"
+        if relation.data.get(unit, {}).get("data"):
+            # TODO consider doing schema-based validation here
+            return True
+        return False
+
+    def get_data(self, relation: Relation, unit: Unit, validate: bool = False) -> "RequirerData":
```
jnsgruk, line 452: Perhaps for another PR, but should the validation be optional? I'm not sure what that gains us on the provider side -- I thought we spoke about wrapping the `jsonschema` import in a try/except to potentially allow charms that don't add that dependency to function, but with a warning. Perhaps here that could reflected in an automatic choice as to whether validation should occur?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -419,10 +437,60 @@ def is_failed(self, relation: Relation = None) -> bool:
...
+
+    def publish_url(self, relation: Relation, unit_name: str, url: str):
+        """Place the ingress url in the application data bag for the units on the requires side.
+
+        Assumes that this unit is leader.
+        """
+        raw_data = relation.data[self.app].get("data", None)
+        data = _deserialize_data(raw_data) if raw_data else {"ingress": {}}
+
+        # TODO: is this necessary?
```
jnsgruk, line 467: This feels like a thing to iron out sooner rather than later...

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -812,16 +712,16 @@ def _publish_ingress_data(self, host: Optional[str], port: int):
         }
         self.relation.data[self.unit]["data"] = _serialize_data(data)
 
-    def request(self, *, host: str = None, port: int):
-        """Request ingress to this unit.
+    def provide_ingress_requirements(self, *, host: str = None, port: int):
```
jnsgruk, line 715: Why do we have `_provide_ingress_requirements` and `provide_ingress_requirements`? The latter seems to just call the former?

jnsgruk (approved):
Overall this is looking much cleaner. Some nits in the comments, but can likely all be addressed in future PRs.

jnsgruk:
(I can't work out why the Github actions don't seem to be running, though??)

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: We are validating the data in the databag before adding to it; it makes sense, so I chose to keep it.
However, I thought, why should we only validate the data before, but not after? So I added an 'output' validation step, so  after touching the databag we validate again. Better to catch it here and now, before we write the data, rather than the next time we have to read from it. Felt more pythonic and debuggable.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: Good point. I uniformed the behaviour and put a global DO_VALIDATION flag depending on the presence of `jsonschema`. Should make more sense now; also it fires a nice warning if it is not installed, and will work with and without.

PietroPasotti:
> (I can't work out why the Github actions don't seem to be running, though??)

Might be because I had originally forked the repo and only later switched to a canonical/traefik-k8s-operator purge-request branch? @mmanciop sorted that out for me, not sure if that has something to do with it

PietroPasotti:
> > (I can't work out why the Github actions don't seem to be running, though??)
> 
> Might be because I had originally forked the repo and only later switched to a canonical/traefik-k8s-operator purge-request branch? @mmanciop sorted that out for me, not sure if that has something to do with it

oh wait, might be because the base isn't main

jnsgruk:
Ah yes, that could be it. We'll get that other PR merged before we close this one and that should be solved.

jnsgruk (approved):

jnsgruk (approved):
LGTM. Let's get this one in!

src/charm.py
```diff
@@ -277,36 +282,55 @@ def _process_ingress_relation(self, relation: Relation):
...
 
-        request = provider.get_request(relation)
-
-        # FIXME Ideally, follower units could instead watch for the data in the
-        # ingress app data bag, but Juju does not allow non-leader units to read
-        # the application data bag on their side of the relation, so we may start
-        # routing for a remote unit before the leader unit of ingress has
-        # communicated the url.
-        if provider == self.ingress_per_app:
+        # TODO: once also IngressPerApp is SDI-free,
```
simskij, line 285:
```suggestion
        # TODO: once IngressPerApp is also SDI-free,
```

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -76,6 +75,20 @@ def _handle_ingress_per_unit(self, event):
...
+    DO_VALIDATION = True
+except ModuleNotFoundError:
+    log.warning(
+        "The `ingress_per_unit` library needs the `jsonschema` package to be able "
+        "to do runtime data validation; without it, it will still work but validation "
+        "will be disabled. \n"
+        "It is recommended to add `jsonschema` to the 'requirements.txt' of your charm, "
+        "which will enable this feature."
+    )
+    DO_VALIDATION = False
```
simskij, line 90: Do we need to surface this to the charm dev in any way or do we feel that a warning in the debug-log is enough?

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -112,45 +125,47 @@ def _handle_ingress_per_unit(self, event):
...
 
+
 # TYPES
-KeyValueMapping = Dict[str, str]
-RequirerUnitData = Dict[Unit, KeyValueMapping]
-ProviderApplicationData = Dict[str, KeyValueMapping]
+try:
+    from typing import TypedDict
+except ImportError:
+    from typing_extensions import TypedDict  # py35 compat
```
simskij, line 133: Nice touch with the backward compat! 🎉

lib/charms/traefik_k8s/v0/ingress_per_unit.py
```diff
@@ -419,13 +421,70 @@ def is_failed(self, relation: Relation = None) -> bool:
...
+        """Report whether the given unit has shared data in its unit data bag."""
+        # sanity check: this should not occur in production, but it may happen
+        # during testing: cfr https://github.com/canonical/traefik-k8s-operator/issues/39
+        assert (
+            unit in relation.units
+        ), "attempting to get ready state for unit that does not belong to relation"
+        if relation.data.get(unit, {}).get("data"):
+            # TODO consider doing schema-based validation here
+            return True
+        return False
```
simskij, line 434:
```suggestion
        # TODO consider doing schema-based validation here
        return relation.data.get(unit, {}).get("data")
```

lib/charms/traefik_k8s/v0/ingress_per_unit.py
simskij: > The current code generator emits no code for an assert statement when optimization is requested at compile time. 

https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

So should be fine.

simskij (changes requested):
All in all this LGTM. Added some minor suggestions.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: Well it depends on how essential we think validation should be to this library. If we want it to work without it, and validation is an opt-in additional feature, then this is fine.
On the other hand I can imagine there can be a situation where 
1) person uses lib without jsonschema, ignores the warnings
2) person installs jsonschema in charm for unrelated reasons
3) charm starts breaking because validation is suddenly active and the schema is wrong/outdated
4) person is puzzled

@jnsgruk @mmanciop what is your view on this? 
I recently merged a pr from @mmanciop, his take was less permissive: install jsonschema or get an Exception.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: I purposefully left it like so because I expect more stuff to go in the is_ready method. 
Also I really like a method called `is_ready` to return `bool` :)

simskij: wrapping it in `bool()` would solve that, but I get your point.

lib/charms/traefik_k8s/v0/ingress_per_unit.py
PietroPasotti: FWIW it's still pretty obvious that there's something wrong:
image

jnsgruk: I'm happy with where it's at - the only thing we could consider is adding something into the status messages when deployed without validation so it's obvious from `juju status`. I definitely feel like we need some kind of "DegradedStatus" at some point

jnsgruk:
One thing that needs doing before merging: bumping the `LIBPATCH` so you can publish the lib! :)

simskij (approved):

PR #44, opened 2022-04-13
Traefik route endpoint

PietroPasotti:
## Feature
This PR adds an integration with the Traefik Route charm (cf. https://github.com/PietroPasotti/traefik-route-k8s-operator), enabling users to configure Traefik on a per-relation basis, overcoming the limitations of ingress-per-unit and ingress-per-app, making Traefik a lot more general.

The role of Traefik in this integration is very limited: it simply has to accept a yaml config file and write it to the folder where traefik will load it up. Therefore the changes to the charm codebase are rather limited. 

## Testing Instructions
Aside from running the utests, you can test this in combination with TraefikRoute and prometheus-k8s in a live deployment (itests pending).

```
juju deploy traefik-route-k8s --channel beta
```

## Release Notes
* Added `traefik_route_k8s/v0/traefik_route.py` lib
* Added `traefik_route` integration in the Traefik charm.


lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -0,0 +1,174 @@
...
+cd some-charm
+charmcraft fetch-lib charms.traefik_route_k8s.v0.traefik_route
+```
+
+```yaml
+requires:
+    traefik_route:
+        interface: traefik_route
+        limit: 1
+```
```
jnsgruk, line 24: As I understand it - the design was such that the workload needing ingress would only ever need to communicate on the `ingress` or `ingress_per_unit` relation. 

This interface should only be present between traefik-k8s and traefik-route-k8s, right?

lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -0,0 +1,174 @@
...
+    # ...
+    self.ingress_per_unit = TraefikRouteProvider(self)
+    self.traefik_route = TraefikRouteRequirer(self)
+
+    self.framework.observe(
+        self.ingress_per_unit.on.request, self.traefik_route.relay
+    )
+    self.framework.observe(
+        self.traefik_route.on.response, self.ingress_per_unit.respond
+    )
```
jnsgruk, line 43: As above - I think something was lost in translation here.

lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -0,0 +1,174 @@
...
+        if self.is_ready(event.relation):
+            # todo check data is valid here?
+            self.on.ready.emit(event.relation)
+
+    @staticmethod
+    def is_ready(relation: Relation) -> bool:
+        """Whether TraefikRoute is ready on this relation: i.e. the remote app shared the config."""
+        return 'config' in relation.data[relation.app]
+
+    @staticmethod
```
jnsgruk, line 127: Do these need to be static?

lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -0,0 +1,174 @@
...
+
+class TraefikRouteRequirer(Object):
+    """Wrapper for the requirer side of traefik-route.
+
+    traefik_route will publish to the application databag an object like:
+    {
+        'config': <Traefik_config>
+    }
+
+    'ingress' is provided by the ingress end-user via ingress_per_unit,
```
jnsgruk, line 142: `ingress` is not included in your example above - should it be?

src/charm.py
```diff
@@ -277,82 +288,86 @@ def _process_ingress_relation(self, relation: Relation):
...
-                    if self.unit.is_leader():
-                        request.respond(unit, unit_url)
+            self._provide_ingress_per_unit(relation)
 
-                if unit_config:
-                    always_merger.merge(config, unit_config)
+    def _provide_routed_ingress(self, relation: Relation):
+        """Provide ingress to a unit related through TraefikRoute."""
+        if not self.traefik_route.is_ready(relation):
+            logger.info(f"TR not ready on {relation}")
```
jnsgruk, line 307:
```suggestion
            logger.info("%s not ready on %s", self.app.name, relation)
```

jnsgruk (changes requested):

lib/charms/traefik_route_k8s/v0/traefik_route.py
PietroPasotti: No, but my pycharm linter complains if they aren't. Do they need to not be static?

lib/charms/traefik_route_k8s/v0/traefik_route.py
PietroPasotti: Uh yes, copy-pasted from the IPU library a bit too freely. On it

lib/charms/traefik_route_k8s/v0/traefik_route.py
jnsgruk: Just seems odd to me, but I guess it's not a huge problem.

jnsgruk (approved):

PR #46, opened 2022-04-19
add traefik-route endpoint

PietroPasotti:
## Feature

This PR adds an integration with the Traefik Route charm (cf. https://github.com/PietroPasotti/traefik-route-k8s-operator), enabling users to configure Traefik on a per-relation basis, overcoming the limitations of ingress-per-unit and ingress-per-app, making Traefik a lot more general.

The role of Traefik in this integration is very limited: it simply has to accept a yaml config file and write it to the folder where traefik will load it up. Therefore the changes to the charm codebase are rather limited.

## Testing instructions
Aside from running the utests, you can test this in combination with TraefikRoute and prometheus-k8s in a live deployment (itests pending).

juju deploy traefik-route-k8s --channel beta

## Release notes
- Added `traefik_route_k8s/v0/traefik_route.py` lib
- Added `traefik_route` integration in the Traefik charm.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,7 +7,7 @@
 import enum
 import json
 import logging
-from typing import Optional, Tuple
+from typing import Tuple
 
 import yaml
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
@@ -17,6 +17,10 @@ from charms.traefik_k8s.v0.ingress_per_unit import (
     DataValidationError,
     IngressPerUnitProvider,
     RequirerData,
+)
+from charms.traefik_route_k8s.v0.traefik_route import (
+    TraefikRouteProvider,
+    TraefikRouteRequirerReadyEvent,
 )
 from deepmerge import always_merger
 from lightkube import Client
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -157,7 +159,7 @@ def _on_traefik_pebble_ready(self, _: PebbleReadyEvent):
                 },
                 # We always start the Prometheus endpoint for simplicity
                 # TODO: Generate this file in the dynamic configuration folder when the
-                # metrics-endpoint relation is established?
+                #  metrics-endpoint relation is established?
```
simskij, line 162:
```suggestion
                # metrics-endpoint relation is established?
```

src/charm.py
```diff
@@ -254,6 +256,14 @@ def _handle_ingress_request(self, event: RelationEvent):
         if isinstance(self.unit.status, MaintenanceStatus):
             self.unit.status = ActiveStatus()
 
+    def _handle_traefik_route_ready(self, event: TraefikRouteRequirerReadyEvent):
+        """A traefik_route charm has published some ingress data."""
+        self._process_ingress_relation(event.relation)
+
+        # go to active if we're in maintenance
```
simskij, line 263:
```suggestion
```

obvious from the code

PietroPasotti:
@jnsgruk I dropped #44 and replaced it with this one because rebasing gave me tons of conflicts, this was the easiest way to squash some commits and get a cleaner diff

simskij (changes requested):

src/charm.py
PietroPasotti: that was intentional. Convention is to indent multiline todo's and fixme's to make them count as one. It might be just a pycharm thing however.

jnsgruk (approved):
LGTM - thanks for the iteration on this.

src/charm.py
PietroPasotti: agreed for consistency with other TODOs/FIXMEs in the file

Issue #48, opened 2022-04-21
ingress_per_unit lacks a proper relation_departed hook


PietroPasotti:
in IPU, when a unit leaves, it would be useful for the charm to know in a more 'native' way.
Proposal: to add a

```
IngressPerUnitProvider.on.departed
```

hook, which can be used by Traefik to wipe ingress for the departing unit.
At the moment one is forced to:

`self.framework.observe(self.on.ingress_per_unit_relation_departed, self._handle_ingress_request)`

mmanciop:
In my experience, there is little value to this granularity of events, as the charm consuming the ingress urls cares only if there are urls coming from ingress or not and the logic triggered is the same. See Loki, Prometheus charms :-)

PietroPasotti:
My point is that it feels awkward to do:

```python
        self.framework.observe(self.ingress_per_unit.on.ready, self._handle_ingress_request)
        self.framework.observe(self.ingress_per_unit.on.failed, self._handle_ingress_failure)
        self.framework.observe(self.ingress_per_unit.on.broken, self._handle_ingress_broken)
        self.framework.observe(self.on.ingress_per_unit_relation_departed, self._handle_ingress_request)
```

mmanciop:
I would reconsider the entire api of the provider, frankly. Those events that mimicry the underpinning OF are awkward on their own.

mmanciop:
Ultimately you have a “need Route!” and “Route must be decommissioned!” Events

PietroPasotti:
That is also a fair point

jnsgruk:
I think just having two states/events to look at is more sensible, so "ready" and "broken" seems like a good way forward.

PR #51, opened 2022-05-09
Purge SDI from ingress (per app)

### LINKED ISSUE #37: Remove SDI dependency from `ingress` relation
Similar to #35 - but for the `ingress` endpoint

PietroPasotti:
Removed SDI from ingress (per app)

## Context
See #37 

## Testing Instructions
- Added integration tests for 
  - ingress: spring-music
  - ingress-per-unit: prometheus-k8s

The relation interface is compatible with older versions of `ingress`, but the API is different. Similarly to ingress-per-unit, the `request/response` language was stripped in favour of a more semantically close one. See #42 and #48

the Requirer has `ready` and `revoked` events: reflecting the fact that the only thing that can happen to ingress is: it can be provided (with a given url), or it can be revoked (i.e. the url is no longer available).
`ready` also includes when ingress was already ready; i.e. when the url has changed.

the Provider has now 'data_provided` and `data_removed` events, following a similar principle: either the requirer has published the data we need to provide ingress, (or changed it), or it has wiped it.

Fixes #37 
Fixes #28 

## Release Notes
- Stripped SDI from ingress interface
- Refactored and simplified the ingress API
- Added itests that check that ingress is compatible with old SDI-based ingress (as implemented by spring-music)
- Added itests that check that ingress-per-unit is compatible with old SDI-based ingress-per-unit (as implemented by prometheus-k8s)

### DIFF
--- a/lib/charms/traefik_k8s/v0/ingress.py
+++ b/lib/charms/traefik_k8s/v0/ingress.py
@@ -53,12 +53,12 @@ class SomeCharm(CharmBase):
 """
 
 import logging
-from typing import Optional, TypeVar, Dict
+from typing import Optional
 
 import yaml
-from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent, RelationRole
-from ops.framework import EventSource, StoredState, Object, ObjectEvents
-from ops.model import Relation, Unit
+from ops.charm import CharmBase, RelationEvent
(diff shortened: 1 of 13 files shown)


jnsgruk (approved):
Nice work, glad to see this landing. Only done a cursory pass, but looks good.

simskij (approved):

PR #61, opened 2022-06-15
fixed icon format

PietroPasotti:
Hopefully this will get the icon to display on charmhub?


jnsgruk (approved):

simskij:
As it already is published with an icon, this change won't do anything. Icons are only updated when you publish for the first time, or publish to stable. @roadmr will be able to assist you with having it changed.

PietroPasotti:
> As it already is published with an icon, this change won't do anything. Icons are only updated when you publish for the first time, or publish to stable. @roadmr will be able to assist you with having it changed.

Yes I know, but I already released to stable and closed a minute later from this branch in order to get the icon `applied`.

PR #63, opened 2022-06-17
Libs v1

### LINKED ISSUE #39: Remove SDI-inherited testing infrastructure

### LINKED ISSUE #48: ingress_per_unit lacks a proper relation_departed hook

### LINKED ISSUE #57: Clean up IPU requirer API
IPU Requirer's use case is to only support a relation at a time.
(What would it mean to have two ingress relations as a requirer?)

Therefore it has a `.relation -> Relation` method that assumes at most one relation exists.
However the `is_ready, is_failed, is_...` methods all take `relation` as an argument, as if there could be many. This results in complexity and ugliness.

PietroPasotti:
In an effort to remove even the last SDI heritage from the ingress libs (and uniform with https://github.com/canonical/charm-relation-interfaces), this pr:
- drops traefik/v0 libs
- introduces traefik/v1/ingress and traefik/v1/ingress-per-unit, where
  - ingress is mostly unchanged, modulo:
    - unnested the ingress data structure: it's no longer yaml-encoded under a "data" field, with all the changes that come with that.
    - some minor internal refactoring to improve understanding and code readability  
  - ingress-per-unit has the same 'unnesting' change as ingress, and on top of that:
    - the API still included some SDI-relics such as the 'ready/available/broken' etc... events. Those have been replaced with the same language as ingress: ready/revoked for the requirer, and data_provided, data_removed for the provider events.
  - in both cases, I removed the '_supported_versions' shims.
  
# QA
The unittests have been also refactored to get rid of the SDI infrastructure, I've replaced most of the remote-side-mockers with much simpler functional wrappers.
The unittests should still have the same coverage, the itests are unchanged (good sign!) except for one test that was about the shape of the relation data, which had to change.

Fixes #57 
Fixes #48
Fixes #39


lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -221,51 +252,51 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
...
+                data["host"],
+            )
 
     def _handle_relation_broken(self, event):
         self.on.data_removed.emit(event.relation)
 
     def wipe_ingress_data(self, relation: Relation):
         """Clear ingress data from relation."""
-        del relation.data[self.app]["data"]
+        del relation.data[self.app]["ingress"]
```
simskij, line 274: `del` seems like a dirty solution. Wouldn't setting it to `None` do just as well?

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -282,32 +313,29 @@ def is_ready(self, relation: Relation = None):
             log.warning("Requirer not ready; validation error encountered: %s" % str(e))
             return False
 
-    def _provider_app_data(self, relation: Relation) -> ProviderApplicationData:
-        """Fetch and validate this provider's app databag."""
-        if not relation.app or not relation.app.name:
+    def _provided_url(self, relation: Relation) -> ProviderIngressData:
+        """Fetch and validate this app databag; return the ingress url."""
+        if not relation.app or not relation.app.name or not not self.unit.is_leader():
```
simskij, line 318: Please refactor this. `if not ... or not ... or not not ...` is not an intuitive read 😝

simskij (changes requested):

lib/charms/traefik_k8s/v1/ingress.py
PietroPasotti: `del` is the official way to do it actually, setting it to `None` would hopefully raise an exception as the values are supposed to be strings. `del` is in fact an alias for `data[key] = ''`

lib/charms/traefik_k8s/v1/ingress.py
PietroPasotti: check!

lib/charms/traefik_k8s/v1/ingress.py
simskij: Ok, makes sense. Thanks

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -586,159 +609,146 @@ def __init__(
...
+            `port`: port to be used by the ingress provider to address the
+                    requirer unit.
+            `listen_to`: Choose which events should be fired on this unit:
+                "only-this-unit": this unit will only be notified when ingress
+                  is ready/revoked for this unit.
+                "all-units": this unit will be notified when ingress is
+                  ready/revoked for any unit of this application, including
+                  itself.
+                "all": this unit will receive both event types (which means it
+                  will be notified *twice* of changes to this unit's ingress!)
```
simskij, line 628: when would this mode be useful?

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -555,28 +485,121 @@ def proxied_endpoints(self) -> dict:
...
         charm: CharmBase,
         relation_name: str = DEFAULT_RELATION_NAME,
         *,
-        host: str = None,
-        port: int = None,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
+        listen_to: Literal["only-this-unit", "all-units", "both"] = "only-this-unit",
     ):
         """Constructor for IngressRequirer.
```
simskij, line 604:
```suggestion
        """Constructor for IngressPerUnitRequirer.
```

tests/integration/conftest.py
```diff
@@ -160,12 +177,38 @@ class RelationData:
...
+        get_relation_data(
+            requirer_endpoint="ipu-tester/0:ingress-per-unit",
+            provider_endpoint="traefik-k8s/0:ingress-per-unit",
+            model=model,
+        )
+        # get_relation_data(
+        #         requirer_endpoint="ipa-tester/0:ingress",
+        #         provider_endpoint="traefik-k8s/0:ingress",
+        #         model=_juju
+        #     )
```
simskij, line 213: Please remove commented code.

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -555,28 +485,121 @@ def proxied_endpoints(self) -> dict:
...
+
+
+class IngressPerUnitRevokedForUnitEvent(RelationEvent):
+    """Ingress is revoked (or has changed) for this unit.
+
+    Is only fired on the unit(s) for which ingress has changed.
+    """
+
+
+class IngressPerUnitRequirerEvents(ObjectEvents):
```
simskij, line 578: The module docs are no longer correct. Still mentions `ingress_changed`.

simskij (changes requested):
Manual testing works flawless for the happy path, however, setting `external_hostname` to "http://external-address.local" makes the scheme appear twice in the resulting URL. Same if you specify port, or both.

```
$ juju config traefik external_hostname=external-address.local:9090
$ juju show-unit prometheus/0 

...
  url: http://external-address.local:9090:80/traefik-test-prometheus-k8s-0
...
```

```
$ juju config traefik external_hostname=http://external-address.local
$ juju show-unit prometheus/0 

...
  url: http://http://external-address.local:80/traefik-test-prometheus-k8s-0
...
```

```
$ juju config traefik external_hostname=http://external-address.local:9090
$ juju show-unit prometheus/0 

...
  url: http://http://external-address.local:9090:80/traefik-test-prometheus-k8s-0
...
```

lib/charms/traefik_k8s/v1/ingress_per_unit.py
PietroPasotti: Maybe you want a specialized handler for when this unit's url changes, and a generic handler for when the url of any peer unit (including yourself) does.

But in general it felt wrong to constrain the API to only one of the two choices. I wanted to leave the option open to 'just fire all events'.

simskij (approved):

PR #66, opened 2022-06-20
POC diff for pydantic port

PietroPasotti:
DO NOT MERGE

This is a proof-of-concept diff to see what the code would look like if we were to replace jsonschema with pydantic in ingress.py.
Code is untested and probably buggy, I expect it to be 90% production-ready. Goal is to see the (rough) diff and have a discussion before finalizing.

- Dropped the conditional dependency: to use the lib you NEED to install pydantic. Before: jsonschema was optional. Could keep that logic (pydantic integrates with TypedDict)
- Pydantic models offer a lot of validation flexibility, I just skimmed the surface here to show what's possible
- Pydantic models can replace the typeddict-based orm I had before: pydantic models have nice pythonic dot-notation access.
- Note the Json field: we can document explicitly what encoding we're using for nested data fields. Nice!

### DIFF
--- a/lib/charms/traefik_k8s/v0/ingress.py
+++ b/lib/charms/traefik_k8s/v0/ingress.py
@@ -51,7 +51,7 @@ class SomeCharm(CharmBase):
 
 import logging
 import socket
-from typing import Optional
+from typing import Optional, Any
 
 import yaml
 from ops.charm import CharmBase, RelationEvent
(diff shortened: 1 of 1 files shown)


lib/charms/traefik_k8s/v0/ingress.py
```diff
@@ -74,84 +74,32 @@ def _handle_ingress(self, event):
...
-
-
-class RequirerData(TypedDict):
-    """Model of the data a unit implementing the requirer will need to provide."""
-
-    model: str
-    name: str
-    host: str
-    port: int
+    import pydantic
```
simskij, line 77: this is extremely promising, and now that we have PYDEPS, resolving and pulling pedantic will likely be a lot neater as well.

PietroPasotti: true that! Maybe worth it to revisit this direction in light of pydeps

PR #69, opened 2022-07-05
promote action

dstathis:
Github action to promote charm taken directly from grafana-agent-k8s.


sed-i (approved):

sed-i (commented):
BTW where are the user permissions for running this action?

PR #72, opened 2022-07-27
Tcp routing

### LINKED ISSUE #18: Support TCP routing
For database endpoints that need to be exposed outside of a Kubernetes cluster (e.g., through most cross-model relations), we should update the relation interfaces and the libraries to provide means of requesting `tcp` routing (through [TCP Routers](https://doc.traefik.io/traefik/routing/routers/#configuring-tcp-routers)) rather than just `http` ones. @wolsen care to provide more details for your use-cases?

PietroPasotti:
Fixes #18 

## Solution
Added a TCP routing mode via ingress-per-unit, only applicable in combination with `path` routing mode.
- IPU requirer databag updated with a `mode: tcp|http`
- Traefik updated to provide tcp routing.

## Context
This is implemented by adding a per-unit entryPoint in traefik's **static configuration**. This means that on each change, Traefik needs to restart. This means that we might drop requests-in-flight. Do we need some rolling-graceful-reboot to mitigate?

## Testing Instructions
Itests under test_charm_tcp should provide some inspiration.

## Please check out:
I would appreciate especially a second pair of eyes on how I refactored the pebble-ready flow. Reason for that is that I needed to re-run some 'setup and start traefik' code on TCP config changes since these require a traefik reboot. Container.replan was not enough because the service hadn't in fact changed, only its static config did. So I added a conditional Container.restart(traefik-service-name). 

## Release Notes
* added TCP routing support for Traefik.


simskij (changes requested):
All in all looks good! Can you add a test that runs tcp and http in parallel for different relations?

PietroPasotti:
> All in all looks good! Can you add a test that runs tcp and http in parallel for different relations?

Technically, that is already the case.
If you run all tests in a single model (like it happens in the CI pipeline since we pass `--model something` in the workflow, you'll get tcp and ipa/ipu tests all in the same model.

Although I guess it's possible that since the suites are run sequentially, the tcp suite might 'break' the earlier suites, so we could do something like 
- test ipa
- test ipu
- test tcp
- test ipu again

to catch that scenario?

Or how would you propose to test them 'in parallel'?

simskij:
> > All in all looks good! Can you add a test that runs tcp and http in parallel for different relations?
> 
> Technically, that is already the case.
> 
> image
> 
> If you run all tests in a single model (like it happens in the CI pipeline since we pass `--model something` in the workflow, you'll get tcp and ipa/ipu tests all in the same model.
> 
> Although I guess it's possible that since the suites are run sequentially, the tcp suite might 'break' the earlier suites, so we could do something like
> 
> * test ipa
> * test ipu
> * test tcp
> * test ipu again
> 
> to catch that scenario?
> 
> Or how would you propose to test them 'in parallel'?

* test ipa
* test ipu
* test tcp
* test ipa to one charm and TCP to another one
* test ipu to one charm and TCP to another one

While I agree that it implicitly does this already, I'd like to see it become an explicit test to lessen the risk for regression down the line.
 
:)

PietroPasotti:
@simskij took me a while to refactor things to avoid duplication, but I think I found a decent solution.
Added a compatibility test that verifies that a tcp relation + regular ipa/ipu work simultaneously

simskij (approved):

PR #77, opened 2022-08-11
Block if hostname is invalid

### LINKED ISSUE #76: external_hostname silently accepts a port and updates reldata with invalid address
### Bug Description

When `juju config traf external_hostname=192.168.1.10:8080`, the other end of the relation sees

```yaml
  - relation-id: 1
    endpoint: ingress
    related-endpoint: ingress-per-unit
    application-data:
      ingress: |-
        prom/0:
          url: http://192.168.1.10:8080:80/welcome-prom-0
```

### To Reproduce

```shell
juju deploy --channel=edge prometheus-k8s prom
juju deploy --channel=edge traefik-k8s traf
juju relate prom:ingress traf
juju config traf external_hostname=192.168.1.10:8080
```

### Environment

microk8s/localhost  2.9.33
traefik-k8s     edge      84

### Relevant log output

```shell
unit-prom-0: 14:49:14.331 ERROR unit.prom/0.juju-log ingress:1: Uncaught exception while in charm code:
  File "./src/charm.py", line 610, in <module>
    main(PrometheusCharm)
  File "/var/lib/juju/agents/unit-prom-0/charm/venv/ops/main.py", line 419, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 139, in __init__
    endpoint_port=external_url.port or self._port,
  File "/usr/lib/python3.8/urllib/parse.py", line 177, in port
    raise ValueError(message) from None
ValueError: Port could not be cast to integer value as '8080:80'
```

### Additional context

_No response_

sed-i:
## Issue
Fixes #76.

## Solution
- Block if hostname is invalid.

## Context
For COS Lite load test, traefik is interfering with the manual nginx ingress. To circumvent that I mistakenly tried to set a custom port as part of `external_hostname`. 

(Note: could address this is a separate PR by introducing an `external_port` config option.)

## Testing Instructions
- utests
- deploy and `juju config external_hostname="withport:8080"` should block.

## Release Notes
Block if hostname is invalid.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,6 +9,7 @@ import json
 import logging
 import typing
 from typing import Tuple, Union
+from urllib.parse import urlparse
 
 import yaml
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
@@ -262,10 +263,24 @@ class TraefikIngressCharm(CharmBase):
             )
             return
 
-        if not self._external_host:
+        hostname = self._external_host
+
+        if not hostname:
             self.unit.status = MaintenanceStatus("resetting ingress relations")
             self._wipe_ingress_for_all_relations()
             self.unit.status = WaitingStatus("gateway address unavailable")
+            return
+
+        if hostname != urlparse(f"scheme://{hostname}").hostname:
+            self.unit.status = MaintenanceStatus("resetting ingress relations")
+            self._wipe_ingress_for_all_relations()
+            self.unit.status = BlockedStatus(f"invalid hostname: {hostname}; see logs.")
+
+            logger.error(
+                "'%s' is not a valid hostname value; "
+                "hostname must not include port or any other netloc components",
+                hostname,
+            )
             return
 
         if not self._traefik_service_running:
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -565,7 +565,7 @@ class TestTraefikIngressCharm(unittest.TestCase):
         assert yaml.safe_load(static_config)["entryPoints"][prefix] == expected_entrypoint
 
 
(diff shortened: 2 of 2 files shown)


sed-i:
Do we want an `external_port` config option? Or expand `external_hostname` into effectively `external_netloc`?

simskij:
> Do we want an `external_port` config option? Or expand `external_hostname` into effectively `external_netloc`?

I don't think we really want the port to be configurable at all. As each deployment of traefik will only have one entrypoint, and won't be using node ports, it's fine that it's hard coded to 80 (and later 443 for SSL) IMHO.

For TCP, which is the only place it would make sense for me to have it, we already have something in place that solves this issue.

PietroPasotti:
@simskij attempting to paraphrase your conclusions:

- it's OK if we add better validation of the external_hostname arg to verify that it's indeed a valid hostname (and just a hostname, no port).
- we shouldn't change traefik's config further.

simskij:
> @simskij attempting to paraphrase your conclusions:
> 
> * it's OK if we add better validation of the external_hostname arg to verify that it's indeed a valid hostname (and just a hostname, no port).
> * we shouldn't change traefik's config further.

Correct!

src/charm.py
```diff
@@ -262,12 +263,26 @@ def _process_status_and_configurations(self):
...
-        if not self._external_host:
+        hostname = self._external_host
+
+        if not hostname:
             self.unit.status = MaintenanceStatus("resetting ingress relations")
             self._wipe_ingress_for_all_relations()
             self.unit.status = WaitingStatus("gateway address unavailable")
             return
 
+        if hostname != urlparse(f"scheme://{hostname}").hostname:
```
simskij, line 274: Clever! 👏🏼

PR #80, opened 2022-08-18
[IPU] Make sure `revoked` is emitted

### LINKED ISSUE #79: possible issue with IPURequirer.on.revoked emission
### Bug Description

ATM we uniformly handle changed and broken, as a result we do:

`current_urls = self.urls or {}`

we cannot guarantee however that if we're in a `broken` context, urls will be None as we expect (Traefik wipes the data, but it might be that juju hasn't delete it yet, or the requirer otherwise still can see it).

Instead, we could force `current_urls` to be `{} if isinstance(event, BrokenEvent)` to generate, regardless of traefik's data, the correct delta.

### To Reproduce

Unclear yet whether this issue results in a real reproducible bug.

### Environment

-- 

### Relevant log output

```shell
--
```

### Additional context

_No response_

sed-i:
## Issue
Closes #79 .

## Solution
- [x] add utest(s) for IPU event
- [x] return an empty dict if in relation-broken event instead of looking into relation data, which may be out of date.

## Context
- `self.ingress.url` was returning old (non-falsey) value even after relation and/or app was removed.
- Seems like relation data is not emptied out even when looking from within a relation broken event. This may be a juju feature rather than a bug though/

### DIFF
--- a/lib/charms/observability_libs/v0/juju_topology.py
+++ b/lib/charms/observability_libs/v0/juju_topology.py
@@ -1,0 +1,306 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""## Overview.
+
+This document explains how to use the `JujuTopology` class to
+create and consume topology information from Juju in a consistent manner.
+
+The goal of the Juju topology is to uniquely identify a piece
+of software running across any of your Juju-managed deployments.
+This is achieved by combining the following four elements:
+
+- Model name
+- Model UUID
+- Application name
+- Unit identifier
+
+
+For a more in-depth description of the concept, as well as a
+walk-through of it's use-case in observability, see
+[this blog post](https://juju.is/blog/model-driven-observability-part-2-juju-topology-metrics)
+on the Juju blog.
+
+## Library Usage
+
+This library may be used to create and consume `JujuTopology` objects.
+The `JujuTopology` class provides three ways to create instances:
+
+### Using the `from_charm` method
+
+Enables instantiation by supplying the charm as an argument. When
+creating topology objects for the current charm, this is the recommended
+approach.
+
+```python
+topology = JujuTopology.from_charm(self)
+```
+
+### Using the `from_dict` method
+
+Allows for instantion using a dictionary of relation data, like the
+`scrape_metadata` from Prometheus or the labels of an alert rule. When
+creating topology objects for remote charms, this is the recommended
+approach.
+
+```python
+scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata", "{}"))
+topology = JujuTopology.from_dict(scrape_metadata)
+```
+
+### Using the class constructor
+
+Enables instantiation using whatever values you want. While this
+is useful in some very specific cases, this is almost certainly not
+what you are looking for as setting these values manually may
+result in observability metrics which do not uniquely identify a
+charm in order to provide accurate usage reporting, alerting,
+horizontal scaling, or other use cases.
+
+```python
+topology = JujuTopology(
+    model="some-juju-model",
+    model_uuid="00000000-0000-0000-0000-000000000001",
+    application="fancy-juju-application",
+    unit="fancy-juju-application/0",
+    charm_name="fancy-juju-application-k8s",
+)
+```
+
+"""
+
+import re
+from collections import OrderedDict
+from typing import Dict, List, Optional
+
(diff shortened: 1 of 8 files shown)


sed-i:
@PietroPasotti I am having trouble with the utest but I'm not sure if it's because of a test bug, lib bug or harness behavior:

The relation is removed with harness, but "removed" is always blank:

```
INFO     root:ingress_per_unit.py:689 unit: ipu-requirer/0, event: RelationCreatedEvent. removed: set(); changed: set()
INFO     root:ingress_per_unit.py:689 unit: ipu-requirer/0, event: RelationJoinedEvent. removed: set(); changed: set()
INFO     root:ingress_per_unit.py:689 unit: ipu-requirer/0, event: RelationChangedEvent. removed: set(); changed: {'ipu-requirer/0'}
INFO     root:ingress_per_unit.py:689 unit: ipu-requirer/0, event: RelationChangedEvent. removed: set(); changed: set()
INFO     root:ingress_per_unit.py:689 unit: ipu-requirer/0, event: RelationBrokenEvent. removed: set(); changed: set()
```

and `revoke` doesn't fire.

sed-i:
`{} if isinstance(event, BrokenEvent)` per @PietroPasotti's suggestion fixed the utest.

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -682,7 +682,7 @@ def _handle_relation(self, event: RelationEvent):
         # we calculate the diff between the urls we were aware of
         # before and those we know now
         previous_urls = self._stored.current_urls or {}  # type: ignore
```
sed-i, line 684: @rbarry82 would you use `_type_convert_stored` here?

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,306 @@
+# Copyright 2022 Canonical Ltd.
```
PietroPasotti, line 1: why did you add juju_topology?

PietroPasotti (approved):
looks good!

lib/charms/observability_libs/v0/juju_topology.py
sed-i: It's a dependency: the new prometheus_scrape (after fetch-lib) depends on it.

Abuelodelanada (approved):

rbarry82 (approved):
Tested manually

PietroPasotti:
@sed-i it seems I didn't pay enough attention while reviewing, or actually, that some changes slipped in that I didn't have a chance to review as they came in after I approved: https://github.com/canonical/traefik-k8s-operator/pull/80/commits/ce32bd3d17f9166f5ef228da7dee58cbb870ed5b

either way, there's a problem with these changes:

1) we renamed `urls` to `_get_urls_from_relation_data`, which shouldn't really be a property anymore. 
2) we replaced `urls` with a function that loads in the urls from stored data, which changes some of the semantics of what's going on, enough to break integration tests in `traefik-route`. So that's a backwards-incompatible change!

sed-i:
> we renamed urls to _get_urls_from_relation_data, which shouldn't really be a property anymore.

Agreed. I noticed that later too. In the IPA PR it is not a property.

> we replaced urls with a function that loads in the urls from stored data... break integration tests

Hmm. The change itself still seems correct to me logically. Can you post the link to the failing test? I'm curious.

PietroPasotti:
It wasn't a CI failure, it was local, so no links, but essentially what happened was:
I was writing a tester charm to represent an ingress-requiring-charm like:

```python
class MyCharm(CharmBase):
    def __init__...
        self.unit.status=Blocked()
        ingress = IPURequirer()
        if ingress.is_ready(ingress.relation):
            self.unit.status=Active()
```
This test was passing before this PR got merged.
reason: is_ready() basically checks if `bool(urls)`.
before this PR, `urls` would fetch the data in real time from the relation.
after this PR, `urls` is cached and only contains data if we are processing some events.

However, events are processed AFTER the charm is initialized, so at charm-init-time, that branch wouldn't be selected.

Took me a long time to figure that one out :)
Guess I learned to write charms always 'the proper way', even though it's just a tester charm.

sed-i:
Interesting. I wonder why at charm init time the old `url` would be non-false-y. And if indeed it was, then was it a mistake?

PR #82, opened 2022-08-19
workaround for ghost-relations bug

### LINKED ISSUE #81: Permission denied error when providing ingress requirements in IPU
### Bug Description

ModelErrors are fired when prometheus tries to access `self.ingress.relation.data`, because `self.relation` simply does `self.relations[0]` and, apparently, relations contains not one but two relations, the first of which is a ghost (possible juju bug?)

The second relation is the one we want.

### To Reproduce

juju deploy prometheus-k8s --channel beta
juju deploy traefik-k8s --channel edge
juju relate prometheus-k8s:ingress traefik

juju remove-application traefik-k8s
juju deploy traefik-k8s --channel edge --application-name='trfk'
juju relate prometheus-k8s:ingress trfk

### Environment

edge

### Relevant log output

```shell
File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/lib/charms/traefik_k8s/v1/ingress_per_unit.py", line 70
3, in _handle_relation                                                                                          
    self._publish_auto_data(typing.cast(Relation, event.relation))                                              
  File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/lib/charms/traefik_k8s/v1/ingress_per_unit.py", line 71
1, in _publish_auto_data                                                                                        
    self.provide_ingress_requirements(host=self._host, port=self._port)                                         
  File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/lib/charms/traefik_k8s/v1/ingress_per_unit.py", line 75
1, in provide_ingress_requirements                                                                              
    self.relation.data[self.unit].update(data)                                                                  
  File "/usr/lib/python3.8/_collections_abc.py", line 832, in update
    self[key] = other[key]
  File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/venv/ops/model.py", line 938, in __setitem__
    self._backend.relation_set(self.relation.id, key, value, self._is_app)
  File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/venv/ops/model.py", line 2137, in relation_set
    return self._run(*args)
  File "/var/lib/juju/agents/unit-prometheus-k8s-0/charm/venv/ops/model.py", line 2036, in _run
    raise ModelError(e.stderr)
ops.model.ModelError: b'ERROR cannot read relation settings: permission denied\n'
```
```


### Additional context

unclear whether it's a juju bug or not

### PR DESCRIPTION
**Pull request by PietroPasotti**
## Issue
Fixes  #81

## Solution
1) add a `_relation_data_exists` helper that checks that `relation.data` is accessible
2) use that helper to filter `model.relations['ingress']` when trying to determine what are the ingress relations (both in IPA and IPU)
3) add a safety check in `self.relation` to verify that `self.relations` contains exactly one relation in [IPU/IPA]Requirer.

Little feature creep:
consolidated leadership checks and raise LeadershipError instead of assertion errors on leader guard fails.

## Testing Instructions
A utest wouldn't catch this.
I could test that this works to fix #81 by:
1) reproduce #81 
2) juju scp ingress_per_unit to both charms
3) juju resolve prometheus-k8s/0

--> 

image

Unclear whether we should be integration-testing this and how.

## Release Notes
* fixed bug where accessing relation data would raise PermissionErrors because the relation was defunct.

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress.py
+++ b/lib/charms/traefik_k8s/v1/ingress.py
@@ -59,7 +59,7 @@ from typing import Any, Dict, Optional, Tuple, Union
 import yaml
 from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent
 from ops.framework import EventSource, Object, ObjectEvents, StoredState
-from ops.model import ModelError, Relation
+from ops.model import ModelError, Relation, Unit
 
 # The unique Charmhub library identifier, never change it
 LIBID = "e6de2a5cd5b34422a204668f3b8f90d2"
@@ -69,7 +69,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 DEFAULT_RELATION_NAME = "ingress"
 RELATION_INTERFACE = "ingress"
@@ -140,12 +140,35 @@ def _validate_data(data, schema):
         raise DataValidationError(data, schema) from e
 
 
+def _relation_data_exists(relation: Relation):
+    """Helper to verify that relation data exists.
+
+    In some cases when we receive a relation event, the relation data
+    is not ready (yet).
+    """
+    try:
+        _ = relation.data
+        return True
+    except ModelError as e:
+        log.debug("relation data access failed with {!r}".format(e))
+        return False
+
+
 class DataValidationError(RuntimeError):
     """Raised when data validation fails on IPU relation data."""
 
 
+class LeadershipError(RuntimeError):
+    """Raised when a follower unit attempts an operation that requires leadership."""
+
+
+def _leader_check(unit: Unit):
+    if not unit.is_leader():
+        raise LeadershipError(unit)
+
+
 class _IngressPerAppBase(Object):
-    """Base class for IngressPerUnit interface classes."""
+    """Base class for IngressPerApp interface classes."""
 
     def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME):
         super().__init__(charm, relation_name)
@@ -167,7 +190,10 @@ class _IngressPerAppBase(Object):
     @property
     def relations(self):
         """The list of Relation instances associated with this endpoint."""
-        return list(self.charm.model.relations[self.relation_name])
+        all_relations = self.charm.model.relations[self.relation_name]
+        # we filter out relations whose data is unaccessible, these are
+        # probably ghost relations (TODO: reference juju bug?)
+        return list(filter(_relation_data_exists, all_relations))
 
     def _handle_relation(self, event):
         """Subclasses should implement this method to handle a relation update."""
@@ -281,7 +307,7 @@ class IngressPerAppProvider(_IngressPerAppBase):
 
     def wipe_ingress_data(self, relation: Relation):
         """Clear ingress data from relation."""
-        assert self.unit.is_leader(), "only leaders can do this"
+        _leader_check(self.unit)
         try:
             relation.data
         except ModelError as e:
@@ -453,6 +479,9 @@ class IngressPerAppRequirer(_IngressPerAppBase):
             self._auto_data = None
 
     def _handle_relation(self, event):
+        if not _relation_data_exists(event.relation):
+            return event.defer()
+
         # created, joined or changed: if we have auto data: publish it
         self._publish_auto_data(event.relation)
 
@@ -487,9 +516,9 @@ class IngressPerAppRequirer(_IngressPerAppBase):
     def _publish_auto_data(self, relation: Relation):
         if self._auto_data and self.unit.is_leader():
             host, port = self._auto_data
-            self.provide_ingress_requirements(host=host, port=port)
+            self.provide_ingress_requirements(host=host, port=port, relation=relation)
 
-    def provide_ingress_requirements(self, *, host: Optional[str] = None, port: int):
+    def provide_ingress_requirements(self, *, host: Optional[str] = None, port: int, relation: Optional[Relation] = None):
         """Publishes the data that Traefik needs to provide ingress.
 
         NB only the leader unit is supposed to do this.
@@ -498,12 +527,17 @@ class IngressPerAppRequirer(_IngressPerAppBase):
             host: Hostname to be used by the ingress provider to address the
              requirer unit; if unspecified, FQDN will be used instead
             port: the port of the service (required)
+            relation: the relation via which ingress should be provided.
+            If unspecified, will default to the (unique) established
+            'ingress' relation.
         """
         # get only the leader to publish the data since we only
         # require one unit to publish it -- it will not differ between units,
         # unlike in ingress-per-unit.
-        assert self.unit.is_leader(), "only leaders should do this."
-        assert self.relation, "no relation"
+        _leader_check(self.unit)
+        rel = relation or self.relation
+        if not rel:
+            raise RuntimeError("no relation; cannot provide ingress requirements")
 
         if not host:
             host = socket.getfqdn()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/charms/traefik_k8s/v1/ingress.py [hunk 1]
```diff
@@ -502,7 +538,12 @@ def provide_ingress_requirements(self, *, host: str = None, port: int):
     @property
     def relation(self):
         """The established Relation instance, or None."""
-        return self.relations[0] if self.relations else None
+        relations = self.relations
+        if not relations:
+            return None
+        if len(relations) > 2:
+            raise RuntimeError("Too many ingress relations: {}".format(relations))
```
rbarry82, line 545: May as well include a suggestion for `limit: 1` in the message as a pointer to authors who may know that option in the metadata.

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -679,6 +707,11 @@ def __init__(
         )
 
     def _handle_relation(self, event: RelationEvent):
+        rel = event.relation
+        if not isinstance(rel, Relation):  # type guard
+            raise RuntimeError('event.relation is not Relation but {}'.format(rel))
+        if not _relation_data_exists(rel):
+            return event.defer()
```
rbarry82, line 714: Do we need to defer, or could we just return? A `relation-created` or `relation-joined` which hasn't set any data yet won't have any when the deferred event is re-scheduled, so it will just re-fire and  cycle here.

When data is actually set, `relation-changed` will hit here and actually have data, so need to defer.

rbarry82 (commented):
An additional handler for `relation_departed` in IPU may have slipped through the cracks, may also help. 

In particular, in IPU, `_handle_relation_broken` emits an event about revocation, where the docstring says:

> Event triggered when a requirer has wiped its ingress data.

But no such wipe ever happened. The data is still in the databag, so any charm author who does not take this docstring as tautologically true can still fetch it. For the particular case we discussed (Loki), on a `ingress_revoked` event, it [tries to determine its external URL](https://github.com/canonical/loki-k8s-operator/blob/b2a758975f679b6299b97e32a9cf3219da5dbbd9/src/charm.py#L195). It does this not by checking relation data, but by directly asking `IPU.url`. Since the data was not actually wiped, in the case where a `relation-broken` may run for Traefik (which emits a revocation event), that revocation may be processed by Loki and since the data wasn't wiped, `IPU.url` happily returns an address.

Realistically, Loki (or our charms) could handle [available and revoked in different methods](https://github.com/canonical/loki-k8s-operator/blob/b2a758975f679b6299b97e32a9cf3219da5dbbd9/src/charm.py#L91), but it's likely to be an easy mistake for other users of the ingress library to make which is easily stopped by wiping the data from the Traefik side.

As a couple of comments on the side:

* We probably don't need to emit for every different in `_handle_relation`. The access pattern appears to be "see an event, ask `ingress.url[s]` for updated endpoints". The repeated invocations don't matter all that much for performance, since OF's `LazyMapping` will already have loaded the key, so it's just a dict lookup, but there won't be a functional difference between `_handle_relation` breaking out of the loop if _any_ `ready_for|revoked_for` would be emitted and just doing it once.
* The comparison in `_handle_relation` between `previous_urls` and `current_urls` will probably always be false. [This comment](https://github.com/canonical/operator/pull/572#issuecomment-931529580) explains why, with example code and a workaround which recurses `Stored*` objects to convert them back. We use this in a couple of a charms with more "complex" datastructures than `{k: str : val : str}`. Given the structure of `urls`, it's likely that this happens here also. The `jsonschema` is simple enough that any change in the relation data _should_ fire it for now, but if the complexity of what's in `urls` ever changes away from `{unit_name: url}` to add Traefik middlewares or something, it may be an unpleasant surprise. If you're relying on equality testing anything in `StoredState` with higher complexity than a string (`StoredDict`, `StoredList`, etc), it's a reasonable safety valve to just convert it all back to Python types beforehand, and that concern magnifies if you have `self.previous_urls = {unit_name: url, other_key: [{foo:bar}, {baz: quux}]}`, because each "level" will get converted, and you'll get back `StoredDict[str: StoredList[StoredDict]]` instead of `dict[str: list[dict]]`. Not necessary for now, just noticed on a readthrough

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -135,12 +135,35 @@ def _validate_data(data, schema):
...
     """Raised when data validation fails on IPU relation data."""
 
 
+class LeadershipError(RuntimeError):
+    """Raised when a follower unit attempts an operation that requires leadership."""
+
+
+def _leader_check(unit: Unit):
+    if not unit.is_leader():
+        raise LeadershipError(unit)
```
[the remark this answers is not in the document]
PietroPasotti: Before: you'd get a `ModelError` telling you you need leadership, now you get a custom, more specific error. I didn't want to change the behaviour of the code, just to provide a different error.
The philosophy is: fail early and loudly. So I don't want `method_I_should_not_be_calling_unless()` to fail silently, I want it to fail loudly and force the user to: `if can_call(): method_I_should_not_be_calling_unless()`

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -273,7 +299,7 @@ def _handle_relation_broken(self, event):
 
     def wipe_ingress_data(self, relation: Relation):
         """Clear ingress data from relation."""
-        assert self.unit.is_leader(), "only leaders can do this"
+        _leader_check(self.unit)
         try:
             relation.data
```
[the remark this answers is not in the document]
PietroPasotti: relation data is loaded on access and cached. so there is no chance that (so far as the OF can know) relation.data changes between the `relation.data` test and the `del` statement, I think.

As for delaying the leader check, is there a chance that fetching the relation data will take longer than 30 seconds?

lib/charms/traefik_k8s/v1/ingress_per_unit.py
PietroPasotti: my thought was:
- relation-joined, no data -> defer
- [other events], still no data -> re-defer
- [relation-changed], now we do have data -> finally consume the deferred relation-joined event.
But, I reckon, we could just as well wait until we get the relation-changed event to signal that there is data.
In that case, what is the point of listening to -joined and -created at all, when we could just wait for -changed?

PietroPasotti:
> * The comparison in `_handle_relation` between `previous_urls` and `current_urls` will probably always be false.

that's a good point, and thanks for the pointer to that interesting discussion :)
added a comment to warn future-me about the issue, were the stored data ever to become more complex.

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -152,23 +175,27 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
...
-        observe(rel_events.relation_joined, self._handle_relation)
-        observe(rel_events.relation_changed, self._handle_relation)
-        observe(rel_events.relation_broken, self._handle_relation_broken)
+        observe(rel_events.relation_created, self._handle_relation_available)
+        observe(rel_events.relation_joined, self._handle_relation_available)
+        observe(rel_events.relation_changed, self._handle_relation_available)
+        observe(rel_events.relation_broken, self._handle_relation_unavailable)
+        observe(rel_events.relation_departed, self._handle_relation_unavailable)
         observe(charm.on.leader_elected, self._handle_upgrade_or_leader)
         observe(charm.on.upgrade_charm, self._handle_upgrade_or_leader)
```
simskij, line 184: In most other charms, `_on_whatever` seems to be the consensus style of naming handlers. This is purely stylistic however, so feel free to dismiss it. :)

simskij (approved):

PietroPasotti:
close because horribly stale

PR #84, opened 2022-08-29
[IPA] Make sure `ingress.url` returns an up-to-date value

sed-i:
An IPA analogue to #80.

## Context
https://github.com/canonical/alertmanager-k8s-operator/pull/94

## Release Notes
[IPA] Make sure `ingress.url` returns an up-to-date value

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress.py
+++ b/lib/charms/traefik_k8s/v1/ingress.py
@@ -54,7 +54,7 @@ class SomeCharm(CharmBase):
 import logging
 import socket
 import typing
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Optional, Tuple
 
 import yaml
 from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent
@@ -446,7 +446,7 @@ class IngressPerAppRequirer(_IngressPerAppBase):
         if self.is_ready():
             # Avoid spurious events, emit only when there is a NEW URL available
             new_url = (
-                ""
+                None
                 if isinstance(event, RelationBrokenEvent)
                 else self._get_url_from_relation_data()
             )
@@ -455,7 +455,7 @@ class IngressPerAppRequirer(_IngressPerAppBase):
                 self.on.ready.emit(event.relation, new_url)
 
     def _handle_relation_broken(self, event):
-        self._stored.current_url = ""
+        self._stored.current_url = None
         self.on.revoked.emit(event.relation)
 
     def _handle_upgrade_or_leader(self, event):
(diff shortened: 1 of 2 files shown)


PietroPasotti (commented):
Why the return type change in 'url'? 
That's a backwards-incompatible change. What if user code is doing `if self.ipa.url is None...`?

tests/unit/test_lib_per_app_requires.py
```diff
@@ -89,3 +90,69 @@ def test_validator(requirer: IngressPerAppRequirer, harness, auto_data, ok):
...
+
+        # AND an ingress is in effect
+        data = {"url": "http://a.b/c"}
+        self.harness.update_relation_data(
+            self.rel_id, "traefik-app", {"ingress": yaml.safe_dump(data)}
+        )
+        self.assertEqual(self.harness.charm.ipa.url, "http://a.b/c")
+
+        # THEN the ready event is emitted
+        after = self.harness.charm.ready_event_count
```
PietroPasotti, line 143: this seems like a good place to use https://discourse.charmhub.io/t/harness-recipe-capture-events/6581

simskij:
> Why the return type change in 'url'? That's a backwards-incompatible change. What if user code is doing `if self.ipa.url is None...`?

Agreeing with @PietroPasotti here. I get that the behavior is equivalent in case of `if not self.ipa.url`, which probably is what users should do in the first place, but we have no guarantees that is the case. It also seems like returning None is the expected behavior here.

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -522,11 +526,21 @@ def url(self) -> Optional[str]:
...
+
+    @property
+    def url(self) -> str:
+        """The full ingress URL to reach the current unit.
+
+        May return None if the URL isn't available yet.
+        """
+        data = self._stored.current_url or ""  # type: ignore
+        assert isinstance(data, str)  # for static checker
+        return data
```
Abuelodelanada, line 546: If this method "May return None", shouldn't be the return type `Optional[str]`?

tests/unit/test_lib_per_app_requires.py
sed-i: Done. This is magic!

lib/charms/traefik_k8s/v1/ingress.py
sed-i: Fixed.

sed-i:
Thanks for the reviews!
I reverted the "" back to None. Originally I was concerned about None going into and from Stored.

PietroPasotti (approved):
nice work!

PR #92, opened 2022-10-18
Increase the visibility of `_external_host`

rbarry82:
## Issue
It is not currently possible to forward `_external_host` as a "known" value back to `TraefikRouteRequirer` without calling private methods, making it unwieldy to inform `TraefikRouteProviders` what the `external_host` may be.

## Solution
Make `_external_host` into `external_host`, and pass it as a constructor argument to `TraefikRouteProvider`

## Context
See this [grafana-k8s issue](https://github.com/canonical/grafana-k8s-operator/issues/143#event-7605411772), for example.

## Release Notes
Increase the visibility of `_external_host`, update `traefik_route` libraries

### DIFF
--- a/lib/charms/observability_libs/v0/juju_topology.py
+++ b/lib/charms/observability_libs/v0/juju_topology.py
@@ -67,16 +67,15 @@ topology = JujuTopology(
 ```
 
 """
-
-import re
 from collections import OrderedDict
 from typing import Dict, List, Optional
+from uuid import UUID
 
 # The unique Charmhub library identifier, never change it
 LIBID = "bced1658f20f49d28b88f61f83c2d232"
 
 LIBAPI = 0
-LIBPATCH = 2
+LIBPATCH = 3
 
 
 class InvalidUUIDError(Exception):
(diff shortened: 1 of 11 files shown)


simskij (approved):

Abuelodelanada (approved):

PietroPasotti (approved):
Looks good! 
I thought we had agreed to increase the visibility of `_external_host` to `external_host` in the charm, so that the TraefikRouteProvider could get it by doing `self.charm.external_host`.
And we said that that was better than adding it as an init arg to TraefikRouteProvider.

I see you went for the latter option, which is fine by me (simplifies the dependencies), but that means that we don't really need to make `external_host` public, since it's the charm passing it down itself: nothing wrong with
```
        self.traefik_route = TraefikRouteProvider(charm=self, external_host=self._external_host)
```

or am I missing something?

rbarry82:
I went the other route primarily due to a debate on the team internally, which we'll probably have to settle out in person. 

I'm happy to change the visibility back. I'm equally happy to leave it as it is. It's kind of a wash from my point of view, and you're the maintainer, so I'll leave it up to you. From my point of view, the somewhat obsession/cargo cult around making absolutely every method private, even if it can't change the data model at all (`_external_host` doesn't have any ability to actually set anything anywhere in the call tree) and doesn't expose any "internals" users shouldn't rely on (like `_get_loadbalancer_status` may) is something I shrug my shoulders about.

In practical terms, beyond the fact that nothing about `external_host` is confusing or exposing internals which may change regardless of visibility, I have an even weaker opinion on visibility on things inside implementations of `CharmBase`. They are not and cannot be used as libraries by anyone, and the worst side effect is more or less that users of vscode/pycharm/emacs/whatever who are writing `self.charm.model.relations....` inside a library will see a single "public" method briefly. 

It's developer experience, sure, but the visibility here only really applies to charm authors working on Traefik itself during development for completion dropdowns, and not "internal implementations" from psycopg or matplotlib or similar. 

If you have a strong opinion about making it private, I'll change it back, otherwise I'd probably just leave it as is.

PietroPasotti:
> I went the other route primarily due to a debate on the team internally, which we'll probably have to settle out in person.

I don't disagree. for charms specifically there seems to be a tradition of making all methods private -- which as a charm is not supposed to be used by any other object makes sense, but is also not really necessary. 
I was mostly wondering what was the reason for the change given that we went the other route.
No need to revert, thanks for the explanation.

PR #94, opened 2022-10-20
Don't blow up integration tests if IPU/IPA/IPR apps from o11y fail common hooks during deploy due to resourcepatch

rbarry82:
## Issue
Due to `--trust=True` and `KubernetesResourcePatch` causing sometimes unpredicatable failures (more common with simultaneous operations on CPU-limited environments, like CI on GH runners), arbitrary hooks (`stop`, `leader-elected`, etc) may momentarily be marked as an error. `python-libjuju` _immediately_ catches those and raised

## Solution
For the "multi-unit tests", do **not** `raise_on_error` during the deploy, and wait for idle+active

## Release Notes
Don't blow up integration tests if IPU/IPA/IPR apps from o11y fail common hooks during deploy due to resourcepatch


PietroPasotti:
Funny how the itests are failing... 
The changes do LGTM though.

rbarry82:
The itests failing on, each time, random charms with semi-random failures looks like the Juju bug that's apparently been happening for 2 weeks or so. The patch was intended to stop the CI from turning read on semi-random "`[stop]` or `[leader-elected]`" execution, which it did, and now Juju is kind of losing track of pods somehow.

I updated a little just to "stick" to a channel (even though it's probably/hopefully `latest/candidate` anyway) to maybe avoid the behavior and the need to re-run flaky integration tests until it releases, especially since I don't have Charmhub permissions to push, even if I build locally. 

If the failures were consistently one charm (IPU, IPA, IPR) or consistently the same hook, I'd look a lot deeper, but the entirely random nature of the failures as well as knowing there's a Juju bug around this right now (and that most of our charms don't actually have any changes which would matter for this anyway) has me trying a bit for a bandaid/workaround for CI which won't compromise what CI is actually testing.

Abuelodelanada (approved):

PR #100, opened 2022-11-02
Add TLS termination support

### LINKED ISSUE #6: TLS termination by Traefik
Traefik should provide an option to expose a TLS-protected endpoint. Ideally, this would occur when Traefik is related to a charm providing certificates, like Vault.

sed-i:
## Issue
Currently all ingressed web traffic is unencrypted.
Now that we have the tls-certificates [library](https://github.com/canonical/tls-certificates-interface/blob/main/lib/charms/tls_certificates_interface/v1/tls_certificates.py) and [interface spec](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/tls_certificates/v1) it is time to add tls support to the traefik charm.

## Solution
- Have the charm code:
  - keep allowing unencrypted traffic (:80) even when TLS is available
  - always generate a TLS config block
  - providing TLS using a certificate obtained via the tls-certificates interface.
- itest using the [tls-certificates-operator](https://github.com/canonical/tls-certificates-operator) as the root ca.

Fixes #6.

Drive-by fixes:
- Update traefik image version from 2.7.0 to 2.9.6.
- Set workload version.

## Context
NTA.

## Testing Instructions
Afaict, the new itest + utest should provide sufficient coverage.

## Release Notes
Add TLS support.


src/charm.py
```diff
@@ -707,4 +810,4 @@ def _get_relation_type(relation: Relation) -> _IngressRelationType:
 
 
 if __name__ == "__main__":
-    main(TraefikIngressCharm)
+    main(TraefikIngressCharm, use_juju_for_storage=True)
```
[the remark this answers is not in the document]
Abuelodelanada: No, that was certainly a mistake, reverting..

src/charm.py
```diff
@@ -683,6 +773,19 @@ def _routing_mode(self) -> _RoutingMode:
         """
         return _RoutingMode(self.config["routing_mode"])
 
+    @property
+    def cert_subject(self) -> str:
+        """Provide certificate subject."""
+        # TODO: Use a good default for when the external_hostname is None
+        return self.model.config.get("external_hostname", "whatever")
```
sed-i, line 780: Iiuc, the only concern with `self.external_url` is that it may return an IP address.

If that is the case, would this work?

```suggestion
        return self.model.config.get("external_hostname", socket.getfqdn())
```

cc: @rbarry82

sed-i:
I [modified](https://github.com/canonical/traefik-k8s-operator/pull/100/commits/e0bc097de8fb7cb30ff92bcda657b069f80274cf) the config a bit and now https works:
```
$ jaddr trfk  
10.43.8.80

$ cat /etc/hosts 
127.0.1.1 charm-dev-2cpu-7g charm-dev-2cpu-7g
127.0.0.1 localhost
10.43.8.80 juju.local

$ curl --insecure http://juju.local/welcome-am/api/v1/alerts   # this is alertmanager
{"status":"success","data":[]}
```

The problem is that it works without any certificates relation in place. That is because in the config I enabled traefik to generate its own self-signed default certificate, which I can see with [`seecert`](https://stackoverflow.com/a/59412853/3516684):
```
$ seecert juju.local
Server:		127.0.0.53
Address:	127.0.0.53#53

Name:	juju.local
Address: 10.43.8.80

depth=0 CN = TRAEFIK DEFAULT CERT
verify error:num=18:self-signed certificate
verify return:1
depth=0 CN = TRAEFIK DEFAULT CERT
verify return:1
DONE
        Validity
            Not Before: Dec  9 23:14:18 2022 GMT
            Not After : Dec  9 23:14:18 2023 GMT
```

Should this mode be at all enabled for when no certificates are related?

cc: @rbarry82 @simskij

sed-i:
Any takers for a first pass of the draft?
@lucabello @Abuelodelanada @PietroPasotti @rbarry82

src/charm.py
```diff
@@ -120,11 +168,55 @@ def __init__(self, *args):
         observe(ipu_events.data_provided, self._handle_ingress_data_provided)
         observe(ipu_events.data_removed, self._handle_ingress_data_removed)
 
-        observe(self.traefik_route.on.ready, self._handle_traefik_route_ready)
+        ipr_events = self.traefik_route.on
```
rbarry82, line 171: "ipr" (ingress per route?) isn't a very meaningful name considering what the charm/relation actually do. Could we change this to `tr_events` or `route_events`?

src/charm.py
```diff
@@ -183,41 +275,72 @@ def _push_static_config(self):
...
+            "log": {
+                "level": "DEBUG",
+            },
+            "entryPoints": {
+                "diagnostics": {"address": f":{self._diagnostics_port}"},
+                # "web": {
+                #     "address": f":{self._port}",
+                #     'http': {
+                #         'redirections': {'entryPoint': {'to': 'websecure', 'scheme': 'https'}}}
+                # },
```
rbarry82, line 288: Do we need this commented block?

src/charm.py
```diff
@@ -675,6 +845,38 @@ def _routing_mode(self) -> _RoutingMode:
...
+
+        version_output, _ = self.container.exec([self._bin_path, "version"]).wait_output()
+        # Output looks like this:
+        # Version:      2.9.6
+        # Codename:     banon
+        # Go version:   go1.18.9
+        # Built:        2022-12-07_04:28:37PM
+        # OS/Arch:      linux/amd64
+
+        if result := re.search(r"Version:\s*(\d*\.\d*\.\d*)", version_output):
```
rbarry82, line 862: `re.search(r'Version:\s+((\d+,?){3,})')`

src/charm.py
```diff
@@ -675,6 +845,38 @@ def _routing_mode(self) -> _RoutingMode:
...
+            self.unit.set_workload_version(version)
+        else:
+            logger.debug(
+                "Cannot set workload version at this time: could not get Traefik version."
+            )
+
+    @property
+    def cert_subject(self) -> str:
+        """Provide certificate subject."""
+        # TODO: Use a good default for when the external_hostname is None
```
rbarry82, line 877: A good default is probably "don't enable TLS if the prerequisite configuration isn't present." Having a subject with a cluster-local DNS name is gonna be a bad time for anyone who deploys Traefik, doesn't configure this, and their browser upgrades a connection to HTTPS only to immediately see an invalid cert.

src/charm.py
```diff
@@ -626,7 +793,7 @@ def _restart_traefik(self):
                 _TRAEFIK_SERVICE_NAME: {
                     "override": "replace",
                     "summary": "Traefik",
-                    "command": "/usr/bin/traefik",
+                    "command": self._bin_path,
```
rbarry82, line 796: Is there a reason to change this? Sure, `_bin_path` is also re-used in `.version`, but given that it's only in a couple of places, leaving it explicit is 50/50.

src/charm.py
```diff
@@ -76,21 +98,35 @@ class TraefikIngressCharm(CharmBase):
...
+        # ports = [
+        #     ServicePort(self._port, targetPort=self._port, name=f"{self.app.name}"),
+        #     ServicePort(self._tls_port, targetPort=self._tls_port, name=f"{self.app.name}-tls"),
+        # ]
+        # self.service_patcher = KubernetesServicePatch(self, ports, "LoadBalancer")
         self.service_patch = KubernetesServicePatch(
             charm=self,
             service_type="LoadBalancer",
-            ports=[(f"{self.app.name}", self._port)],
+            ports=[(f"{self.app.name}", self._port), (f"{self.app.name}-tls", 443)],
```
rbarry82, line 129: Not `self._tls_port`?

src/charm.py
```diff
@@ -354,10 +478,9 @@ def _handle_ingress_data_provided(self, event: RelationEvent):
 
     def _handle_ingress_data_removed(self, event: RelationEvent):
         """A unit has removed the data we need to provide ingress."""
-        # if this is because the relation is gone, we don't need to do anything
-        if isinstance(event, RelationBrokenEvent):
-            return
-        self._wipe_ingress_for_relation(event.relation)
+        self._wipe_ingress_for_relation(
+            event.relation, wipe_rel_data=not isinstance(event, RelationBrokenEvent)
```
rbarry82, line 482: Are we sure we want to remove it on a `RelationDepartedEvent` and leave it on a `RelationBrokenEvent`?

src/charm.py
```diff
@@ -491,7 +630,7 @@ def _generate_middleware_config(
 
     def _generate_per_unit_config(self, data: "RequirerData_IPU") -> Tuple[dict, str]:
         """Generate a config dict for a given unit for IngressPerUnit."""
-        config = {"http": {"routers": {}, "services": {}}}  # type: Dict[str, Any]
+        # config = {"http": {"routers": {}, "services": {}}}  # type: Dict[str, Any]
```
rbarry82, line 633: Why not just remove it?

src/charm.py
```diff
@@ -56,8 +70,16 @@
...
-_CONFIG_DIRECTORY = "/opt/traefik/juju"
-_STATIC_CONFIG_PATH = "/etc/traefik/traefik.yaml"
+_DYNAMIC_CONFIG_DIR = "/opt/traefik/juju"
+_STATIC_CONFIG_DIR = "/etc/traefik"
+_STATIC_CONFIG_PATH = _STATIC_CONFIG_DIR + "/traefik.yaml"
+_DYNAMIC_CERTS_PATH = _DYNAMIC_CONFIG_DIR + "/certificates.yaml"
+
+# FIXME:
+#  1. Is it ok to store certs/keys in the dynamic config path?
+#  2. If overwritten (in /etc/...), would certs/keys be re-read without certificates.yaml changing?
```
rbarry82, line 80: AFAIK, yes, but we can probably test this and watch the logs.

src/charm.py
```diff
@@ -120,11 +168,55 @@ def __init__(self, *args):
...
 
         # Action handlers
         observe(self.on.show_proxied_endpoints_action, self._on_show_proxied_endpoints)
 
+    def _on_install(self, event) -> None:
+        # Generate key without a passphrase as traefik does not support it
+        # https://github.com/traefik/traefik/pull/6518
+        private_key = generate_private_key()
+        self._stored.private_key = private_key.decode()
+        # FIXME this (occasionally?!) doesn't survive a charm upgrade, and results in:
```
rbarry82, line 183: Despite the docs in the library, the [Juju docs](https://juju.is/docs/sdk/install-event) are pretty clear about the caveats here. This may be much better as `on_start` -- check whether `self._stored.private_key` evaluates to true, and if it doesn't, generate a key.

rbarry82 (commented):
LGTM in general, just some questions.

src/charm.py
sed-i: Modified the logic here a bit. Wdyt?

src/charm.py
```diff
@@ -125,6 +157,48 @@ def __init__(self, *args):
...
+        self._stored.csr = csr.decode()
+        self.certificates.request_certificate_creation(certificate_signing_request=csr)
+
+    def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
+        self._stored.certificate = event.certificate
+        self._stored.ca = event.ca
+        self._stored.chain = event.chain
+        # TODO: Store files in container and modify config file
+        self.container.push(_CERTIFICATE_PATH, self._stored.certificate, make_dirs=True)
+        self.container.push(_CERTIFICATE_KEY_PATH, self._stored.private_key, make_dirs=True)
```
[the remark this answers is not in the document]
rbarry82: Possibly/probably yes. This is more of a PM question, but considering the essentials of certs and subdomains (n >= 2 total domains/subdomains requires multiple certs), it's probably a safe bet to design it around that assumption, make `self._stored....` values for certs into dicts, and drag `_type_convert_stored` in here, too and start from a position of flexibilty rather than needing to try to potentially migrate it later.

src/charm.py
```diff
@@ -125,6 +157,48 @@ def __init__(self, *args):
...
+        self._push_config()
+        self._process_status_and_configurations()
+
+    def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
+        old_csr = self._stored.csr
+        private_key_password = self._stored.private_key_password
+        private_key = self._stored.private_key
+        new_csr = generate_csr(
```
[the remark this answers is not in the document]
rbarry82: There don't seem to be any guards in the library, but the [module itself definitely can](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.load_pem_private_key). An aggregate for `TypeError | ValueError | cryptography.exceptions.UnsupportedAlgorithm` looks like it should cover enough footguns, and should probably just be in the lib itself. @ghislainbourgeois thoughts?

src/charm.py
rbarry82: I actually fat-fingered the suggestion, which should have been:

```python
re.search(r'Version:\s+((\d+\.?){3,})') # note: \. instead of , 
```

But, nominally, these will also match the current regexp:

```
Version:123%456$789^
Version:     123456$7%9#
```

If we don't care how many groups, it could be `...((\d+\.?)+)`. If we want at least 3, then `{3,}`. Other than that, we probably _do_ to make sure that it's `..\s+((\d+\.?){1,}...)` or `..\s+((\d+\.?)+...)`as "at least one whitespace character, followed by at least one block of numbers.

If we really, really wanted to future proof it, just make it:

```python
if result := re.search(r'Version:\s+(?P<version>(\d+\.?){0,}(\d+$))', version_output):
  return re.groupdict()["version"]
```

There's some examples of potentially "valid" and "invalid" matches in a tester [here](https://regex101.com/r/LaG1ey/1)

src/charm.py
rbarry82: :+1: Testing the other path (if a `LoadBalancer` has a hostname set) will almost need to be a little ugly in integration tests (it's mostly "cloud" load balancers which set this), but that's probably ok.

src/charm.py
rbarry82: The counterpoint to this is that variables which are used in very few places may require the author to mentally "bounce" around in the codebase to keep track of where they are, and it's a hardcoded literal there also.

Considering that "upstream" images may use, say, `/bin/foo` and ROCKs are almost certain to use `/usr/bin/foo`, if we're going to have something in the code (as a property or method, rather than attribute), it may as well wait for the container to be up and actually check the filesystem for where the path is.

It's more of a nit than anything, but a class-level private attribute for a hardcoded string may as well be a constant like other container paths (`_STATIC_CONFIG_DIR`, for example, though I have no idea why _that_ is private either), leave it hardcoded, or actually look in the container for it.

src/charm.py
rbarry82: Sure, we can test that, but from the doc:

* Typically, **operations performed on install should also be considered for [upgrade-charm](https://juju.is/docs/sdk/upgrade-charm-event).**
* In some cases, [config-changed](https://juju.is/docs/sdk/config-changed-event) can be used instead of install and upgrade-charm because it is guaranteed to fire after both.

It seems to be "expected" that install event may only fire once. It's also possible that the relation join event is simply triggering first. 

* The events in `on_install` may simply be better in `on_start` (and maybe also in `on_upgrade_charm`)
* The checks in `on_certificates_relation_joined` should guard against the value being `None`
* The charm should check whether a private key already exists _before_ blindly generating one (which may change)
* And, really, we may just want to move the whole thing to the Juju Secrets API. Those patches are in OF now.

Lastly, the guess in the comments (about using Juju storage) is probably a case of trying to guess unknowable runtime state from a more or less stateless workflow (since the charm doesn't know which events have already run since startup). In the cases where you upgrade and you see this, what's the ordering of events?

`traefik-k8s` also doesn't say _not_ to use Juju storage. [If it's not set](https://github.com/canonical/operator/blob/3d0018019988a7a81470466b0bd8786835d46be3/ops/main.py#L399), OF tries to figure it out for you. If local storage has never been used (likely true), `"kubernetes"` is not in the series (true here -- no series is set), and the controller version is new enough (also true), `use_juju_for_storage` will default to `True`. We probably _are_ already using it, and that can be verified by looking for [this log message](https://github.com/canonical/operator/blob/3d0018019988a7a81470466b0bd8786835d46be3/ops/main.py#L352).

What events are being fired?

src/charm.py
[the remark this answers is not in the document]
sed-i: (I forgot to resolve this comment.)
Iirc we discussed this a while ago and decided there will be only one and also added `llimit: 1` to metadata.
Afaiu, Traefik's `external_host` is just one host, and any subdomains could be added to the same CSR.
Am I missing anything here?

src/charm.py
sed-i: Nice regex-foo!
I think you convinced me to change to `Version:\s*(.+)`.

src/charm.py
sed-i: Moved to module level attr for consistency.
Let's leave the `_` and the remaining class attributes for another time.

src/charm.py
sed-i: > `use_juju_for_storage` will default to `True`

Good point.

I also haven't encountered this error since I put in the comment.
I'll try to write a few utests around this and we'll take it from there.

sed-i: Confirmed that we need to manually set `use_juju_for_storage=True` (see https://github.com/canonical/operator/issues/880).

tests/integration/test_traefik_deployed_after_metallb.py
```diff
@@ -82,14 +100,87 @@ async def test_build_and_deploy(ops_test: OpsTest, traefik_charm):
...
+
+    ip = await get_address(ops_test, trfk.name)
+    await curl_endpoints(ops_test, temp_dir, cert_path, ip)
+
+
+@pytest.mark.abort_on_fail
+async def test_tls_termination_after_charm_upgrade(ops_test: OpsTest, traefik_charm, temp_dir):
+    logger.info(
+        "Refreshing charm to test TLS termination still works with the same certificate after"
+        " charm upgrade..."
```
sed-i, line 179: Not sure this is the desired behavior: does it matter that traefik's cert changed?
As long as the root-ca cert is the same, https should keep working.

rbarry82: This is a bad assumption. Unless the user actually imports the root CA into the store (browser or otherwise), most browsers these days do (or did, probably still do -- there were lots of debates about it in the Mozilla community) bring up a warning page again if a self-signed cert changed for any reason.

sed-i: Ok sure we can keep the current behavior then. We can deal with stored state later.

rbarry82 (approved):

tests/integration/test_traefik_deployed_after_metallb.py
dstathis: I see no issue with having to re-accept a self signed cert. This seems like an acceptable burden if you are using self signed certs

requirements.txt
```diff
@@ -1,6 +1,30 @@
...
 lightkube >= 0.8.1
 lightkube-models >= 1.22.0.4
-# The following is needed by lib/charms/traefik_k8s/*
-jsonschema
+
+# Operator Framework
+# Code: https://github.com/canonical/operator/
+# Docs: https://ops.rtfd.io/
+# Deps: charm
+ops
```
dstathis, line 30: Love the helpful comments here!

src/charm.py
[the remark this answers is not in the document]
rbarry82: The real question being that, like an earlier discussion, you can't have a wildcard for `*.example.org` which also works for `*.foo.example.org`, and they can't be part of the same CSR even with a wildcard. 

We also decided on "one Traefik per subdomain for now, but we may as well make it a little flexible just in case that needs to change so we don't have to break API if it comes up later".

For safety, a comma-separated list of domains (which could be merged with the domain of `external_host`) which are kept in storedstate, and making `_stored.private_key | certificate` into lists/dicts would possibly save a lot of potential pain in the future.

tests/integration/test_traefik_deployed_after_metallb.py
rbarry82: It is, generally. There's a case here where, because Traefik will use an auto-generated self-signed cert if it doesn't have any, connecting to Traefik will give you an entirely different cert chain, then relating to a TLS provider (which grants a cert) will override the entire cert chain (signing/root CAs and all).

Since we aren't using a middleware to upgrade the connection, users would kind-of need to opt-in (unless they deployed an application which set HSTS, but then that's on them), so programmatic clients _probably_ wouldn't care. _Most_ clients (urllib, requests, tls.Client in golang, etc don't cache certs).

dstathis:
I tried to deploy this and ran in to some issues:

```
dylan@protostar:~/traefik$ juju deploy tls-certificates-operator --config generate-self-signed-certificates=true --config ca-common-name='example.com' tls-certs --channel edge
Located charm "tls-certificates-operator" in charm-hub, revision 20
Deploying "tls-certs" from charm-hub charm "tls-certificates-operator", revision 20 in channel edge on focal
dylan@protostar:~/traefik$ juju deploy ./traefik-k8s_ubuntu-20.04-amd64.charm tfk --trust --resource traefik-image=docker.io/jnsgruk/traefik:2.9.6
Located local charm "traefik-k8s", revision 1
Deploying "tfk" from local charm "traefik-k8s", revision 1 on focal
dylan@protostar:~/traefik$ juju relate tfk tls-certs
```
Which produced the following error
```
unit-tfk-0: 08:24:17 ERROR unit.tfk/0.juju-log certificates:6: Uncaught exception while in charm code:
  File "./src/charm.py", line 916, in <module>
    main(TraefikIngressCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/ops/main.py", line 438, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/ops/main.py", line 150, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/ops/framework.py", line 355, in emit
    framework._emit(event)  # noqa
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/ops/framework.py", line 856, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/ops/framework.py", line 931, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 186, in _on_certificates_relation_joined
    csr = generate_csr(
  File "/var/lib/juju/agents/unit-tfk-0/charm/lib/charms/tls_certificates_interface/v1/tls_certificates.py", line 799, in generate_csr
    subject_name = [x509.NameAttribute(x509.NameOID.COMMON_NAME, subject)]
  File "/var/lib/juju/agents/unit-tfk-0/charm/venv/cryptography/x509/name.py", line 131, in __init__
    raise TypeError("value argument must be a str")
TypeError: value argument must be a str
unit-tfk-0: 08:24:17 ERROR juju.worker.uniter.operation hook "certificates-relation-joined" (via hook dispatching script: dispatch) failed: exit status 1
```

sed-i:
Thanks @dstathis. This happened because there was no `juju config traefik external_hostname=demo.local` set, or because metallb returned an IP and not a hostname.
Added guard + utest but the plot thickened quickly when manually tested `config-changed` and `relate` ordering. Need to rework the charm a bit.

I'm glad you tested!

rbarry82:
A unit test may not be meaningful, but we're possibly overcomplicating this, and there's more than one failure case. Even if it were an IP, [`_get_loadbalancer_status`](https://github.com/canonical/traefik-k8s-operator/blob/43679447b3b28bef252cf5cd088a2d6df65427f4/src/charm.py#L900) returns properties [which are strings](https://gtsystem.github.io/lightkube-models/1.26/models/core_v1/index.html#loadbalanceringress). That would, from a code flow perspective, nominally pass muster as a `subject_name` (even if it's wrong). The problem is that [`cert_subject` just blindly returns `None`](https://github.com/canonical/traefik-k8s-operator/blob/43679447b3b28bef252cf5cd088a2d6df65427f4/src/charm.py#L889) if it _is_ an IP and not a DNS name. 

Neither side of the `TLSCertificates[Provider|Requirer]` actually cares about `relation_joined`, only `relation_changed`. 

We could/should guard against `None` in `_on_certificates_relation_joined`.

Making it:

```python
def _on_certificates_relation_joined(self, event: RelationJoinedEvent) -> None:
    self._check_csr()
      
def _check_csr(self):
    if not self.cert_subject:
        logger.debug("No valid hostname is available for a TLS certificate subject")
        return
    if self._stored.csr:
        # CSR already exists
        return
    private_key = self._stored.private_key
    csr = generate_csr(
        private_key=private_key.encode("utf-8"),
        subject=self.cert_subject,
     )
    self._stored.csr = csr.decode()
    self.certificates.request_certificate_creation(certificate_signing_request=csr)
```

This could be called from `_process_status_and_configurations`. It's idempotent, doesn't care about event ordering, won't override existing CSRs. As an aside, we should probably also blank `self._stored.csr` [here](https://github.com/canonical/traefik-k8s-operator/blob/43679447b3b28bef252cf5cd088a2d6df65427f4/src/charm.py#L195).

As an aside, we can also do better about `cert_subject`. It's way too prescriptive right now.

For comparison:

```
kubectl get svc -n traefik -o wide
NAME                TYPE           CLUSTER-IP       EXTERNAL-IP    PORT(S)                      AGE    SELECTOR
traefik             LoadBalancer   10.100.2.107     192.168.2.1    80:30149/TCP,443:31746/TCP   406d   app.kubernetes.io/instance=traefik,app.kubernetes.io/name=traefik
```

```
dig grafana.k8s.rbarry.org

; <<>> DiG 9.10.6 <<>> grafana.k8s.rbarry.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42919
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 1, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;grafana.k8s.rbarry.org.                IN      A

;; ANSWER SECTION:
k8s.rbarry.org.         60      IN      DNAME   k8s.rbarry.priv.
grafana.k8s.rbarry.org. 60      IN      CNAME   grafana.k8s.rbarry.priv.
grafana.k8s.rbarry.priv. 0     IN      A       192.168.2.1

;; AUTHORITY SECTION:
k8s.rbarry.priv.       60      IN      NS      ns.k8s.rbarry.priv.

;; ADDITIONAL SECTION:
ns.k8s.rbarry.priv.    60      IN      A       192.168.0.3
```

```
dig -x 192.168.2.1

; <<>> DiG 9.10.6 <<>> -x 192.168.2.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45607
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;1.2.168.192.in-addr.arpa.     IN      PTR

;; ANSWER SECTION:
1.2.168.192.in-addr.arpa. 3600 IN      PTR     grafana.k8s.rbarry.priv.

;; AUTHORITY SECTION:
2.168.192.in-addr.arpa. 86400   IN      NS      ns.k8s.rbarry.priv.

;; ADDITIONAL SECTION:
ns.rbarry.priv.        86400   IN      A       192.168.0.3

;; Query time: 59 msec
;; SERVER: 192.168.0.3#53(192.168.0.3)
;; WHEN: Sat Jan 14 14:33:33 EST 2023
;; MSG SIZE  rcvd: 118
```

```
>>> socket.gethostbyaddr("192.168.2.1")                                                                                                                                                                                                            
('grafana.k8s.rbarry.priv', ['1.2.168.192.in-addr.arpa'], ['192.168.2.1'])

This is: kube doesn't show me a hostname for the external IP. But it has a DNS record, and reverse DNS, and Python can get it. So...

```python
    @property
    def cert_subject(self) -> Optional[str]:
        """Provide certificate subject."""
        host_or_ip = self.external_host
        try:
            ipaddress.ip_address(host_or_ip)
        except ValueError:
            # This is not an IP address so assume it's a hostname that can be used as the subject.
            # Note: a ValueError will be raised if `host_or_ip` is None, which is ok here.
            return host_or_ip
        else:
            # This is an IP address, which cannot be used for the subject. Try to look it up.
            try:
                return socket.gethostbyaddr(ipaddr)[0]
            except OSError:
                # We do not want to return `socket.getfqdn()` because the user's browser would
                # immediately complain about an invalid cert. If we can't resolve it via any method,
                # return None
                return None
```

The change to `_on_certificates_relation_joined` (to guard against `None` and make sure we call it from `_process...` so it doesn't get missed) is all that's needed to more or less permanently resolve this. Trying to lookup the address is just a nicety, but a technically simple one.

sed-i:
Excellent input @rbarry82. Incorporated in next commit.

rbarry82:
Failure not related -- just a `charmcraft pack ...` timeout. Re-running

PR #107, opened 2022-11-25
Refresh url on access

sed-i:
## Issue
https://github.com/canonical/catalogue-k8s-operator/issues/3

## Solution
Refresh url on access ([example](https://github.com/canonical/traefik-route-k8s-operator/commit/905b1b8897740ab1e7ea136304c137c0edf4cf82#diff-b09930835d29e4f97b9b799ee47942218e84c948f74389576c52586657d8a6a0R163)) by pre-fetching urls from relation data instead of stored state.

## Context
See thread: https://github.com/canonical/catalogue-k8s-operator/pull/4#issuecomment-1326823357

## Testing Instructions
Deploy the bundle and see which URL catalogue has for prom: fqdn or ingress.

```bash
$ ./render_bundle.py local.yaml --traefik ../traefik-k8s-operator/traefik-k8s_ubuntu-20.04-amd64.charm
$ juju deploy ./local.yaml --trust
$ juju ssh --container catalogue catalogue/0 cat /web/config.json | jq '.apps'
```

## Release Notes
Prefetch urls from reldata instead of stored state.


tests/unit/test_lib_per_unit_requires.py
```diff
@@ -199,3 +199,49 @@ def test_ipu_events(self):
         # AND a revoked event fires
         after = self.harness.charm.revoked_event_count
         self.assertGreater(after, before)
+
+
+class TestInterlibDependency(unittest.TestCase):
```
sed-i, line 204: @rbarry82 could you please take a look at this test?
Is that what you had in mind on the [catalogue PR](https://github.com/canonical/catalogue-k8s-operator/pull/4#issuecomment-1326823357)?

rbarry82: Not exactly -- to make it reflective of what I had in mind (or what traefik-route is already doing), `ingress_per_unit` would need changes.

Put differently, if you [walk the code back](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v1/ingress_per_unit.py#L840), `_get_urls_from_relation_data` still will not be called (nor will it update `.url`) until an event comes through. And event which would trigger the observer _after_ it's resolved, because `copy(x)` still doesn't create a closure anyway IIRC.

sed-i: > `ingress_per_unit` would need changes.

Yes of course. This is just a failing test - I haven't started touching the code.
I meant: does this test, as written, capture the use case you had in mind, for which I'd then need to change the IPU code so the test passes.

rbarry82: Ah, in that case, mostly yes. I guess that when I think of "inter-lib dependency", I would still think of it as simply:

```python
self.dependee = { "value": self.ipu.url }
```

Without the `copy()`, but it's not super important

sed-i: The copy is intentional, to convey that we do not rely on passing object reference.

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -832,16 +831,16 @@ def urls(self) -> Dict[str, str]:
 
         May return an empty dict if the URLs aren't available yet.
         """
-        data = _type_convert_stored(self._stored.current_urls or {})  # type: ignore
```
PietroPasotti, line 835: can we get rid of stored altogether?

PietroPasotti (commented):
couple of food-for-thought points, but it looks good.

PietroPasotti (approved):

lib/charms/traefik_k8s/v1/ingress_per_unit.py
sed-i: Currently it's used to compare current vs. previous for emitting the custom events.
Do you mean use peer data instead?

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -832,16 +831,16 @@ def urls(self) -> Dict[str, str]:
...
-        assert isinstance(data, dict)  # for static checker
-        return data
+        current_urls = self._get_urls_from_relation_data
```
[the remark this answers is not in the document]
sed-i: Done in next commit.

lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -832,16 +831,16 @@ def urls(self) -> Dict[str, str]:
...
         """The full ingress URL to reach the current unit.
 
         May return None if the URL isn't available yet.
         """
-        if not self.urls:
+        urls = self.urls
+        if not urls:
             return None
-        return self.urls.get(self.charm.unit.name)
+        return urls.get(self.charm.unit.name)
```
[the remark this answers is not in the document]
sed-i: If we could use walrus then I'd use walrus.
It's to avoid fetching the urls property twice.
It mattered more before I refactored.

PR #115, opened 2023-01-04
alert rules for self-monitoring

Abuelodelanada:
## Issue

This PR add a new alert rule and fix an older one

## Solution

The new alert rule we are adding checks whether the traefik config reload fails or not.

## Testing Instructions

Pack the charm, deploy and relate it with prometheus charm.

### DIFF
--- a/src/prometheus_alert_rules/reload_failure.rule
+++ b/src/prometheus_alert_rules/reload_failure.rule
@@ -5,4 +5,4 @@ labels:
   severity: critical
 annotations:
   summary: Traefik config reload failure (instance {{ $labels.instance }})
-  description: "Traefik config reload failure rate is above 0"
+  description: "Traefik keeps failing to reload its config file"
--- a/src/prometheus_alert_rules/reload_failure.rule
+++ b/src/prometheus_alert_rules/reload_failure.rule
@@ -1,0 +1,8 @@
+alert: TraefikConfigReloadFailure
+expr: rate(traefik_config_reloads_failure_total[3m]) > bool 0
+for: 1m
+labels:
+  severity: critical
+annotations:
+  summary: Traefik config reload failure (instance {{ $labels.instance }})
+  description: "Traefik config reload failure rate is above 0"
--- a/src/prometheus_alert_rules/unit_unavailable.rule
+++ b/src/prometheus_alert_rules/unit_unavailable.rule
@@ -1,5 +1,5 @@
 alert: TraefikIngressUnitIsUnavailable
-expr: up{%%juju_topology%%} < 1
+expr: up < 1
 for: 0m
 labels:
   severity: critical


src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -1,5 +1,5 @@
 alert: TraefikIngressUnitIsUnavailable
-expr: up{%%juju_topology%%} < 1
```
PietroPasotti, line 2: why was this here in the first place?

PietroPasotti (approved):

src/prometheus_alert_rules/unit_unavailable.rule
Abuelodelanada: Because in ancient times we used `%%juju_topology%%` to be replaced with the actual topology... now the topology is injected and this "tag" is no longer necessary

PietroPasotti: I thought that this kind of tag replacement is precisely what "injection" means. Maybe I need to dig up a bit better how that works

Abuelodelanada: You can take a look at [this class](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L2252) and [this tool](https://github.com/canonical/cos-tool/)

src/prometheus_alert_rules/reload_failure.rule
```diff
@@ -0,0 +1,8 @@
+alert: TraefikConfigReloadFailure
+expr: rate(traefik_config_reloads_failure_total[3m]) > bool 0
+for: 1m
```
sed-i, line 3: Might be handy to have a docstring for the rationale.
For example, why not `increase(traefik_config_reloads_failure_total[3m]) > 0` `for 1m`?

src/prometheus_alert_rules/reload_failure.rule
```diff
@@ -0,0 +1,8 @@
...
+labels:
+  severity: critical
+annotations:
+  summary: Traefik config reload failure (instance {{ $labels.instance }})
+  description: "Traefik config reload failure rate is above 0"
```
sed-i, line 8: I think "failure rate is above 0" could be made more grokable for a busy on-caller. How about:

```suggestion
  description: "Traefik keeps failing to reload its config file"
```

(At some point we should write a styleguide for alerts.)

sed-i (approved):

src/prometheus_alert_rules/reload_failure.rule
Abuelodelanada: > (At some point we should write a styleguide for alerts.)

You are reading my mind @sed-i !

I think we will also at some point need to curate our alerts and dashboards to have a unified UX.

src/prometheus_alert_rules/reload_failure.rule
Abuelodelanada: I like it, seems more explicit than my initial version. Thanks!

PR #117, opened 2023-01-09
catch charmcraft pack errors in spellbook

PietroPasotti:
## Issue
spellbook's `do_build` was naively assuming `charmcraft pack` would succeed.
Also changed tox.ini so that linting is only ran on `src/`, because I was getting errors from charm libs (not owned by traefik)

## Solution
Wrapped the command in a try/catch that aborts on nonzero return codes, raising a helpful message.

## Testing Instructions
Hope the CI passes

## Release Notes
* fixed bug in spellbook cache

### DIFF
--- a/tests/integration/spellbook/cache.py
+++ b/tests/integration/spellbook/cache.py
@@ -6,7 +6,7 @@ import shutil
 from datetime import datetime
 from hashlib import md5
 from pathlib import Path
-from subprocess import getoutput
+from subprocess import CalledProcessError, check_output, getoutput
 from typing import List, Union
 
 import yaml
@@ -40,7 +40,7 @@ def _get_libpath(base, source):
     return root.absolute()
 
 
-def spellbook_fetch(  # ignore: C901
+def spellbook_fetch(  # noqa: C901
     charm_root: Union[str, Path] = "./",
     charm_name: str = None,
     hash_paths: List[Path] = None,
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -49,7 +49,7 @@ commands =
     codespell . --skip .git --skip .tox --skip build --skip lib --skip venv* \
       --skip .mypy_cache --skip icon.svg
     # pflake8 wrapper supports config from pyproject.toml
-    pflake8 {[vars]all_path}
+    pflake8 {[vars]src_path} {[vars]lib_path}
```
sed-i, line 52: I think this should stay `{[vars]all_path}`.
What errors were you seeing?

sed-i (approved):

tox.ini
PietroPasotti: after you've run integration tests, the tester charms contain (for example) traefik_route.py, which fails linting.
So I guess it depends on the order in which you run the tests

we could selectively exclude all non-traefik charm libs from the tester charms, but that felt like too much work.

PietroPasotti: (or we could fix linting in traefik_route, but I guess the general issue of linting third party charm libs remains)

sed-i: Already [fixed](https://github.com/canonical/traefik-route-k8s-operator/pull/25).
But on github they run in isolation so it probably won't fail..?

PietroPasotti: oh. Yeah I'll revert then. Tomorrow first thing

PietroPasotti: done

tests/integration/spellbook/cache.py
```diff
@@ -75,8 +75,14 @@ def spellbook_fetch(  # ignore: C901
 
     def do_build():
         logging.info(f"building {charm_root}")
-        pack_out = getoutput(f"charmcraft pack -p {charm_root}")
-        return (Path(os.getcwd()) / pack_out.split("\n")[-1].strip()).absolute()
+        try:
+            pack_out = check_output(("charmcraft", "pack", "-p", str(charm_root)))
+        except CalledProcessError as e:
+            raise RuntimeError(
+                "Charmcraft pack failed. " "Attempt a `charmcraft clean` or inspect the logs."
```
sed-i, line 82:
```suggestion
                "Charmcraft pack failed. Attempt a `charmcraft clean` or inspect the logs."
```

sed-i (approved):

PR #104, opened 2022-11-21
observability

PietroPasotti:
This PR adds loki-logging and grafana-dashboarding integrations to the traefik charm.

## Testing Instructions
deploy cos-lite, then refresh the traefik charm with one built from these sources. Visit the grafana frontend to see the traefik dashboard with the logs.

## Release Notes
- Added loki-logging and grafana-dashboarding integrations.

PR #120, opened 2023-01-10
add new Loki based alert rule

Abuelodelanada:
## Issue

Based on Traefik log format, this PR adds a Loki based alert rule that will fire if a `level=error` string is found.

The alert rule will work once PR #104 (or a PR that replace it) is finally merged.

For the time being, this file is harmless to the functionality of charm.

### DIFF
--- a/src/loki_alert_rules/error_level_logs_found.rule
+++ b/src/loki_alert_rules/error_level_logs_found.rule
@@ -1,0 +1,7 @@
+alert: ErrorLevelLogsFound
+expr: rate({%%juju_topology%%} |= "level=error"[1m]) > 0
+for: 1m
+labels:
+  severity: critical
+annotations:
+  summary: error level logs found.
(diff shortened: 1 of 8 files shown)


PietroPasotti (approved):

PietroPasotti:
failing CI -> snappy outage

src/loki_alert_rules/error_level_logs_found.rule
```diff
@@ -0,0 +1,7 @@
+alert: ErrorLevelLogsFound
+expr: rate({%%juju_topology%%} |= "level=error"[1m]) > 0
```
sed-i, line 2: Shall we start using the `job~=.+` thingy instead of `%%juju_topology%%`?

sed-i (approved):

src/loki_alert_rules/error_level_logs_found.rule
Abuelodelanada: Not until we make a decision [on this](https://github.com/canonical/loki-k8s-operator/issues/229).

sed-i: My impression was that it's independent: we don't need `%%juju_topology%%`, and with it it won't pass cos-tool validation, and with `job=~".+"` everything works, so might as well..?

PR #122, opened 2023-01-11
Fix missing relation app

rbarry82:
## Issue
Resolving `all((some, tuple))` or other is not a generator, is not lazily-evalutated, and will immediately throw an `AttributeError` at resolution if something cannot be found.

## Solution
Use a comprehension to provide a generator, and short-circuit on the first failure.

## Context
See [this comment](https://github.com/canonical/grafana-k8s-operator/issues/165#issuecomment-1375436662), reported on Grafana, but in this library.

## Release Notes
Guard for `relation.app = None` in `ingress_per_unit`. Move tox to 3.8

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v1/ingress_per_unit.py
@@ -82,7 +82,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
 
 log = logging.getLogger(__name__)
 
@@ -801,7 +801,7 @@ class IngressPerUnitRequirer(_IngressPerUnitBase):
         if not relation:
             return {}
 
-        if not all((relation.app, relation.app.name)):  # type: ignore
+        if not all([r for r in [relation.app, relation.app.name]]):  # type: ignore
             # FIXME Workaround for https://github.com/canonical/operator/issues/693
             # We must be in a relation_broken hook
             return {}
--- a/lib/charms/traefik_k8s/v1/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v1/ingress_per_unit.py
@@ -801,7 +801,7 @@ class IngressPerUnitRequirer(_IngressPerUnitBase):
         if not relation:
             return {}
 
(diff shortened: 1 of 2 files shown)


lib/charms/traefik_k8s/v1/ingress_per_unit.py
```diff
@@ -801,7 +801,7 @@ def _urls_from_relation_data(self) -> Dict[str, str]:
         if not relation:
             return {}
 
-        if not all((relation.app, relation.app.name)):  # type: ignore
+        if not all([r for r in [relation.app, relation.app.name]]):  # type: ignore
```
sed-i, line 804: I can easily imagine someone refactoring out the "redundancy".

sed-i (approved):

lib/charms/traefik_k8s/v1/ingress_per_unit.py
PietroPasotti: I see what you're doing but IMHO `if relation.app and relation.app.name` is so much more clear

PietroPasotti (changes requested):
there was a bug because there is complexity involved in a (negated) `all` and lazy expression evaluation.
So instead of fixing the bug by correcting the mistake, why don't we take this chance to simplify that unnecessary complexity and use a plain old boring `if a and a.b:`?
It feels like the original code was 'overengineered', so far as a single line can be, so I'd be happier if we strip things down a little bit :)
Makes things more maintainable. What if someone doesn't see the point of that (obscure) expression and factors it out? What if someone needs to add a check for `relation.app.uuid` or something else and doesn't realize that it needs to be inserted **after** `relation.app`?

rbarry82:
That was actually somewhat the conversation on the Grafana comment, and a brief suggestion to @sed-i in a chat comment. I agree that simplicity is better. "What if someone needs to add another condition" gets to the uncanny valley. Maybe two or _possibly_ three `if foo and bar and baz:` are ok. But `if foo and bar and baz and quuz:` looks long enough that someone will want to refactor/simplify it, and then we're possibly back to this bug again.

I like the simple method, but doing it via the "list comprehensions create a generator which lazily evaluates" may prevent a regression later (if I added a comment explaining why). There's potential tradeoffs either way, so it's somewhat just down to what we (democratically) agree on.

rbarry82:
Ok, I'll take no response as "simplicity, please" :)

PR #127, opened 2023-01-18
Pin ops to 2.0.0rc2 [test; do not merge]

sed-i:


benhoyt:
Test fixes for this in https://github.com/canonical/traefik-k8s-operator/pull/128. Actually most of the errors were from ops 1.5.3 (this charm is pinned to 1.5.2 currently), but a minor breakage on 2.0.0rc2 too.

PR #128, opened 2023-01-19
Fixes for newer versions of ops (1.5.4 and 2.0.0rc2)

benhoyt:
This PR fixes issues with the charm's tests on newer versions of `ops`. Currently ops is pinned to 1.5.2.

**Most of the failures are from changes in 1.5.3 (the latest 1.x version is 1.5.4).** Specifically:

* test_show_proxied_endpoints_action_no_relations previously had error
  "'ConfigData' object does not support item assignment". This was
  caused by canonical/operator#787 and the fix is to stop setting items
  on Model.config._data directly. Do it with update_config, but
  suppress the config-changed event it triggers.
* test_ingress_app_provider_relate_provide previosly had error
  ops.model.RelationDataTypeError: relation data values must be strings, not <class 'bool'>
  caused by canonical/operator#786 and the fix is to send string values
  instead of Python bools (lowercase as these would be sent to
  relation-set as YAML).
* Fix unit name in test_lib_per_unit* tests (it should be the remote
  app) to silence the warnings about that.

Also remove the unexpected config value "asyncio_mode" in
pyproject.toml's pytest settings to suppress the warning about that.

In addition, tweak some type annotations in the ingress charm libs to
fix these Pyright errors (I haven't done "charmcraft publish-lib" or
similar though):

```
pyright 1.1.290
/home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress.py
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress.py:218:16 - error: Expression of type "_RelationEventSnapshot" cannot be assigned to return type "dict[Unknown, Unknown]"
    "_RelationEventSnapshot" is incompatible with "dict[Unknown, Unknown]" (reportGeneralTypeIssues)
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress.py:221:25 - error: Argument of type "dict[Unknown, Unknown]" cannot be assigned to parameter "snapshot" of type "_RelationEventSnapshot" in function "restore"
    "dict[Unknown, Unknown]" is incompatible with "_RelationEventSnapshot" (reportGeneralTypeIssues)
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress.py:253:10 - error: Expression of type "IngressPerAppProviderEvents" cannot be assigned to declared type "property"
    "IngressPerAppProviderEvents" is incompatible with "property" (reportGeneralTypeIssues)
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress.py:409:10 - error: Expression of type "IngressPerAppRequirerEvents" cannot be assigned to declared type "property"
    "IngressPerAppRequirerEvents" is incompatible with "property" (reportGeneralTypeIssues)
/home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress_per_unit.py
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress_per_unit.py:320:10 - error: Expression of type "IngressPerUnitProviderEvents" cannot be assigned to declared type "property"
    "IngressPerUnitProviderEvents" is incompatible with "property" (reportGeneralTypeIssues)
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress_per_unit.py:578:16 - error: Expression of type "_RelationEventSnapshot" cannot be assigned to return type "dict[Unknown, Unknown]"
    "_RelationEventSnapshot" is incompatible with "dict[Unknown, Unknown]" (reportGeneralTypeIssues)
  /home/ben/w/traefik-k8s-operator/lib/charms/traefik_k8s/v1/ingress_per_unit.py:581:25 - error: Argument of type "dict[Unknown, Unknown]" cannot be assigned to parameter "snapshot" of type "_RelationEventSnapshot" in function "restore"
    "dict[Unknown, Unknown]" is incompatible with "_RelationEventSnapshot" (reportGeneralTypeIssues)
7 errors, 0 warnings, 0 informations
```

**But there are also a couple of failures from 2.0.0rc2:**

Just remove a couple of usages of the now-removed "key" parameter to
CharmBase.__init__() in tests.

This will still work fine with ops 1.x, but also works with 2.0.

**This change is broken into two commits with the fixes for 1.x and 2.0 separately.**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
+asyncio_mode = "auto"
 
 # Formatting tools configuration
 [tool.black]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,6 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
-asyncio_mode = "auto"
 
 # Formatting tools configuration
 [tool.black]
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -566,13 +566,14 @@ class TestTraefikIngressCharm(unittest.TestCase):
             pass
 
     @patch("charm._get_loadbalancer_status", lambda **unused: None)
+    @patch("charm.TraefikIngressCharm._on_config_changed", lambda self, event: None)
     @patch("charm.TraefikIngressCharm._traefik_service_running", lambda **unused: True)
     @patch("charm.KubernetesServicePatch", lambda **unused: None)
     def test_show_proxied_endpoints_action_no_relations(self):
         self.harness.begin_with_initial_hooks()
 
         action_event = Mock(spec=ActionEvent)
-        self.harness.charm.model.config._data["external_hostname"] = "foo"
+        self.harness.update_config({"external_hostname": "foo"})
         self.harness.charm._on_show_proxied_endpoints(action_event)
         action_event.set_results.assert_called_once_with({"proxied-endpoints": "{}"})
 
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -566,11 +566,11 @@ class TestTraefikIngressCharm(unittest.TestCase):
             pass
 
     @patch("charm._get_loadbalancer_status", lambda **unused: None)
-    @patch("charm.TraefikIngressCharm._on_config_changed", lambda self, event: None)
     @patch("charm.TraefikIngressCharm._traefik_service_running", lambda **unused: True)
     @patch("charm.KubernetesServicePatch", lambda **unused: None)
     def test_show_proxied_endpoints_action_no_relations(self):
         self.harness.begin_with_initial_hooks()
+        self.harness.container_pebble_ready("traefik")
 
         action_event = Mock(spec=ActionEvent)
         self.harness.update_config({"external_hostname": "foo"})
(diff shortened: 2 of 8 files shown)


tests/unit/test_charm.py
```diff
@@ -566,13 +566,14 @@ def test_relation_broken(self):
             pass
 
     @patch("charm._get_loadbalancer_status", lambda **unused: None)
+    @patch("charm.TraefikIngressCharm._on_config_changed", lambda self, event: None)
```
sed-i, line 569: The patch won't be necessary if you add `self.harness.container_pebble_ready("traefik")` after `begin_with_initial_hooks`.
I think that's all I had to do in #100 to make it work.

pyproject.toml
```diff
@@ -11,7 +11,6 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
-asyncio_mode = "auto"
```
sed-i, line 14: This is needed for itests to work

sed-i (approved):
Thanks @benhoyt!

tests/unit/test_charm.py
benhoyt: Ah true -- updated. And after you've updated to ops 2.0.0 (when final is out) you can remove all of those too, as `begin_with_initial_hooks()` now calls that.

pyproject.toml
benhoyt: Reinstated.

PR #136, opened 2023-02-03
Interface tests for `ingress`

PietroPasotti:
This PR adds integration with charm-relation-interfaces repo using the [interface tester pytest plugin](https://github.com/PietroPasotti/interface-tester-pytest).

This means that the charm is now able to test its implementation of the ingress interface against the scenario tests at https://github.com/canonical/charm-relation-interfaces/pull/57

This is achieved by targeting a specific branch, because that PR is not merged yet.
As soon as that pr is merged, we should update the refs in here to point to https://github.com/canonical/charm-relation-interfaces@main (which is the default).


tests/interfaces/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+from interface_tester import InterfaceTester
+from ops.pebble import Layer
+from scenario.state import Container, State
+
+from charm import TraefikIngressCharm
+
+
+@pytest.fixture
+def interface_tester(interface_tester: InterfaceTester):
+    with patch("charm.KubernetesServicePatch", lambda **unused: None):
```
sed-i, line 15: I wonder if it would be possible to pass a list of mocks that would apply throughout the tests, for example:
```python
interface_tester.run(
    mocks=[
        patch.object(MyCharm, "_version", property(lambda *_: "0.0.0")),
        patch.object(MyCharm, "_get_pvc_capacity"),
        ...
    ]
)
```

sed-i (commented):
This PR helped me a great deal to understand the interface tests solution!

tests/interfaces/conftest.py
PietroPasotti: what would the advantages of this design be in your opinion?

tests/interfaces/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+        interface_tester.configure(
+            # TODO: remove when the tester branch hits main
+            repo="https://github.com/PietroPasotti/charm-relation-interfaces",
+            branch="interface_tester/tester_plugin",
+            charm_type=TraefikIngressCharm,
+            state_template=State(
```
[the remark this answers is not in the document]
PietroPasotti: for each interface there can be any number of tests. One test=one scenario=one state. This state template is used to initialize the states for all tests; if none is provided, the tester will use the null state (`State()`).
This template is later merged with whatever initial state is provided by the tests themselves.

lucabello:
Looks good to me!
Small suggestion: not sure if it's too late to change this, but I think the `tests/interfaces` folder should probably be named with the singular `interface`, to be consistent with the others: `unit` as in `unit testing`, `interface` as in `interface testing`; it also would match the name of the tox environment like the other ones !

Just a minor nitpick though :)

lucabello (approved):

tox.ini
```diff
@@ -94,3 +94,14 @@ deps =
...
+[testenv:interface]
+description = Run interface tests
+deps =
+    pytest
+    ops-scenario
+    # todo: replace with pypi package when published
+    git+https://github.com/canonical/interface-tester-pytest
+    -r{toxinidir}/requirements.txt
+commands =
+    pytest -v --tb native {[vars]tst_path}/interface --log-cli-level=INFO -s {posargs}
```
Abuelodelanada, line 107: Does it make sense to use coverage as we do in unit tests?

Abuelodelanada (approved):

pyproject.toml
```diff
@@ -11,7 +11,6 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
-asyncio_mode = "auto"
```
sed-i, line 14: Iirc, removing this would fail the itests.

requirements-test.txt
sed-i, line 1: This file seems unused.
For reusability, we can use tox env and specify a tox env under deps.

tests/interface/conftest.py
sed-i, line 1: Should we add a note along the lines of:

> All interface tests are centrally hosted in https://github.com/PietroPasotti/charm-relation-interfaces.
> Add or modifying an interface tests in the repo there.

tox.ini
Abuelodelanada: It doesn't. I thought that where regular scenario tests, not interface

pyproject.toml
PietroPasotti: not any longer. locally they pass, and the previous CI runs only the integration tests were red.

tests/interfaces/test_ingress_per_unit.py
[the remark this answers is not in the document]
PietroPasotti: discussed privately, confusion is dispelled :boom:

tests/interfaces/conftest.py
PietroPasotti: @sed-i closing this as not relevant (it feels like a stylistic preference/implementation detail). If you want to discuss this further I'm open to that, but let's do that in charm-relation-interfaces or interface-tester-pytest.

PR #141, opened 2023-02-16
Peer Storage

PietroPasotti:
## Issue
This PR implements a PeerStorage class, a thin peer-relation-data wrapper exposing the same API as StoredState.

The idea is that you can:
- add a peer relation (by default the endpoint is `stored`, with interface `peer-storage`)
- replace: 
  ```python 
  class MyCharm(CharmBase):
      _stored = StoredState()
  ```

  with:
    ```python 
  class MyCharm(CharmBase):
      def __init__(...):
           self._stored = PeerStorage(self)
  ```
- and the rest should work out of the box.

The ultimate goal is to remove any (controller-backed) StoredState dependency from the Traefik charm.


PietroPasotti:
closed because in the foreseeable future this workaround will not be necessary. Keep the branch alive so that if in the future we need to remove use_controller_storage, then this is a possible way to do it.

PR #142, opened 2023-02-17
Moves private key from stored state to Juju secret

sanchezfdezjavier:
## Issue
Private key stored in stored state.

## Solution
Using Juju secrets for storing the private key.

## Context
Juju 3 has introduced secret management.

## Testing Instructions
1. Bootstrap your Juju `3.1` controller, for it you'll need microk8s with strict confinement.
2. Deploy `traefik`
```bash
juju deploy ./traefik-k8s_ubuntu-20.04-amd64.charm traefik-ingress --trust --resource traefik-image=docker.io/jnsgruk/traefik:2.6.1
```
3. Deploy `tls-operator` for providing dummy self signed certs
```bash
juju deploy tls-certificates-operator \
  --config generate-self-signed-certificates=true \
  --config ca-common-name=dummycommonname
```
4. Relate both apps
```bash
juju relate traefik-k8s tls-certificates-operator
```

5. Make sure the private key secret is stored using Juju secrets
```bash
❯ juju secrets 
ID                    Owner            Rotation  Revision  Last updated
cfnkrncviroh030nrjbg  traefik-ingress  never            1  52 minutes ago  
```

```bash
❯ juju show-secret cfnkrncviroh030nrjbg                    
cfnkrncviroh030nrjbg:
  revision: 1
  owner: traefik-ingress
  label: PRIVATE_KEY
  created: 2023-02-17T09:50:22Z
  updated: 2023-02-17T09:50:22Z
```

6. Additionally we can check as well going into the container
```bash
❯ microk8s.kubectl exec -n traefik-secrets-3 -it traefik-ingress-0 -c traefik -- /bin/bash
root@traefik-ingress-0:/# ls
bin  boot  charm  dev  etc  home  lib  lib32  lib64  libx32  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
root@traefik-ingress-0:/# cd /opt/traefik/juju
root@traefik-ingress-0:/opt/traefik/juju# ls
certificate.cert  certificate.key  certificates.yaml
```

## Release Notes
Traefik's private key was previously stored in `self._stored`, with the recent introduction of Juju secrets it has been moved there and removed from stored state.


jnsgruk:
This is interesting work, thanks @sanchezfdezjavier. We need to consider backward compatibility, though.

There are two options:

1. Burn the world, include `assumes: [ "juju > 3.0" ]` in the `metadata.yaml`
2. Detect the version of Juju, use secrets if they're available, and revert back to the existing impl if they dont.

I'll let the other folks here decide the best way forward :) @simskij and @gruyaume likely have opinions here :)

sanchezfdezjavier:
> This is interesting work, thanks @sanchezfdezjavier. We need to consider backward compatibility, though.
> 
> There are two options:
> 
> 1. Burn the world, include `assumes: [ "juju > 3.0" ]` in the `metadata.yaml`
> 2. Detect the version of Juju, use secrets if they're available, and revert back to the existing impl if they dont.
> 
> I'll let the other folks here decide the best way forward :) @simskij and @gruyaume likely have opinions here :) (edit: and @benhoyt)

Good catch, thanks Jon. Is there any way to know how many people is using this charm? I guess it would be key for the choice.

PietroPasotti:
> Good catch, thanks Jon. Is there any way to know how many people is using this charm? I guess it would be key for the choice.

Afraid there's no way to know that. IMHO this charm has burned the world enough for the time being, I'm not against using secrets to store this key, but if the goal is to get rid of storedstate, and since this secret is not being shared anyway, an equally good solution would be to write it to disk, I think?

PietroPasotti:
> 

on second thought, I think this is a perfect use case for secrets. But I still think we should maintain backwards compatibility

jameinel:
So you can do a runtime check to see if "secret-create" exists in $PATH, and update your logic according to that. I don't know if that is preferred to a user config parameter that forces it to fail if it isn't available.

Offhand, I'm pretty sure the secret key has to end up on disk anyway, so that the application that is running can use it to handle TLS connections. (It may depend on the application, and maybe on disk it is passphrase protected and unblocked by the charm, but then how does the charm know what the key is to unlock the certificate...)

It is probably reasonable to store some amount of information 'remotely', either in a relation config, or on the juju controller, etc, so that someone who gets access to the disk of the container cannot immediately have access to the content (you'd have the information you need to interact with a juju controller, or the backend secret store, etc, but that access could be disabled if you lost the disk of a container).

So... it seems worthwhile to use secret-set/get or state-set/get to hold onto the passphrase of the key file, but then allow the encrypted content to sit on disk (and then use pebble-ready to unlock the content of the key for the app at runtime), *if* the app supports it.

simskij:
I'm afraid option 2 will be the way to go on this one. This charm has burnt the world way too many times already. 😅

gruyaume:
Ultimately adding juju secrets to any existing charms will make it so that people have to use juju > 3.0. Do we want to place such logic about juju feature availability in charms? It seems like a slope most should avoid. I think it's fine to say to a given version of a charm only works on juju > 3.0. Example traefik-k8s 1.0/stable works with juju 2.x and traefik-k8s 2.0/stable works only with juju 3.x.

rbarry82:
The probable damage here (if any of the certs get lost, really, since the serial would almost certainly change anyway) is that a pod reschedule/recycle would lead to users seeing a warning about the cert being change.

These are probably viable for unit lifecycle and good targets for putting in state if secrets aren't available, rather than storing on disk.

benhoyt:
I agree that for this charm we should almost certainly still support Juju 2.9, so we have to do some kind of in-charm check. I don't think it's too onerous for to have an if/else (the cost of backwards compatibility).

There's an `ops` spelling for "does the Juju version we're running on support secrets?" and it is as follows:

```python
from ops.jujuversion import JujuVersion
...
if JujuVersion.from_environ().has_secrets:
    # use secrets
else:
    # use stored state (current code)
```

sanchezfdezjavier:
> Ultimately adding juju secrets to any existing charms will make it so that people have to use juju > 3.0. I don't think we want to add logic about juju feature availability in most charms, it seems like a slope most should avoid. This being said, it seems like a good usecase for tracks. Example traefik-k8s 1.0/stable works with juju 2.x and traefik-k8s 2.0/stable works only with juju 3.x.

I see several problems from this approach:
1. It would require us to maintain both versions of the charm.
2. I didn't think much of it, but I don't know how would having two charms would do with upgrades.
3. Do we want the user to be taking care of Juju compatibility? Our goal is to make the operator's life easier with Juju, not introduce more complexity. Besides, If the cost for devs is as @benhoyt mentions, is "not that bad" and the most simple approach for the end user, that is what I think we should prioritise.

However, this is a good point, correct me if I'm wrong.

benhoyt:
I guess the one thing that might be a trickier than an if-else with having to support 2.9 and 3.0 is migration. When the charm is upgraded to Juju 3.0 we need to have an upgrade hook or something that migrates the keys from stored state to proper secrets. Shouldn't be rocket science, but won't be the quickest thing to test end-to-end.

src/charm.py
```diff
@@ -260,7 +263,8 @@ def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent):
     def _on_all_certificates_invalidated(self, event: RelationBrokenEvent) -> None:
         if self.container.can_connect():
             self._stored.certificate = None
-            self._stored.private_key = None
+            secret = self.model.get_secret(label=PRIVATE_KEY_SECRET_LABEL)
+            secret.remove_all_revisions()
```
sed-i, line 267: Currently the charm would go into error state if this raises, which I think is good, because Juju would keep auto retrying.
But would this raise if there are no secrets/revisions left?

src/charm.py
```diff
@@ -289,7 +296,9 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
             return
 
         new_csr = generate_csr(
-            private_key=private_key.encode(),
+            private_key=self.model.get_secret(label=PRIVATE_KEY_SECRET_LABEL)
+            .get_content()["private-key"]
```
sed-i, line 300: Could/should this code block become a property in the tls-certs library?

For backwards-compatibility you may want a "secret store" wrapper class that chooses a backend (stored or secrets).

src/charm.py
sanchezfdezjavier: Good point, fixed.

src/charm.py
sanchezfdezjavier: I've created the secret store, let me know if that works.

simskij:
> Ultimately adding juju secrets to any existing charms will make it so that people have to use juju > 3.0. I don't think we want to add logic about juju feature availability in most charms, it seems like a slope most should avoid. This being said, it seems like a good usecase for tracks. Example traefik-k8s 1.0/stable works with juju 2.x and traefik-k8s 2.0/stable works only with juju 3.x.

In principle I agree with you, but in this case I think **not** supporting both isn't really an option.

src/charm.py
```diff
@@ -109,6 +117,8 @@ def __init__(self, *args):
             chain=None,
         )
 
+        self.secret_store.migrate_private_key()
```
sed-i, line 120: **Question**: Is it safe to read/write `secret_store` backends in the constructor?

Not sure if:
- On install, can we already read/write secrets?
- Before pebble-ready, can we read/write stored state?
- (Nitpick) can we touch stored state/secrets on stop/remove?

src/charm.py
```diff
@@ -212,7 +222,7 @@ def _on_install(self, event) -> None:
         # Generate key without a passphrase as traefik does not support it
         # https://github.com/traefik/traefik/pull/6518
         private_key = generate_private_key()
-        self._stored.private_key = private_key.decode()
+        self.secret_store.store_private_key(private_key.decode())
```
sed-i, line 225: **Praise**: Delegating to a separate class (which could in the future become a WorkloadManager) is a great idea.

src/charm.py
```diff
@@ -991,6 +1004,99 @@ def is_hostname(st: Optional[str]) -> bool:
...
+
+    def store_private_key(self, private_key: str) -> None:
+        """Stores the private key in the appropriate backend.
+
+        Args:
+            private_key(str): Private key to store.
+        """
+        if self.using_juju_secrets:
+            content = {"private-key": private_key}
+            self.charm.app.add_secret(content, label=self._PRIVATE_KEY_SECRET_LABEL)
```
sed-i, line 1027: **Question**: Is it safe to call `add_secret` multiple times with the same content/label?

sed-i (commented):
Thanks @sanchezfdezjavier and sorry for the delay in reviewing!
I think this looks good and the functionality is certainly desirable.

I am happy to help rebasing!

simskij:
This PR has been stale for a long time. Closing without action.

PR #143, opened 2023-02-23
Fix has_ingress integration tests

ghislainbourgeois:
Integration tests for `ingress-per-app` and `ingress-per-unit` were validating that the ingress is present by calling `ping` with the `-p` flag and the port. The `-p` flag is not for specifying a port however, but to specify padding data to put in the payload of the `ICMP` packet.

Because the `metallb` configuration is using the local IP of the host running `microk8s`, it is expected to always answer to pings. Therefore, those tests were not testing anything.

This change replaces the `ping` with a direct socket, testing the TCP connectivity to the ingress port.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -5,6 +5,7 @@ import grp
 import logging
 import os
 import shutil
+import socket
 import subprocess
 from collections import defaultdict
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


tests/integration/conftest.py
```diff
@@ -317,9 +318,19 @@ def get_relation_data(
     return RelationData(provider=provider_data, requirer=requirer_data)
 
 
-def assert_can_ping(ip, port):
-    response = os.system(f"ping -c 1 {ip} -p {port}")
-    assert response == 0, f"{ip}:{port} is down/unreachable"
+def _can_connect(ip, port) -> bool:
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.connect(ip, int(port))
```
sed-i, line 324:
```suggestion
        s.connect((ip, int(port)))
```

tests/integration/conftest.py
```diff
@@ -317,9 +318,19 @@ def get_relation_data(
...
+        return True
+    except OSError:
```
sed-i, line 326: There are [more kinds](https://docs.python.org/3/library/socket.html#socket.socket.connect). For itests perhaps we can have a bare `except`?

sed-i (commented):
Thanks @ghislainbourgeois, good catch!

tests/integration/conftest.py
ghislainbourgeois: Good point, changed to a bare `except`.

tests/integration/conftest.py
ghislainbourgeois: Good catch, fixed

sed-i (approved):
LGTM.
@PietroPasotti?

PietroPasotti (approved):
Good catch! Thanks for the fix :)

PR #146, opened 2023-03-10
Switch Traefik to use the ROCK on ghcr.io

simskij:
## Issue
We are currently running Traefik using the upstream OCI image.

## Solution
We are now building our own ROCKs for the observability stack. We should also make sure we're using them so that effort is not in vain.

## Testing Instructions
- Deploy the charm with the new resource
- Verify functionality is still there

## Release Notes
Switch from upstream OCI image to ROCK

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -25,9 +25,9 @@ containers:
 resources:
   traefik-image:
     type: oci-image
-    description: OCI image for traefik (https://hub.docker.com/_/traefik/)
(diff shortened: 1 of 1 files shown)


Abuelodelanada (approved):

Abuelodelanada:
@simskij May you re-run itest again before merging?

dstathis (approved):

sed-i:
This has been accomplished by #204. Closing.

PR #150, opened 2023-03-16
Reconcile CI

rbarry82:
## Issue
Fetch-lib and start a PR the bot won't clobber to stablize CI


PietroPasotti (commented):
the itests are failing. Can't build. Wondering if that has to do with setuptools/pkg-config?

sed-i (approved):

rbarry82:
I mean, it passes, but do we really need a 25 minute timeout @sed-i ? Green is green, but it seems like this needs a follow-up issue... somewhere.

PR #165, opened 2023-04-20
fix itests

PietroPasotti:
itests are red on main; try to fix them

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -318,7 +318,7 @@ async def deploy_traefik_if_not_deployed(ops_test: OpsTest, traefik_charm):
 
     # now we're most definitely active.
     async with ops_test.fast_forward():
-        await ops_test.model.wait_for_idle(["traefik-k8s"], status="active", timeout=1000)
+        await ops_test.model.wait_for_idle(["traefik-k8s"], timeout=1000)
 
 
 async def deploy_charm_if_not_deployed(ops_test: OpsTest, charm, app_name: str, resources=None):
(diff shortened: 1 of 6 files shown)


PietroPasotti:
it looks like when deploy_traefik_if_not_deployed times out, (1000 seconds later), traefik is still at scale 0:
image

tests/integration/conftest.py
```diff
@@ -318,7 +318,7 @@ async def deploy_traefik_if_not_deployed(ops_test: OpsTest, traefik_charm):
 
     # now we're most definitely active.
     async with ops_test.fast_forward():
-        await ops_test.model.wait_for_idle(["traefik-k8s"], status="active", timeout=1000)
+        await ops_test.model.wait_for_idle(["traefik-k8s"], timeout=1000)
```
sed-i, line 321: I'm surprised this ever passed: fast-forward would emit update-status every 5 sec so the charm will never be idle for the [default 15sec](https://github.com/juju/python-libjuju/blob/d809caf11dc2e16fd2dc2821383a8a452c634a9e/juju/model.py#L2472).

PietroPasotti: that is also a good point

PietroPasotti: but that line of code is in a lot of places, why did it only fail there?

PietroPasotti: maybe that's why they were flaky: if it happens to check the status while it's performing an update, it will get maintenance, otherwise it will get active

PietroPasotti: I'm getting rid of all fast forwards to see what happens

PietroPasotti: maybe it would also work to remove the 'active' constraint only

PR #168, opened 2023-04-27
use ruff instead of flake8

lucabello:
The **flake8** plugins we use are re-implemented in **ruff**, which is why they have been removed from the dependencies.

**isort** is enabled (and used when formatting) through the `I001` rule.

The "exclude list" has been updated to omit folders that **ruff** excludes by default.

Everything else is the same, the only difference being two new rules being ignored (which weren't checked before anyway): 
* `N818: Exception name should be named with an Error suffix`, which I excluded due to the big refactoring required for it;
* `RET504: Unnecessary variable assignment before return statement`, which we do everywhere as it increases readability.

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress.py
+++ b/lib/charms/traefik_k8s/v1/ingress.py
@@ -69,7 +69,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 DEFAULT_RELATION_NAME = "ingress"
 RELATION_INTERFACE = "ingress"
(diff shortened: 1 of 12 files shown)


lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -229,11 +229,11 @@ class IngressPerAppDataProvidedEvent(_IPAEvent):
...
     if typing.TYPE_CHECKING:
-        name = None  # type: Optional[str]
-        model = None  # type: Optional[str]
-        port = None  # type: Optional[str]
-        host = None  # type: Optional[str]
-        strip_prefix = False  # type: bool
+        name: Optional[str] = None
+        model: Optional[str] = None
+        port: Optional[str] = None
+        host: Optional[str] = None
```
dstathis, line 235: I thought the preferred way to do this was `name: str = None`. If that doesn't work, feel free to ignore this comment.

dstathis (approved):

lib/charms/traefik_k8s/v1/ingress.py
lucabello: It does work for **ruff**'s linting, but **pyright** complains about it!

PR #178, opened 2023-05-11
Add a constructor arg and support for redirecting to HTTPS

### LINKED ISSUE #176: HTTPS endpoint in databag
### Bug Description

Even after tls is enabled the http endpoint is provided in the databag. This is not the behavior I would expect.

Applications related to traefik need to be aware of when HTTPS is enabled so that they can advertise that endpoint.

I would be happy to open a PR with a fix for this, if we reach an agreement on the behavior of the charm.

### To Reproduce

I did the following steps to deploy my charms, but simpler deployments can probably be found:
```
juju deploy traefik-k8s traefik-public 
juju deploy kratos --channel edge --trust
juju deploy tls-certificates-operator
juju deploy postgresql-k8s --channel edge --trust

juju config tls-certificates-operator ca-common-name=demo.ca.local generate-self-signed-certificates=true

juju integrate kratos postgresql-k8s
juju integrate kratos:public-ingress traefik-public
juju integrate traefik-public tls-certificates-operator
```

The contents of the databag can be seens by running:
```
juju show-unit kratos/0
```

### Environment

This was tested using juju=3.1.2 and microk8s=v1.27.1 with the metallb plugin to provide an ip address to traefik.

### Relevant log output

```shell
- relation-id: 8
    endpoint: admin-ingress
    related-endpoint: ingress
    application-data:
      ingress: |
        url: http://10.64.140.43:80/test-kratos
    related-units:
      traefik-admin/0:
        in-scope: true
        data:
          egress-subnets: 10.64.140.43/32
          ingress-address: 10.64.140.43
          private-address: 10.64.140.43
```

### Additional context

_No response_

### LINKED ISSUE #177: HTTP to HTTPS redirect
### Enhancement Proposal

After tls is enabled in traefik, a rule could be created for HTTP to HTTPS redirects. I see no reason not to do that.

An example of how to implement that can be found [here](https://jensknipper.de/blog/traefik-http-to-https-redirect/).

rbarry82:
## Issue
Even though TLS support has been added, there is no meaningful way for users who always want to use it to force redirection.

## Solution
Add a `redirect_https` argument (defaulting to false) to the IPA/IPU constructors, add a key to relation data, and ensure that it's set even for subdomain routing.

In the absence of a TLS relation, Traefik's own self-signed cert will be used, but HTTPS is HTTPS, so that's up to administrators.

Closes #177

This will not, in fact, update the URL in the databag, but will force clients to HTTPS, flagging the redirect as permanent (so browsers and well-behaved clients will keep track)

## QA
Doc + manual testing described in a [new how-to](https://discourse.charmhub.io/t/traefik-k8s-docs-force-https-redirect/10810).

## Release Notes
Add a constructor arg and support for redirecting to HTTPS


lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -418,6 +429,7 @@ def __init__(
         host: Optional[str] = None,
         port: Optional[int] = None,
         strip_prefix: bool = False,
+        redirect_https: Optional[str] = None,
```
sed-i, line 432: Shouldn't this be 
```suggestion
        redirect_https: bool = False,
```

src/charm.py
```diff
@@ -709,14 +709,17 @@ def _get_prefix(data: Union["RequirerData_IPU", "RequirerData_IPA"]):
...
-        if self._routing_mode is _RoutingMode.path and data.get("strip-prefix", False):
-            return {
-                f"juju-sidecar-noprefix-{prefix}": {
-                    "stripPrefix": {"prefixes": [f"/{prefix}"], "forceSlash": False}
-                }
-            }
+        """Generate a middleware config."""
+        config = {}  # type: Dict[str, Dict[str, Any]]
+        if self._routing_mode is _RoutingMode.path:
+            if data.get("strip-prefix", False):
```
sed-i, line 715: Should this (and below) be
```suggestion
            if data.get("strip-prefix") == "true":
```

PietroPasotti: the library does the conversion
image

tests/unit/test_charm.py
```diff
@@ -48,6 +49,9 @@ def _requirer_provide_ingress_requirements(
     if strip_prefix:
         data["strip-prefix"] = "true"
 
+    if redirect_https:
+        data["redirect-https"] = "true"
```
PietroPasotti, line 53: [thought] I personally don't like optionally present keys. I'd rather it be there and be 'false' than it maybe being there and it not being there might mean that the interface is obsolete OR that redirect-https is False. 
Still, for backwards compatibility, the provider shouldn't complain if it is missing: that's ok.

tests/unit/test_charm.py
```diff
@@ -127,14 +131,15 @@ def test_pebble_ready_with_gateway_address_from_config_and_path_routing_mode(sel
...
+        for inject_middleware in (False, True):
             with self.subTest():
                 _requirer_provide_ingress_requirements(
                     harness=self.harness,
                     relation=relation,
                     host="10.1.10.1",
                     port=9000,
-                    strip_prefix=strip_prefix,
+                    strip_prefix=inject_middleware,
+                    redirect_https=inject_middleware,
```
PietroPasotti, line 142: shouldn't we test for all combinations?

tests/unit/test_charm.py
```diff
@@ -369,15 +401,16 @@ def test_pebble_ready_with_gateway_address_from_config_and_subdomain_routing_mod
...
+        for inject_middleware in (False, True):
             # in subdomain routing mode this should not have any effect
             with self.subTest():
                 _requirer_provide_ingress_requirements(
                     harness=self.harness,
                     relation=relation,
                     host="10.1.10.1",
                     port=9000,
-                    strip_prefix=strip_prefix,
+                    strip_prefix=inject_middleware,
```
PietroPasotti, line 412: same

tests/unit/test_charm.py
```diff
@@ -444,15 +498,16 @@ def test_pebble_ready_no_leader_with_gateway_address_from_config_and_subdomain_r
...
```
PietroPasotti, line 509: same

PietroPasotti (approved):

tests/unit/test_charm.py
rbarry82: In principle, yes.

For the time being, I wonder. Even adding an additional is pretty extreme, because it suddenly becomes a combinatorics problem with `(n*2)!` (where `n` is the number of possible middlewares) testcases, and mapping it up to anything, even `itertools.product` would probably require `functools.partial`, leaving only the middleware args, and a similar helper to actually "build" the middlewares.

It's a lot of moving pieces which are certainly necessary if/when we add more, but I wonder where it's strictly needed now, since the crucial case of differences (no `stripPrefix` on path-based routing, but `redirectSchema` stays) is actually covered in this patch.

tests/scenario/test_middlewares.py
```diff
@@ -0,0 +1,62 @@
...
+            config={"routing_mode": "path", "external_hostname": "testhostname"},
+            containers=self.containers,
+            relations=[relation],
+        )
+
+        # WHEN a `relation-changed` hook fires
+        out = Context(charm_type=TraefikIngressCharm).run(relation.changed_event, state)
+
+        # THEN the rendered config file contains middlewares
+        with out.get_container("traefik").filesystem.open("/opt/traefik/juju/juju_ingress_ingress_now_what.yaml", "rt") as f:
```
sed-i, line 55: Would you be able to help with this scenario test @PietroPasotti?
Currently this errors out with `FileNotFoundError: /opt`.

What I would like to do is:
1. Find the path to the config file rendered by the charm (e.g. `/opt/traefik/jujujuju_ingress_ingress_7_alertmanager.yaml`).
2. Read it out so I could compare to the expected.

PietroPasotti: this is how you setup a mocked container so that when the charm reads it it will find some file:
```python
text = "lorem ipsum/n alles amat gloriae foo"
    file = tempfile.NamedTemporaryFile()
    pth = Path(file.name)
    pth.write_text(text)

    def callback(self: CharmBase):  # charm code
        container = self.unit.get_container("foo")
        baz = container.pull("/bar/baz.txt")
        assert baz.read() == text

    ctx.run(
        state=State(
            containers=[
                Container(
                    name="foo",
                    can_connect=True,
                    mounts={"bar": Mount("/bar/baz.txt", pth)},
                )
            ]
        ),
        event="start",
        post_event=callback,
    )
```

and this is how you assert that the charm has written some file to a mocked filesystem:
```python
    td = tempfile.TemporaryDirectory()
    state = State(
        containers=[
            Container(
                name="foo", can_connect=True, mounts={"foo": Mount("/foo", td.name)}
            )
        ]
    )

    out = ctx.run(
        event="start",
        state=state,
        post_event=callback,
    )
    file = out.get_container("foo").filesystem.open("/foo/bar/baz.txt")
    assert file.read() == text
```

tests/scenario/test_middlewares.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import unittest
```
PietroPasotti, line 4:
```suggestion
import pytest
```

tests/scenario/test_middlewares.py
```diff
@@ -0,0 +1,62 @@
...
+    @patch("charm.TraefikIngressCharm.external_host", PropertyMock(return_value="testhostname"))
+    @patch("charm.TraefikIngressCharm._traefik_service_running", PropertyMock(return_value=True))
+    @patch("charm.TraefikIngressCharm._tcp_entrypoints_changed", MagicMock(return_value=False))
+    def test_start_traefik_is_not_running(self, *_):
+        # GIVEN an ingress (per app) relation is requesting middlewares (strip prefix and
+        #  redirect-https)
+        relation = Relation(
+            endpoint="ingress",
+            interface="ingress",
+            remote_app_name="remote",
```
PietroPasotti, line 32: this is the default

tests/scenario/test_middlewares.py
```diff
@@ -0,0 +1,62 @@
...
+            relation_id=0,
```
PietroPasotti, line 33: this ID you might need to pass so that the filename is predictable: `.../juju_ingress_ingress_<THIS_ID>_alertmanager`

tests/scenario/test_middlewares.py
```diff
@@ -0,0 +1,62 @@
...
+class TestMiddlewares(unittest.TestCase):
+    def setUp(self) -> None:
+        self.containers = [Container(name="traefik", can_connect=True)]
+
+        version_patcher = patch.object(TraefikIngressCharm, "version", property(lambda *_: "0.0.0"))
+        self.version_patch = version_patcher.start()
+        self.addCleanup(version_patcher.stop)
+
+    @patch("charm.TraefikIngressCharm.external_host", PropertyMock(return_value="testhostname"))
+    @patch("charm.TraefikIngressCharm._traefik_service_running", PropertyMock(return_value=True))
```
PietroPasotti, line 24: you might get rid of this by declaring the service in the Container object, but this is quicker.

tests/scenario/test_setup.py
```diff
@@ -21,46 +20,39 @@ def test_start_traefik_is_not_running(*_):
...
-            config={"routing_mode": "path"},
-            containers=[Container(name="traefik", can_connect=False)],
-        ),
-        "start",
-        TraefikIngressCharm,
+    state = State(
+        config={"routing_mode": "path"},
+        containers=[Container(name="traefik", can_connect=False)],
     )
+    out = Context(charm_type=TraefikIngressCharm).run("start", state)
```
PietroPasotti, line 55: tip: avoid repeating `Context(charm_type=TraefikIngressCharm)` by putting it in a fixture in `/tests/scenario/conftest.py`

tests/scenario/test_middlewares.py
sed-i: Could you please elaborate/example link?

tests/scenario/test_middlewares.py
sed-i: The "remote" literal is all over the place so I'd rather have it explicit.

tests/scenario/test_middlewares.py
sed-i: Good point. Parameterized!

tests/scenario/test_middlewares.py
sed-i: Thanks! This worked well.

TROUBLESHOOTING.md
sed-i, line 1: Moved to https://discourse.charmhub.io/t/traefik-k8s-docs-tls-termination-using-a-local-ca/10779

sed-i:
The test_charm_tcp test in traefik was flaky because:

- The static config (the one in /etc/traefik) wasn't updated on relation changed, only on other hooks, such as update-status.
- The default update-status interval of 5m was interfering with the test.

"Disabling" update-status exposed the problem and a patch in traefik code fixed the issue.

tests/scenario/test_middlewares.py
PietroPasotti: for example here: https://github.com/canonical/traefik-k8s-operator/pull/180/files#diff-965c0ce47f14d31c34286d6286f9c7a08375cce8ad74ade0a2c274fb6a9affd9R55 I set up the traefik container so that _traefik_service_running returns True. (without needing to mock it).

See this test https://github.com/canonical/traefik-k8s-operator/pull/180/files#diff-965c0ce47f14d31c34286d6286f9c7a08375cce8ad74ade0a2c274fb6a9affd9R62 for usage

PR #180, opened 2023-05-22
Ingress per leader/ingress per app

### LINKED ISSUE #159: Ingress-per-app is in reality ingress-by-leader
### Bug Description

Currently, ingress-per-app does not really do ingressing per application using the service loadbalancer. Rather, it always routes to the leader. See https://github.com/canonical/traefik-k8s-operator/blob/34a0aa5ca57bf2d2200a37d3f884b8a13b0d4920/lib/charms/traefik_k8s/v1/ingress.py#L509-L510

### To Reproduce

-

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

See https://github.com/jnsgruk/hello-kubecon/pull/16 for example.

PietroPasotti:
This pr:
- copies v1/ingress into v0/ingress-per-leader without changes (except renaming the default endpoint, interface name and a few types)
- fixes v1/ingress so that it actually provides ingress per app instead of providing ingress per leader

Fixes #159 

## Solution
In order to provide ingress per app, we need host/port for each remote unit, so I refactored the requirer to provide over app databag the model/name and toplevel configs, but over unit databag the hostname/port pairs (I reckon the port could also be lifted to app data though?)

Open questions:
- should we lift unit databag's 'port' to app databag?
- should I just use getfqdn for the host (on the requirer side)?
- this is a lot more than a vbump to ingress v1: this changes the schema and the interface; it is effectively ingress v2. I think this warrants a new pull request to charm-relation-interfaces.

### DIFF
--- a/lib/charms/traefik_k8s/v0/ingress_per_leader.py
+++ b/lib/charms/traefik_k8s/v0/ingress_per_leader.py
@@ -8,7 +8,6 @@ and provides a Python API for both requesting and providing per-application
 ingress, without load balancing across Traefik units: all traffic is routed
 through the leader instead.
 This is not a typical ingress mode and is intended only for special cases.
-Should we further warn the reader?
 This is not a typical ingress mode and is intended only for special cases.
 Unless you're sure this is something for you, you should probably be looking at `ingress` or
 `ingress-per-unit` instead.


src/charm.py
```diff
@@ -626,19 +631,24 @@ def _provide_routed_ingress(self, relation: Relation):
         self._push_configurations(relation, config)
 
     def _provide_ingress(
-        self, relation: Relation, provider: Union[IngressPerAppProvider, IngressPerAppProvider]
+            self,
+            relation: Relation,
+            provider: Union[IngressPerAppProvider, IngressPerAppProvider, IngressPerLeaderProvider],
```
Abuelodelanada, line 636: Is this correct? I mean, the duplicated  `IngressPerAppProvider`

PietroPasotti: fixing it

src/charm.py
```diff
@@ -193,13 +201,10 @@ def __init__(self, *args):
...
 
-        ipa_events = self.ingress_per_app.on
-        observe(ipa_events.data_provided, self._handle_ingress_data_provided)
-        observe(ipa_events.data_removed, self._handle_ingress_data_removed)
-
-        ipu_events = self.ingress_per_unit.on
-        observe(ipu_events.data_provided, self._handle_ingress_data_provided)
-        observe(ipu_events.data_removed, self._handle_ingress_data_removed)
+        # observe data_provided and data_removed events for all types of ingress we offer:
+        for ingress_per in (self.ingress_per_leader, self.ingress_per_unit, self.ingress_per_app):
```
sed-i, line 205: The `_per` may be a bit confusing. How about simply:

```diff
-ingress_per
+ingress
```

tests/scenario/conftest.py
```diff
@@ -0,0 +1,11 @@
+from unittest.mock import patch
+
+import pytest
+from charm import TraefikIngressCharm
+
+
+@pytest.fixture
+def traefik_charm():
+    with patch("charm.KubernetesServicePatch"):
+        with patch("lightkube.core.client.GenericSyncClient"):
```
sed-i, line 10: I think it's possible to
```suggestion
    with patch("charm.KubernetesServicePatch"), patch("lightkube.core.client.GenericSyncClient"):
```

tests/scenario/conftest.py
```diff
@@ -0,0 +1,11 @@
...
+            yield TraefikIngressCharm
```
sed-i, line 11: TIL `yield` fixtures are recommended over `return` fixture because they let you add cleanup code.
So should we:
```python
svc_patcher = patch("charm.KubernetesServicePatch")
svc_patch = svc_patcher.start()

client_patcher = patch("lightkube.core.client.GenericSyncClient")
client_patch = client_patcher.start()

yield TraefikIngressCharm

svc_patcher.stop()
client_patcher.stop()
```

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+
+@pytest.mark.parametrize("port, host", ((80, "1.1.1.1"), (81, "10.1.10.1")))
+@pytest.mark.parametrize("event_name", ("joined", "changed"))
+def test_ingress_per_app_created(
+    context, port, host, model, traefik_container, event_name, temp_opt
+):
+    """Check the config when a new ingress per leader is created or changes (single remote unit)."""
+    ipl = Relation(
```
sed-i, line 66: Typo?
```diff
-ipl
+ipa
```

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,172 @@
...
+                    "loadBalancer": {"servers": [{"url": f"http://{host}:{port}"}]}
+                }
+            },
+        }
+    }
+    cfg_file.write_text(yaml.safe_dump(initial_cfg))
+
+    def _get_mock_data(n: int):
+        return {
+            "port": str(port + n),
```
sed-i, line 132: Not sure why the port number changes per unit.

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,172 @@
...
+@pytest.mark.parametrize("port, host", ((80, "1.1.1.{}"), (81, "10.1.10.{}")))
+@pytest.mark.parametrize("n_units", (2, 3, 10))
+def test_ingress_per_app_scale(context, host, port, model, traefik_container, temp_opt, n_units):
+    """Check the config when a new ingress per leader unit joins."""
+    cfg_file = temp_opt.joinpath("traefik", "juju", "juju_ingress_ingress_1_remote.yaml")
+    cfg_file.parent.mkdir(parents=True)
+
+    # config that would have been generated from mock_data_0
+    # same as config output of the previous test
+    initial_cfg = {
```
sed-i, line 106: This test concerns `...["servers"]`.

Instead of hard-coding the entire relation data, would it make more sense to have a fixture that brings the charm through an equivalent of `begin_with_inital_hooks`?

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,172 @@
...
+        relation_id=1,
+    )
+    state = State(
+        model=model,
+        config={"routing_mode": "path", "external_hostname": "foo.com"},
+        containers=[traefik_container],
+        relations=[ipl],
+    )
+
+    event = getattr(ipl, f"{event_name}_event")
```
sed-i, line 83: It would be very helpful to have some comments inline, for example:
```suggestion
    # WHEN any relevant event fires
    event = getattr(ipl, f"{event_name}_event")
```

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    assert generated_config["http"]["services"]["juju-test-model-remote-0-service"] == {
+        "loadBalancer": {"servers": [{"url": f"http://{host}:{port}"}]}
+    }
+
+
+@pytest.mark.parametrize("port, host", ((80, "1.1.1.{}"), (81, "10.1.10.{}")))
+@pytest.mark.parametrize("n_units", (2, 3, 10))
+def test_ingress_per_app_scale(context, host, port, model, traefik_container, temp_opt, n_units):
```
sed-i, line 99: The setup for this test looks quite different from the previous one. It also seems to overlap. Do we need the previous test?

tests/scenario/test_status.py
```diff
@@ -1,70 +1,76 @@
...
-
-        # THEN unit status is `blocked`
-        self.assertEqual(
-            out.status.unit, ("blocked", "invalid routing mode: invalid_routing; see logs.")
-        )
-
-
-if __name__ == "__main__":
-    unittest.main()
+@pytest.fixture
```
sed-i, line 13: :skull: :skull: :skull: pytest :skull: :skull:  :skull:

tests/scenario/test_setup.py
```diff
@@ -5,15 +5,13 @@
...
+from charm import _TRAEFIK_SERVICE_NAME
 from scenario import Container, State
 from scenario.runtime import trigger
 
 
-@patch("charm.KubernetesServicePatch")
-@patch("lightkube.core.client.GenericSyncClient")
 @patch("charm.TraefikIngressCharm.external_host", PropertyMock(return_value="foo.bar"))
-def test_start_traefik_is_not_running(*_):
+def test_start_traefik_is_not_running(*_, traefik_charm):
```
sed-i, line 14: Where is the `traefik_charm` variable coming from?

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -69,7 +69,7 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 13
+LIBPATCH = 14
```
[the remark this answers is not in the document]
PietroPasotti: yes I think so. updated.

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -115,19 +121,21 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
...
+# Model of the data each requirer unit will need to provide.
+RequirerFollowerData = TypedDict("RequirerFollowerData", {"host": str, "port": int})
 # Provider ingress data model.
 ProviderIngressData = TypedDict("ProviderIngressData", {"url": str})
 # Provider application databag model.
 ProviderApplicationData = TypedDict("ProviderApplicationData", {"ingress": ProviderIngressData})  # type: ignore
 
 
-def _validate_data(data, schema):
+def _validate_data(data: Any, schema: Any):
```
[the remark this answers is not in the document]
PietroPasotti: Explicit is better than implicit.
But yes, might have taken more care into actually typing it out :)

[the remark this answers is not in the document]
PietroPasotti: unfortunately the data we validate is not str:str, but str:AnyJson. Will work it out.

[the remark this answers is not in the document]
PietroPasotti: not sure I'm making it better. the recursive type alias means some horrible shenanigans here and there. It's done but it's ugly.

tests/scenario/test_setup.py
PietroPasotti: [conftest.py](https://stackoverflow.com/questions/13641973/how-and-where-does-py-test-find-fixtures)

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -293,43 +308,74 @@ def wipe_ingress_data(self, relation: Relation):
...
+            # relation_broken events.
+            # FIXME https://github.com/canonical/traefik-k8s-operator/issues/34
+            return False
+
         try:
-            return bool(self._get_requirer_data(relation))
+            app_data, units_data = self._get_requirer_data(relation)
         except DataValidationError as e:
             log.warning("Requirer not ready; validation error encountered: %s" % str(e))
             return False
```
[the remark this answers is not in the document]
PietroPasotti: you mean, if for some reason `_get_requirer_data` were to return something that's not a 2-tuple unpackable thing?

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -491,34 +537,31 @@ def _publish_auto_data(self, relation: Relation):
...
         Args:
             host: Hostname to be used by the ingress provider to address the
              requirer unit; if unspecified, FQDN will be used instead
             port: the port of the service (required)
         """
-        # get only the leader to publish the data since we only
-        # require one unit to publish it -- it will not differ between units,
-        # unlike in ingress-per-unit.
-        assert self.unit.is_leader(), "only leaders should do this."
         assert self.relation, "no relation"
```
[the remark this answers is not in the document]
PietroPasotti: those are more sanity checks (or type guards) than anything: the caller should ensure that doesn't happen.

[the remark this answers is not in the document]
PietroPasotti: and we do want to know if that's the case, an early return is too silent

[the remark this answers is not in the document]
PietroPasotti: if anything, we should raise a proper exception, that I could agree with

tests/scenario/conftest.py
PietroPasotti: yup! I just learned that a couple of days ago and since then I've been refactoring them all over :)

tests/scenario/conftest.py
PietroPasotti: AFAIAO this code is equivalent to the with ... yield statement I have. I believe the with ... context calls start() on enter and stop() on exit. No need to manually cleanup when you have contexts. Which is why I love contexts.

tests/scenario/test_ingress_per_app.py
PietroPasotti: yes, thanks. I got so confused by that renaming thing...

tests/scenario/test_ingress_per_app.py
PietroPasotti: It would, but I'm trying to keep each test isolated.
I'm sure we'll refactor this once we change the schema and realize all of the tests break because we're overfitting the data in its current form, but for now I vote to keep it as-is

lib/charms/traefik_k8s/v0/ingress_per_leader.py
```diff
@@ -5,15 +5,16 @@
 
 This library wraps relation endpoints using the `ingress` interface
 and provides a Python API for both requesting and providing per-application
-ingress, with load-balancing occurring across all units.
+ingress, without load balancing across Traefik units: all traffic is routed
+through the leader instead.
```
sed-i, line 9: Should we further warn the reader?
```suggestion
through the leader instead.
This is not a typical ingress mode and is intended only for special cases.
```

lib/charms/traefik_k8s/v0/ingress_per_leader.py
```diff
@@ -28,26 +29,26 @@
 Then, to initialise the library:
 
 ```python
-from charms.traefik_k8s.v1.ingress import (IngressPerAppRequirer,
-  IngressPerAppReadyEvent, IngressPerAppRevokedEvent)
+from charms.traefik_k8s.v2.ingress import (IngressPerLeaderRequirer,
```
sed-i, line 32: Mismatch with folder name. Not sure if we should start from 0 though.
```suggestion
from charms.traefik_k8s.v0.ingress import (IngressPerLeaderRequirer,
```

lib/charms/traefik_k8s/v2/ingress.py
sed-i, line 1: Should we split this PR so the diff is clear?
If you think these should be in the same PR then I can download and diff locally.

lib/charms/traefik_k8s/v0/ingress_per_leader.py
PietroPasotti: umh, not sure what's going on. On local it's in v2.

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: how do you propose to split it? One PR adding ingress-per-leader, one PR adding ingress V2?

lib/charms/traefik_k8s/v0/ingress_per_leader.py
PietroPasotti: ah, my bad. It's in a comment. It should be `v0.ingress_per_leader`. Fixed in next commit

lib/charms/traefik_k8s/v2/ingress.py
sed-i: Yes, that would work. Or if you prefer, I could diff manually.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -0,0 +1,624 @@
...
+        """Constructor for IngressPerAppProvider.
+
+        Args:
+            charm: The charm that is instantiating the instance.
+            relation_name: The name of the relation endpoint to bind to
+                (defaults to "ingress").
+        """
+        super().__init__(charm, relation_name)
+
+    def _handle_relation(self, event):
```
sed-i, line 289: Seems outdated:
image

sed-i (commented):
Hey @PietroPasotti,
I posted a few screenshots for diffs that seem to be outdated, but there are more.
Mind taking another look at the file?

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: I broke this PR up into #190 and #191, to be reviewed and merged in that order, presumably.

Still not sure about how to go about merging them though, since this is a serious breakage of backwards compatibility.

PR #184, opened 2023-06-06
Fix itests

sed-i:
## Issue
Itests were relying on update-status.

## Solution
Do not rely on update-status in itests.

## Context
NTA.

## Testing Instructions
No functional changes. Passes locally. If passes in CI we're probably ok with this change.

## Release Notes
Do not rely on update-status in itests.

### DIFF
--- a/tests/integration/test_tcp_ipu_compatibility.py
+++ b/tests/integration/test_tcp_ipu_compatibility.py
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 import asyncio
 
+import pytest
 import pytest_asyncio
 from pytest_operator.plugin import OpsTest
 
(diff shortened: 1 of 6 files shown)


tests/integration/test_tcp_ipu_compatibility.py
```diff
@@ -19,6 +20,15 @@
 )
 
 
+@pytest.mark.abort_on_fail
```
PietroPasotti, line 23: I would move this to conftest.py and make it an autouse fixture.

```
@pytest.fixture(autouse=True, scope='module')
async def setup_env(ops_test: OpsTest):
    # ensure update-status does not fire "too" quickly, else traefik will flip between
    # active/idle and maintenance: updating ingress configuration
    await ops_test.model.set_config(
        {"update-status-hook-interval": "60m", "logging-config": "<root>=WARNING; unit=DEBUG"}
    )
```

do this, and it will run before every module is run, so we don't need to repeat in each file

PietroPasotti (changes requested):
The changes look good, but I'd refactor the config-setting 'test' to a fixture. Better practice, pytest-wise, and it seems like it's a good idea to do that globally.

Also I thought the itests were fixed in yesterday's merge, did you find more issues, or how does this pr come to be?

PietroPasotti (approved):

PR #190, opened 2023-06-23
ingress(-per-app) v2

PietroPasotti:
This PR fixes the "ingress" library to provide ingress-per-app instead of ingress-per-leader, under the same name "ingress".

Ingress-per-leader functionality is thereby REMOVED.
A [separate pr](https://github.com/canonical/traefik-k8s-operator/pull/191) will introduce back the ingress-per-leader functionality under the name "ingress-per-leader"

Given the non-backwards-compatible nature of the changes, this means releasing a new ingress v2 lib

Changes:
=======

- added schema for ingress relation unit databag


sed-i:
Something's odd in the diff here:
- There seem to be too many conflicting files
- There seem to be too many files changed

PietroPasotti:
was missing a rebase on main. @sed-i should be ready now

metadata.yaml
```diff
@@ -49,6 +49,10 @@ provides:
     interface: ingress_per_unit
     description: |
       Provides ingress-like routing to the single units of the related Juju application
+  ingress-per-leader:
+    interface: ingress_per_leader
+    description: |
+      Provides ingress-like routing to only the leader of the related Juju application
```
sed-i, line 55: Needs to be dropped from _this_ PR :)

src/charm.py
```diff
@@ -634,19 +633,22 @@ def _provide_routed_ingress(self, relation: Relation):
         self._push_configurations(relation, config)
 
     def _provide_ingress(
-        self, relation: Relation, provider: Union[IngressPerAppProvider, IngressPerAppProvider]
+        self,
+        relation: Relation,
+        provider: Union[IngressPerAppProvider, IngressPerAppProvider],
```
sed-i, line 638: Elements of Union are identical.
Interesting that it wasn't picked up.

src/charm.py
```diff
@@ -655,12 +657,14 @@ def _get_configs_per_app(self, relation: Relation):
         provider = self.ingress_per_app
 
         try:
-            data: "RequirerData_IPA" = provider.get_data(relation)
+            data: Tuple[
+                "RequirerAppData_IPA", Iterable["RequirerUnitData_IPA"]
+            ] = provider.get_data(relation)
```
sed-i, line 662: Why do we type-hint here if `get_data` is typed?

tests/integration/test_charm_tcp.py
```diff
@@ -21,6 +21,13 @@
 }
 
 
+@pytest.mark.abort_on_fail
+async def test_setup_env(ops_test: OpsTest):
```
sed-i, line 25: I thought we already have this as an autouse fixture.

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,173 @@
...
+from pathlib import Path
+
+import pytest
+import yaml
+from ops import pebble
+from scenario import Container, Context, Model, Mount, Relation, State
+
+
+@pytest.fixture
+def context(traefik_charm):
```
sed-i, line 17: Do we need this given `traefik_ctx` in conftest?

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,173 @@
...
+    return Context(charm_type=traefik_charm)
+
+
+@pytest.fixture
+def model():
+    return Model(name="test-model")
+
+
+@pytest.fixture
+def temp_opt():
```
sed-i, line 27: Seems like this functionality is [built-in](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html).

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,173 @@
...
+        name="traefik",
+        can_connect=True,
+        layers={"traefik": layer},
+        service_status={"traefik": pebble.ServiceStatus.ACTIVE},
+        mounts={"opt": opt},
+    )
+
+
+@pytest.mark.parametrize("port, host", ((80, "1.1.1.1"), (81, "10.1.10.1")))
+@pytest.mark.parametrize("event_name", ("joined", "changed"))
```
sed-i, line 61: This would result in a matrix test with combos we're not entirely interested in: do we really spec the same outcome for both `joined` and `changed` given the same context?

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,173 @@
...
+                },
+            },
+            "services": {
+                "juju-test-model-remote-0-service": {
+                    "loadBalancer": {"servers": [{"url": f"http://{host}:{port}"}]}
+                }
+            },
+        }
+    }
+    cfg_file.write_text(yaml.safe_dump(initial_cfg))
```
sed-i, line 129: File is dumped before `{host}:{port}` are rendered?

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -95,12 +95,19 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
...
+    "required": ["model", "name"],
+}
+
+INGRESS_REQUIRES_UNIT_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "host": {"type": "string"},
+        "port": {"type": "integer"},
+    },
+    "required": ["host", "port"],
```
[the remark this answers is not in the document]
PietroPasotti: not going to solve this here, no

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -275,15 +288,14 @@ def _handle_relation(self, event):
         # created, joined or changed: if remote side has sent the required data:
         # notify listeners.
         if self.is_ready(event.relation):
-            data = self._get_requirer_data(event.relation)
+            leader_data, followers_data = self.get_data(event.relation)
```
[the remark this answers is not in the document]
PietroPasotti: fixed

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -303,40 +315,61 @@ def wipe_ingress_data(self, relation: Relation):
...
 
-        For convenience, we convert 'port' to integer.
+        We cast port to int for convenience.
         """
-        if not relation.app or not relation.app.name:  # type: ignore
-            # Handle edge case where remote app name can be missing, e.g.,
-            # relation_broken events.
-            # FIXME https://github.com/canonical/traefik-k8s-operator/issues/34
-            return {}
+        out: List[RequirerUnitData] = []
```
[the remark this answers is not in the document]
PietroPasotti: yeah because the object we're appending to `out` is an (almost untyped) dict. Working with typeddicts can be annoying at times.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -303,40 +315,61 @@ def wipe_ingress_data(self, relation: Relation):
...
+            remote_unit_data["port"] = (
+                int(remote_unit_data["port"]) if remote_unit_data["port"] else None
+            )
+            _validate_data(remote_unit_data, INGRESS_REQUIRES_UNIT_SCHEMA)
+            out.append(typing.cast(RequirerUnitData, remote_unit_data))
+        return out
+
+    def _get_requirer_leader_data(self, relation: Relation) -> RequirerAppData:
+        """Fetch and validate the requirer's app databag."""
+        app: Optional[Application] = relation.app
```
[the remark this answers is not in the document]
PietroPasotti: you're right, we don't

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -303,40 +315,61 @@ def wipe_ingress_data(self, relation: Relation):
...
+        if app is None:
+            raise NotReadyError(relation)
```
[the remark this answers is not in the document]
PietroPasotti: it's too deep in the library code to surface to the charm in any other meaningful manner I feel, and I'd rather people code defensively and do `if ingress.is_ready: data=ingress.get_data` rather than write their code to handle a possible None being returned from `get_data`.

But if we don't like it we can revert

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -303,40 +315,61 @@ def wipe_ingress_data(self, relation: Relation):
...
 
-        databag = relation.data[relation.app]
-        remote_data: Dict[str, Union[int, str]] = {}
-        for k in ("port", "host", "model", "name", "mode", "strip-prefix", "redirect-https"):
+        databag = relation.data[app]
+        remote_app_data: Dict[str, Union[int, str]] = {}
+
+        for k in ("model", "name", "mode", "strip-prefix"):
```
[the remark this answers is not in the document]
PietroPasotti: uh, good catch.

src/charm.py
PietroPasotti: ha.

tests/scenario/test_ingress_per_app.py
PietroPasotti: yeah but pytest-operator breaks it 
image

however it does work when running it from tox as it does not install pytest-operator.

tests/scenario/test_ingress_per_app.py
PietroPasotti: I don't think so, that's an f-string. or are you referring to something else?

tests/scenario/test_ingress_per_app.py
PietroPasotti: is there such a thing as too much testing? 

Anywhoos, the ingress libs will emit 'data_provided' on changed, joined and created.
So actually we're missing 'created'.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -0,0 +1,629 @@
+# Copyright 2022 Canonical Ltd.
```
Abuelodelanada, line 1:
```suggestion
# Copyright 2023 Canonical Ltd.
```

Abuelodelanada (approved):

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -0,0 +1,629 @@
...
+                "Probably a ghost of a dead relation is still "
+                "lingering around.".format(e, relation.name)
+            )
+            return
+        del relation.data[self.app]["ingress"]
+
+    def _get_requirer_units_data(self, relation: Relation) -> List[RequirerUnitData]:
+        """Fetch and validate the requirer's app databag.
+
+        We cast port to int for convenience.
```
sed-i, line 330: Should probably use "convert" or similar because "cast" is a no-op at runtime.
```suggestion
        We convert port to int for convenience.
```

sed-i: Or entirely remove this comment :D

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -0,0 +1,629 @@
...
+        """
+        out: List[RequirerUnitData] = []
+
+        unit: Unit
+        for unit in relation.units:
```
sed-i, line 335: (Blacking out on the details here for a sec)
Here, `relation` is for a particular `rel-id`, correct?
I.e. the units in `relation.units` are all from the same app?
I.e. this still works as expected if two different apps are related over the same relation name?

tests/scenario/test_ingress_per_app.py
```diff
@@ -0,0 +1,162 @@
...
+        return {
+            "port": str(port + n),
+            "host": host.format(n),
+        }
+
+    ipa = Relation(
+        "ingress",
+        remote_app_data={
+            "model": "test-model",
+            "name": "remote/0",
```
sed-i, line 130: I'm a bit confused:
1. Who is being scaled here - traefik or the app asking for an ingress url?
2. Why app name has a unit index? And why was it converted to dash by IPA (`remote-0`)

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -0,0 +1,629 @@
...
+    total=False,
+)
+
+# Model of the data each requirer unit will need to provide.
+RequirerUnitData = TypedDict("RequirerUnitData", {"host": str, "port": int})
+
+# Provider ingress data model.
+ProviderIngressData = TypedDict("ProviderIngressData", {"url": str})
+# Provider application databag model.
+ProviderApplicationData = TypedDict("ProviderApplicationData", {"ingress": ProviderIngressData})  # type: ignore
```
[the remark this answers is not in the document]
PietroPasotti: no, but they can be made private. They're not meant for external usage, it's really about typing the raw data structures.

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: yes, I think the units from one relation should not leak to another relation's

tests/scenario/test_ingress_per_app.py
PietroPasotti: 1. the app asking for an ingress url is being scaled: you can see that from remote_units_data being populated by `n` databags
2. very sharply spotted. that's a leftover from a IPU test probably. It was converted to a dash because IPA and IPU prefixes are handled the same, and charm.TraefikIngressCharm._get_prefix converts slash to dash to normalize unit names.

PietroPasotti:
@sed-i your comments on the typeddicts prompted me to ditch jsonschema and go full on pydantic.
It looks a lot better now, and I think that that `DatabagIOMixin` I added could be reusable in a bunch of places

PietroPasotti:
todo (for me): test against https://github.com/canonical/charm-relation-interfaces/pull/85

PR #197, opened 2023-07-12
Reduce the number of API calls to K8S

javacruft:
The charm makes a large number of API calls to retrieve information about the IP address allocated to the K8S service; cache this information to reduce the number of calls made and speed up hook performance.

## Issue
Improve the performance of hook executions

## Solution
Cache response from K8S to reduce the amount of API calls made.

## Context
The traefik charm stands out as always the last to settle in Sunbeam deployments; hopefully this may improve this situation a bit.

## Testing Instructions
https://microstack.run/docs -> "multi-node tutorial"

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,6 +5,7 @@
 """Charm Traefik."""
 import contextlib
 import enum
+import functools
 import ipaddress
 import json
 import logging
@@ -1271,6 +1272,7 @@ class TraefikIngressCharm(CharmBase):
         ]
 
 
+@functools.lru_cache
 def _get_loadbalancer_status(namespace: str, service_name: str):
     client = Client()  # type: ignore
     traefik_service = client.get(Service, name=service_name, namespace=namespace)


javacruft:
new in py39 - drat!

Let me take another run at this.

PietroPasotti:
@javacruft seems like performance has been decreasing lately, we can consider merging this work, what's the status?

javacruft:
@PietroPasotti refreshed for 3.8 compatibility - I'm taking it for a test drive now

simskij (approved):
Should be relatively risk-free as the charm exits after every invocation, meaning that the caching will only be done throughout the lifetime of a single event.

PietroPasotti:
itests failing due to some ingress gateway unavailable

javacruft:
@PietroPasotti any idea what the cause of the gateway unavailable error is?   I had a dig around in the logs but can't see anything obvious

PietroPasotti:
> @PietroPasotti any idea what the cause of the gateway unavailable error is? I had a dig around in the logs but can't see anything obvious

no idea either: https://discourse.charmhub.io/t/traefik-k8s-docs-troubleshooting-gateway-address-unavailable/10813

PietroPasotti:
> @PietroPasotti any idea what the cause of the gateway unavailable error is? I had a dig around in the logs but can't see anything obvious

't was flaky

PR #196, opened 2023-07-05
Ingress/scheme

PietroPasotti:
Based on #190 

Fixes #188 

In tandem with: https://github.com/canonical/alertmanager-k8s-operator/pull/173

PR #199, opened 2023-07-13
implement cert_transfer (misnomer mTLS interface)

lucabello:
## Issue
Users need to be able to programmatically extract **traefik**'s certificate (and possibly the CA) to trust it.

## Solution
Use the [`mutual_tls`](https://github.com/canonical/mutual-tls-interface/blob/main/lib/charms/mutual_tls_interface/v0/mutual_tls.py) interface, which (besides what the - wrong - name suggests) helps transfer the Provider's certificate over relation data to the Requirers.

## Possible TODOs
- [ ] Ideally add an integration test with a Requirer and test it (grafana-agent will soon implement this in a parallel PR)

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -67,6 +67,7 @@ provides:
     interface: mutual_tls
     description: |
       Sends the TLS information (certificate, ca, chain) over relation data.
+      Note: The interface name is a misnomer, as this is not in fact mTLS.
 
 requires:
   certificates:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -63,7 +63,10 @@ provides:
     description: |
       Forwards the built-in grafana dashboard(s) for monitoring traefik.
     interface: grafana_dashboard
-
+  mutual-tls:
+    interface: mutual_tls
+    description: |
+      Sends the TLS information (certificate, ca, chain) over relation data.
 
 requires:
   certificates:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -63,7 +63,7 @@ provides:
     description: |
       Forwards the built-in grafana dashboard(s) for monitoring traefik.
     interface: grafana_dashboard
-  mutual-tls:
+  cert-transfer:
     interface: mutual_tls
     description: |
       Sends the TLS information (certificate, ca, chain) over relation data.
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,6 +17,7 @@ from urllib.parse import urlparse
 import yaml
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
+from charms.mutual_tls_interface.v0.mutual_tls import MutualTLSProvides
 from charms.observability_libs.v1.kubernetes_service_patch import (
     KubernetesServicePatch,
     ServicePort,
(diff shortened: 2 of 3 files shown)


lucabello:
(note that the **Inclusive Naming check** CI will fail due to "Dummy" being present in the `mutual_tls` library)

metadata.yaml
```diff
@@ -63,7 +63,10 @@ provides:
     description: |
       Forwards the built-in grafana dashboard(s) for monitoring traefik.
     interface: grafana_dashboard
-
+  cert-transfer:
```
sed-i, line 66: Should we use terminology similar to `send-remote-write` and `receive-remote-write`?

metadata.yaml
```diff
@@ -63,7 +63,10 @@ provides:
...
+    interface: mutual_tls
+    description: |
+      Sends the TLS information (certificate, ca, chain) over relation data.
```
sed-i, line 69:
```suggestion
      Sends the TLS information (certificate, ca, chain) over relation data.
      Note: The interface name is a misnomer, as this is not in fact mTLS.
```

src/charm.py
```diff
@@ -303,6 +310,13 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
         )
         self._stored.csr = new_csr.decode()
 
+    def _on_cert_transfer_relation_joined(self, event: RelationJoinedEvent):
+        if self.certificates.cert:
+            certificate = self.certificates.cert
+            ca = self.certificates._ca_cert
+            chain = self.certificates._chain
```
sed-i, line 317: I wonder why these are private. Perhaps raise an issue on the tls-cert lib repo?

src/charm.py
```diff
@@ -303,6 +310,13 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+            self.mutual_tls.set_certificate(certificate=certificate, ca=ca, chain=chain)
```
sed-i, line 318: Is this stored in unit data?
Mind confirming scaling the app works?

src/charm.py
```diff
@@ -303,6 +310,13 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
         )
         self._stored.csr = new_csr.decode()
 
+    def _on_cert_transfer_relation_joined(self, event: RelationJoinedEvent):
```
sed-i, line 313: We also need cleanup on departed.

src/charm.py
PietroPasotti: kinda related: we forgot it as well in certhandler. https://github.com/canonical/observability-libs/pull/54

PietroPasotti:
closing this one as the conflict surface area with #196 is too large and this PR to small to be rebase-worthy.
Copying the relevant parts over to a new PR since traefik is now using certhandler

src/charm.py
PietroPasotti: it is stored in unit data, not sure i see why it wouldn't work with scaling, but will give it a go as soon as I find a charm to integrate this with

metadata.yaml
PietroPasotti: we can always rename that later, but what would you propose?
`send-certificate` feels OK, but we also have a `certificates` endpoint, should we call that `receive-certificate`? however that is not the counterpart to this one.
Do we want `generate-certificate`, `send-ceritficate`, `receive-certificate`? Feels a bit confusing. Let's agree on this another time :)

PR #200, opened 2023-07-14
Ingressv1 backwards compat

PietroPasotti:
adds ingress v1 backwards compat on top of ingress v2


lucabello (approved):

src/charm.py
```diff
@@ -634,35 +641,60 @@ def _provide_routed_ingress(self, relation: Relation):
...
 
+    def _get_configs_per_leader(self, relation: Relation):
+        """Generates ingress per leader config."""
+        # this happens to be the same behaviour as ingress v1 (legacy) provided.
+        ipa = self.ingress_per_appv1
+
+        try:
+            data = ipa.get_data(relation)
+        except DataValidationError as e:
+            logger.error(f"invalid data shared through {relation}... Error: {e}.")
```
sed-i, line 677: Would `relation.name` be better here?

src/charm.py
```diff
@@ -917,12 +967,26 @@ def _provider_from_relation(self, relation: Relation):
         """Returns the correct IngressProvider based on a relation."""
         relation_type = _get_relation_type(relation)
         if relation_type is _IngressRelationType.per_app:
-            return self.ingress_per_app
+            # first try to tell if remote is speaking v2
+            if self.ingress_per_appv2.is_ready(relation):
```
sed-i, line 971: Any chance `is_ready` could be false positive in the future?
Should we use `try ... get_data` instead?

sed-i (approved):
This is great!

PR #202, opened 2023-07-18
removed leader guard

PietroPasotti:
now that ingress requirer also expose unit data, we need to remove a leader guard preventing auto-data from being published on joined and changed without charm intervention.

Fixes a bug discovered by @gregory-schiano


dstathis (approved):

gregory-schiano (approved):
LGTM

PR #207, opened 2023-07-25
added tracing

PietroPasotti:
adds charm tracing and tempo tracing integration from builtin traefik instrumentation


metadata.yaml
```diff
@@ -63,7 +63,11 @@ provides:
     description: |
       Forwards the built-in grafana dashboard(s) for monitoring traefik.
     interface: grafana_dashboard
-
+  tracing:
```
sed-i, line 66: I forget, does the server must be the provider?
If yes, then it seems like traefik would need to be the requirer side?

pyproject.toml
```diff
@@ -1,5 +1,14 @@
...
 # See LICENSE file for licensing details.
+[project]
+name = "traefik-k8s"
+version = "0.1"  # this is in fact irrelevant
+
+[project.optional-dependencies]
+lib_pydeps = [
+    "opentelemetry-exporter-otlp-proto-grpc",
+    "pydantic<2.0"
+]
```
sed-i, line 11: Our deps are already split into reqs.txt, tox.ini, charmcraft.yaml, pydeps.
Do we need this one as well?

tests/scenario/test_charm_tracing.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # THEN we get some traces
+    with patch("opentelemetry.exporter.otlp.proto.grpc.exporter.OTLPExporterMixin._export") as f:
+        f.return_value = opentelemetry.sdk.metrics._internal.export.MetricExportResult.SUCCESS
+        # WHEN traefik receives <any event>
+        traefik_ctx.run(tracing.changed_event, state_in)
+
+    assert "Setting up span exporter to endpoint: foo.com:81" in caplog.text
+    assert "Starting root trace with id=" in caplog.text
+    span = f.call_args_list[0].args[0][0]
```
sed-i, line 25: Anything pretty we should/can do with the `[0]`?

src/charm.py
```diff
@@ -353,6 +403,41 @@ def _get_tls_config(self) -> dict:
             }
         }
 
+    def _get_tracing_config(self) -> dict:
+        """Return dictionary with opentelemetry configuration if available."""
+        # ref: https://doc.traefik.io/traefik/ma<woke forced me to add this>ster/observability/tracing/opentelemetry/
+        if not self._is_tracing_enabled():
+            logger.error("tracing not enabled")
```
sed-i, line 410: Without tempo, will we see an error in traefik logs every juju event?

src/charm.py
```diff
@@ -353,6 +403,41 @@ def _get_tls_config(self) -> dict:
...
+            logger.error(
+                "tracing integration is active but none of the "
+                "protocols traefik supports is available."
+            )
+            return {}
+
+        otlp_cfg: Dict[str, Any] = {"address": addr}
+        if self._is_tls_enabled():
+            # todo: we have an option to use CA or to use CERT+KEY authentication.
+            #  what is the difference and which one is better?
```
sed-i, line 428: Cert+key is probably mTLS. We're not going there yet.

src/charm.py
```diff
@@ -353,6 +403,41 @@ def _get_tls_config(self) -> dict:
...
+            otlp_cfg["ca"] = _CA_CERT_PATH
+        else:
+            otlp_cfg["insecure"] = True
```
sed-i, line 431: Not sure what this `else` means here. Who is the server and who is the client?

I wonder if:
- we should always set "insecure" to True
- we shouldn't set "ca" because we assume the other side (or us) already has it from `update-ca-certificates`

src/charm.py
```diff
@@ -353,6 +403,41 @@ def _get_tls_config(self) -> dict:
...
+
+            # todo: determine when and if we have to use insecureSkipVerify
+            otlp_cfg["tls"] = {"insecureSkipVerify": True}
```
sed-i, line 434: What's the diff between `insecureSkipVerify` here and `insecure` above?

src/charm.py
```diff
@@ -1085,10 +1181,10 @@ def _scrape_jobs(self) -> list:
 
 
 def _get_loadbalancer_status(namespace: str, service_name: str):
-    client = Client()
+    client = Client()  # type: ignore
     traefik_service = client.get(Service, name=service_name, namespace=namespace)
 
-    if status := traefik_service.status:
+    if status := traefik_service.status:  # type: ignore
```
sed-i, line 1187: I'm curious why, from one PR to another, a few hours apart, you suddenly need all the `type: ignore`s.

src/charm.py
```diff
@@ -353,6 +403,41 @@ def _get_tls_config(self) -> dict:
             }
         }
 
+    def _get_tracing_config(self) -> dict:
+        """Return dictionary with opentelemetry configuration if available."""
+        # ref: https://doc.traefik.io/traefik/ma<woke forced me to add this>ster/observability/tracing/opentelemetry/
```
lucabello, line 408: I'd say change this to a space or a dot and add a comment below ?

src/charm.py
PietroPasotti: uh good point, I'll tune this down.

metadata.yaml
PietroPasotti: no it's correct like this. 
However I also get it mixed up all the time, we should consider changing the directionality.

tests/scenario/test_charm_tracing.py
PietroPasotti: I'm open to suggestions. I also have a feeling it's brittle, but that part of the api is a bit obscure and I'm not sure how to make it more robust.

src/charm.py
PietroPasotti: insecureskipverify = the TLS connection to the [other end] accepts any certificate presented by the server regardless of the hostnames it covers.

src/charm.py
PietroPasotti: setting insecure to True is an error if we also present "ca"
"ca" (or cert+key) are required if insecure=False (the default).

pyproject.toml
PietroPasotti: so I have 3+ tox envs that need to install the pydeps for all of the charm libs we have.
I'd rather have the pydeps-inherited-dependencies all in one place, and I'm not sure how else to do that without repeating myself in all tox envs that need them (and consequently being bothered keeping them in sync)

PietroPasotti: would you rather repeat them in all tox envs? what if we add one more charm lib with its pydeps?

src/charm.py
sed-i: Great. So what's `insecure` above?

pyproject.toml
sed-i: Sounds like these need to go into `requirements.txt` in the first place, no?

PietroPasotti: I thought the whole point of pydeps was to be able to avoid that

src/charm.py
PietroPasotti: that's not in place, not that I know of. is that part of the mTLS story?

src/charm.py
PietroPasotti: accept http. Without insecure=true, it will complain if it cannot connect over https.

src/charm.py
```diff
@@ -353,6 +418,44 @@ def _get_tls_config(self) -> dict:
             }
         }
 
+    def _get_tracing_config(self) -> dict:
+        """Return dictionary with opentelemetry configuration if available."""
+        # ref: https://doc.traefik.io/traefik/ma<woke>ster/observability/tracing/opentelemetry/
+        #  note: remove the <woke> tag
```
sed-i, line 424: @ru-fu pointed out that woke supports [inline excludes](https://docs.getwoke.tech/ignore/#in-line-and-next-line-ignoring)

```suggestion
        # wokeignore:rule=master
        # ref: https://doc.traefik.io/traefik/master/observability/tracing/opentelemetry/
```

pyproject.toml
sed-i: Yeah, we don't need it at pack time because charmcraft pulls it from pydeps, but if we already need it for the test envs, then I think we should:
1. Add them to requriements.txt with a comment `# the deps below are from pydeps and needed here to tox test envs`; or
2. Move _all_ tox deps into one place instead of splitting into yet another place.

PietroPasotti: image
<cries in python>

src/charm.py
```diff
@@ -214,12 +245,40 @@ def _is_tls_enabled(self) -> bool:
...
+    def _on_tracing_endpoint_removed(self, event) -> None:
+        if not self.container.can_connect():
+            event.defer()
+            return
+        self._clear_tracing_config()
+
+    def _on_tracing_endpoint_changed(self, event) -> None:
+        # On slow machines, this event may come up before pebble is ready
+        if not self.container.can_connect():
+            event.defer()
```
sed-i, line 264: The logic captured in this defer will not be executed if we upgrade soon after deploying (and relation events do not re-fire on upgrade) so tracing config won't "ever" be written.

src/charm.py
sed-i: It's part of the TLS story. @lucabello has 2 merged examples at this point?

src/charm.py
sed-i: Had a chat with @simskij and we said we shouldn't have insecureSkipVerify anywhere.

src/charm.py
```diff
@@ -366,12 +469,20 @@ def _clear_all_configs_and_restart_traefik(self):
         # configuration files on a storage volume that survives the pod churn, before
         # we start traefik we clean up all Juju-generated config files to avoid spurious
         # routes.
+        self._clear_tracing_config()
```
sed-i, line 472: Should we instead move all config to non-persistent storage?

src/charm.py
```diff
@@ -612,26 +730,26 @@ def _get_configs_per_leader(self, relation: Relation) -> Optional[dict]:
...
-        config = self._generate_per_leader_config(prefix, data)
+        prefix = self._get_prefix(data)  # type: ignore
+        config = self._generate_per_leader_config(prefix, data)  # type: ignore
         if self.unit.is_leader():
             ipa.publish_url(relation, self._get_external_url(prefix))
 
         return config
 
-    def _get_configs_per_app(self, relation: Relation) -> Optional[dict]:
+    def _get_configs_per_app(self, relation: Relation) -> Dict[str, Any]:
```
sed-i, line 742: Should we have a new recursive type e.g. StrDict?
```python
StrDict = Dict[str, Union[str, "StrDict"]]
```

src/charm.py
```diff
@@ -667,14 +785,14 @@ def _get_configs_per_unit(self, relation: Relation) -> Optional[dict]:
                 )
                 continue
 
-            prefix = self._get_prefix(data)
+            prefix = self._get_prefix(data)  # type: ignore
```
sed-i, line 788: The `type: ignore` is becoming a burden.
- Would pinning pyright help?
- Is there a pyright setting to error out on unused ignores?

src/charm.py
```diff
@@ -962,7 +1080,9 @@ def _restart_traefik(self):
 
         if _TRAEFIK_SERVICE_NAME not in current_services:
             self.unit.status = MaintenanceStatus(f"creating the {_TRAEFIK_SERVICE_NAME!r} service")
-            self.container.add_layer(_TRAEFIK_LAYER_NAME, layer, combine=True)
+            self.container.add_layer(
+                _TRAEFIK_LAYER_NAME, typing.cast(LayerDict, layer), combine=True
```
sed-i, line 1084: Seems wrong that we need the cast here.

src/charm.py
PietroPasotti: - pinning pyright would help.
- yes, and it is the default.

PietroPasotti: (I'm pinning pyright as part of this pr)

src/charm.py
PietroPasotti: either here, or at layer instantiation. Pyright doesn't like `foo:LayerDict = {...}` either, for reasons unknown. It's too complex a type. Too many optional fields

src/charm.py
PietroPasotti: elsewhere I had an AnyJson type, probably a better fit as there can be booleans, arrays and integers as well

PietroPasotti: all type checkers are fussy with recursive types either way in my experience. 
Gave it half an hour to try out adding an AnyJson type but it got way worse. Dropping it for now, if we like we can have a separate pr to improve the typing of traefik as a whole, but that'd feel like feature creep in this context.

metadata.yaml
```diff
@@ -63,7 +63,11 @@ provides:
...
+    description: |
+      Tracing endpoint for integrating with tempo.
+    limit: 1
```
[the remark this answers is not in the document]
PietroPasotti: was not aware of the 'server should be provider' discussion. This will require changing the directionality of the tracing relation, submitting a new pr to CRI, etc...

src/charm.py
```diff
@@ -137,18 +147,21 @@ def __init__(self, *args):
...
-                self.traefik_route.on.ready,
-                self.traefik_route.on.data_removed,
+                ipa_v1.on.data_provided,  # type: ignore
+                ipa_v2.on.data_provided,  # type: ignore
+                ipa_v1.on.data_removed,  # type: ignore
+                ipa_v2.on.data_removed,  # type: ignore
+                self.ingress_per_unit.on.data_provided,  # type: ignore
+                self.ingress_per_unit.on.data_removed,  # type: ignore
+                self.traefik_route.on.ready,  # type: ignore
+                self.traefik_route.on.data_removed,  # type: ignore
```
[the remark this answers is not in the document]
PietroPasotti: I don't think that's a nice way to solve it in this case, and I'd rather make do with what we have until https://github.com/canonical/operator/pull/982 for example pans out

metadata.yaml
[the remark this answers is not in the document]
PietroPasotti: relating to multiple tempo apps is not supported by traefik. We can configure traefik with ONE tracing endpoint to send traces to.

I don't see how tempo having multiple units is of interest to Traefik, can you explain?

src/charm.py
```diff
@@ -108,7 +116,7 @@ def __init__(self, *args):
             self,
             key="trfk-server-cert",
             peer_relation_name="peers",
-            extra_sans_dns=[self.cert_subject],
+            extra_sans_dns=[self.cert_subject] if self.cert_subject else [],
```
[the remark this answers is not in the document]
PietroPasotti: this is a bug for CertHandler. Opened it there.

src/charm.py
PietroPasotti: on upgrade, we'll run `_clear_all_configs_and_restart_traefik` which in turn calls `_push_config` that in turn calls `_push_tracing_config`. So it's all right.

What I get from this is that my feeling that traefik's config management is getting too complex is confirmed. We should consider refactoring it to use manifests

src/charm.py
```diff
@@ -214,12 +245,40 @@ def _is_tls_enabled(self) -> bool:
...
+        """Return True if tracing is enabled."""
+        eps = self._tracing.endpoints
+        if not eps or not eps.host or not eps.ingesters:
+            return False
+        return True
+
+    def _on_tracing_endpoint_removed(self, event) -> None:
+        if not self.container.can_connect():
+            event.defer()
+            return
```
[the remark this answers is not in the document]
PietroPasotti: - the upgrade issue is addressed by my reply to the other comment
- what if there's a random container churn in the middle of a 'regular' `juju remove-relation` event flow? then deferring could still have sense as the charm container is still alive and will run this event again when it receives the next event. But yeah, this is complex to reason about.
- if re-emitted, all container-related logic will be on hold either way, so I'm not sure I see the danger. 

traefik is already doing some sort of common exit hook: `process_status_and_configurations`, maybe we could consolidate all that's not in there yet.

PietroPasotti:
in light of https://github.com/canonical/charm-relation-interfaces/pull/92, I propose we merge this one as-is, an open a follow-up replacing provider with requirer

lucabello (approved):

src/charm.py
PietroPasotti: also see: https://github.com/canonical/traefik-k8s-operator/pull/213

metadata.yaml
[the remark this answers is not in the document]
sed-i: > how tempo having multiple units is of interest to Traefik

Just confirming if there are any assumptions in the code, e.g. how we get/iterate relations.

sed-i (commented):
Are we to rebase this on top of the direction flipping PR?

PR #220, opened 2023-08-14
Introduce 'ip' field to unit data in ingress relation v2

weiiwang01:
## Issue

The current version of `charms.traefik_k8s.v1.ingress` by default provides only the unit's FQDN for ingress endpoints. However, there are limitations to this approach:
1. Some ingress providers, such as Kubernetes nginx-ingress, do not support FQDN endpoints.
2. FQDN might not be resolvable in certain scenarios, such as within machine charms or k8s charms when cluster DNS is disabled.

## Solution
This pull request introduces the unit's IP address to the ingress relation v2 unit databag. This acts as an alternative means of addressing for ingress endpoints. By default, the system will retrieve the IP address using the Juju network APIs.


lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -646,9 +663,22 @@ def provide_ingress_requirements(self, *, host: Optional[str] = None, port: int)
...
+                raise RuntimeError(
+                    f"can't find IP address for relation {self.relation}, network binding is None"
+                )
+            binding_ip = network_binding.network.bind_address
+            if binding_ip is None:
+                raise RuntimeError(
+                    f"can't find IP address for relation {self.relation}, bind address is None"
+                )
+            ip = str(binding_ip)
+
```
simskij, line 678: This raise is a bit dangerous as it's quite common for sidecar charms to not yet have a bind address during the first events executed. This could lead to the charm getting stuck in an error state.

simskij (changes requested):

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: How about use `socket.gethostbyname(socket.gethostname())` only when the value isn't in the `127/8` range. This will likely decrease the risk of errors, though it won't entirely eliminate them.

simskij: I think even in the case of being served a local loopback, we should go with that. Once it has access to a bind address, this would auto-heal, which I much prefer to it being **completely** broken in the meantime.

simskij: > I've considered it, but I think the issue is primarily with Juju. When the charm responds to an event, the network should already be ready. And I am not sure if we should workaround that in the charm level.

Well, it would be a breaking change to current users as their ingress would potentially stop working even though they might not even use the ip (which is the case with traefik itself). Unfortunately, that will be the case regardless of whos "fault" it actually is. 😅

weiiwang01: I'm concerned about scenarios involving ingress controllers like Traffic or NGINX serving multiple applications. I'm not completely certain about the precise implementation, but if the ingress interprets the loopback address as-is, it could create a short circuit within the ingress. This might redirect traffic to a different application, potentially bypassing certain security measures.

For instance, if a requirer provide `127.0.0.1` as the endpoint address, it might generate a rule like the one below. This rule could route traffic from port `80` to endpoint `http://internal-app:8080`:

```
server {
  listen 8080;
  proxy_pass http://internal-app:8080;
}

server {
  listen 80;
  proxy_pass http://127.0.0.1:8080;
}
```

simskij: > I'm concerned about scenarios involving ingress controllers like Traffic or NGINX serving multiple applications. I'm not completely certain about the precise implementation, but if the ingress interprets the loopback address as-is, it could create a short circuit within the ingress. This might redirect traffic to a different application, potentially bypassing certain security measures.
> 
> For instance, if a requirer provide `127.0.0.1` as the endpoint address, it might generate a rule like the one below. This rule could route traffic from port `80` to endpoint `http://internal-app:8080`:
> 
> ```
> server {
>   listen 8080;
>   proxy_pass http://internal-app:8080;
> }
> 
> server {
>   listen 80;
>   proxy_pass http://127.0.0.1:8080;
> }
> ```

Then I suggest we make it so that if it's getting a loopback (i.e `localhost`, `127.0.0.1` or `::1`, it instead sets it up to proxy pass to `about:blank`.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -534,12 +543,13 @@ def __init__(
         relation_name: str = DEFAULT_RELATION_NAME,
         *,
         host: Optional[str] = None,
+        ip: Optional[str] = None,
```
sed-i, line 546: Do we need both ip and and host?
Can we instead have something like
```suggestion
        host: Optional[Union[str, ipaddress.IPv4Address]] = None,  # TODO in v3, rename to address
```
- If None, use fqdn;
- If able to figure out that the "fqdn" is not valid (no cluster DNS or machine charm) then get bind address.

lib/charms/traefik_k8s/v2/ingress.py
sed-i: Hopefully the `ip` field can be merged with the `host` (see comment above) and then:
- If we keep using fqdn as default then this issue goes away
- If we start using bind_address as a fallback for invalid fqdn then need a custom exception and add a `Raises:` section to the docstring.

> it's quite common for sidecar charms to not yet have a bind address during the first events executed

Seems like asking the caller to guard for these Nones would be more messy than try-except.

simskij: > Seems like asking the caller to guard for these Nones would be more messy than try-except.

Well, the problem is going to be that it might have the charm go into a crash loop if the information isn't available when it checks. I at least can't for certain say that Juju will proceed with assigning an IP if the charm is stuck in `error`.

sed-i: Yeah it's an avenue we'd need to explore, but I hope we won't get there as I imagine we can have both merged into one arg, and always use fqdn as the only default. IP would have to be via kwarg.

PietroPasotti: +1 to what leon said. Don't like the idea of adding it as a separate field to ingress.
We can consider wrapping that try/except in the library call, so the user doesn't have to remember doing it manually?

weiiwang01: Unfortunately, we cannot provide only the FQDN without the IP. This is because the Nginx Ingress Controller requires IP addresses as endpoints, not hosts. If the IP address is absent from the relation data, the `nginx-ingress-integrator` charm won't receive notifications of IP address changes. Therefore, to ensure compatibility with Nginx ingress in Kubernetes, it's essential to have the IP address present in the relation data.

As a solution, we could initially source the binding address from Juju. If that doesn't work We can then use `socket.gethostbyname(socket.gethostname())` if it's not in `127/8`. If all else fails, we can fallback to `socket.gethostbyname(socket.getfqdn())`. 

This approach ensures our library will never fail in environments in the environment mandated by Traefik-compatible applications (i.e. cluster DNS is enabled). And it will only fail when there isn't a cluster DNS, in that case, the Traefik charm can't function properly as well.

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: The problem of merging the `ip` and `host` into one field and let the requirer to choice to provide `ip` or `host` is that the requirer charm doesn't have the information whether it's connected with the `traefik-k8s` charm of `nginx-ingress-integrator` charm. And for `nginx-ingress-inegrator` the `ip` is a must even the cluster DNS is enabled due to the limitation of Nginx ingress in Kubernetes.

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: How does the `nginx-ingress-integrator` charm run the `juju status` command?

When IP addresses are changed, the `nginx-ingress-integrator` needs to respond by updating the Kubernetes Nginx Ingress controller with the new IP address. This integrator operates based on Juju events. How can the `nginx-ingress-integrator` detect changes in IP addresses when they aren't present in the relation?

weiiwang01: I found some context from the previous meeting about the decision: https://chat.charmhub.io/charmhub/pl/m6mb68ijxjrwpkthsgf6xxju7r

simskij: @PietroPasotti and @sed-i:

The issue here is not whether we should use ip or not in the traefik charm, cause we won't - we really don't have any reason to. The issue is rather that `ingress`, as a charm relation interface, needs to work with both nginx and traefik.

We could make ip optional. We could make host optional. The problem is that they'll be required in one interface implementation each. And if we make both required, the charms will crash when there is no bound ip yet, which is something traefik currently survives.

PietroPasotti: ah so it's really a matter of making ingress interoperable between traefik and nginx?

PietroPasotti: if so then I feel this is a discussion for canonical/charm-relation-interfaces

sed-i: > How does the `nginx-ingress-integrator` charm run the `juju status` command?
> 
> When IP addresses are changed, the `nginx-ingress-integrator` needs to respond by updating the Kubernetes Nginx Ingress controller with the new IP address. This integrator operates based on Juju events. How can the `nginx-ingress-integrator` detect changes in IP addresses when they aren't present in the relation?

It doesn't need to run `juju status` - that information is already in relation data:
```
$ juju show-unit am/0
am/0:
  opened-ports: []
  charm: ch:amd64/focal/alertmanager-k8s-88
  leader: true
  life: alive
  relation-info:
  - relation-id: 1
    endpoint: replicas
    related-endpoint: replicas
    application-data: {}
    local-unit:
      in-scope: true
      data:
        egress-subnets: 10.152.183.216/32
        ingress-address: 10.152.183.216
        private-address: 10.152.183.216
  - relation-id: 2
    endpoint: ingress
    related-endpoint: ingress
    application-data:
      ingress: '{"url": "http://10.43.8.206/test-am"}'
    related-units:
      trfk/0:
        in-scope: true
        data:
          egress-subnets: 10.43.8.206/32
          ingress-address: 10.43.8.206
          private-address: 10.43.8.206
  provider-id: am-0
  address: 10.1.166.95
```

Note how `private-address` is present for all relations' unit data.

weiiwang01: In Juju for Kubernetes, the `private-address` does not point to the unit's IP address; instead, it points to the cluster IP of service created by Juju for the entire application. To utilize that service, applications must use the `open-port`, which is only available in Juju 3.

gregory-schiano: Shouldn't we organize a call to discuss and take a decision on that topic ? @sed-i @simskij @PietroPasotti

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -651,40 +654,40 @@ def provide_ingress_requirements(
...
-                    f"can't find IP address for relation {self.relation}, bind address is None"
-                )
-            ip = str(binding_ip)
+                ip = None
+            else:
+                binding_ip = network_binding.network.bind_address
+                if binding_ip is None:
+                    ip = None
+                else:
+                    ip = str(binding_ip)
```
gregory-schiano: I think that this can be simplified to:
```python
if network_binding is None and (binding_ip := network_binding.network.bind_address) is None:
   return None
ip = str(binding_ip)
```

lib/charms/traefik_k8s/v2/ingress.py
gregory-schiano: @sed-i @simskij and @PietroPasotti Weii adapted based on our discussion yesterday, could you review it again ?

PietroPasotti (commented):
I'm missing the context from yesterday's meeting, so I hope I'm not reopening some already sealed pandora's boxes, but here are my two concerns

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: Gotcha, updated, thanks!

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -200,14 +193,32 @@ def validate_port(cls, port):  # noqa: N805  # pydantic wants 'cls' as first arg
...
         """Validate host."""
         assert isinstance(host, str), type(host)
         return host
 
+    @validator("ip", pre=True)
+    def validate_ip(cls, ip):  # noqa: N805  # pydantic wants 'cls' as first arg
+        """Validate ip."""
+        if ip is None:
+            return None
+        assert isinstance(ip, str), type(ip)
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -200,14 +193,32 @@ def validate_port(cls, port):  # noqa: N805  # pydantic wants 'cls' as first arg
 class IngressRequirerUnitData(DatabagModel):
     """Ingress requirer unit databag model."""
 
-    host: str = Field(description="Hostname the unit wishes to be exposed.")
+    host: str = Field(description="Hostname at which the unit is reachable.")
+    ip: Optional[str] = Field(description="IP at which the unit is reachable.")
```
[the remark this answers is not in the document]
weiiwang01: The meeting's conclusion is that the `ip` field can be set to `None` exclusively when IP information is unattainable from juju. This serves as a solution to the current issue we have with juju. It is the responsibility of the ingress provider to appropriately handle the `None` IP fields. The public API does not permit users to manually set the `ip` field to `None`.

I've updated the description of the `ip` field. Thanks!

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: Yes, while the `ip` field is now technically nullable, it should only be set to `None` in the event of an issue with juju. For users, the `ip` field remains mandatory.

PietroPasotti: mmmh ok, so should we do something in case it's None, like log a warning? To prevent people from passing None and think it's a regular supported use case. I'm not sure how to do it with pydantic, but we can also consider not typing it at all as `Optional[]` and only in a validator, accept a None passed as value and log that warning.

weiiwang01: For users, the methods to consider are `IngressPerAppRequirer.__init__` and `IngressPerAppRequirer.provide_ingress_requirements`. When a user passes `None` as the `ip` value, it defaults to the IP address from Juju. Therefore, users cannot intentionally set the `ip` to `None` using the public API. As for the data class, the `ip` field should be typed as nullable, as the provider need to act depend on whether the `ip` field is `None` or not, and the provider need to get that from the data class.
Logging is definitely a good idea, added, thanks!

gregory-schiano: I follow Weii in his conclusion. We didn't want to have IP as mandatory field because depending on the provider this information is unnecessary, so we didn't want a charm related to Traefik to Error because it can't get its IP field. Setting it to nullable delegates the decision of what is needed to the provider, if the provider needs the IP field and this field is None, it can act accordingly (error, blocked, message, ...)
As Weii mentioned the Library user can't set this field using the public interface, so the risks there are limited

Is it ok for you @PietroPasotti ?

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -633,34 +653,53 @@ def provide_ingress_requirements(
...
     def _publish_unit_data(
         self,
         host: Optional[str],
+        ip: Optional[str],
         relation: Relation,
     ):
         if not host:
             host = socket.getfqdn()
 
+        if not ip:
```
PietroPasotti, line 685: I think this whole `if not ip` block can be simplified:

```suggestion
if ip is None:
    network_binding = self.charm.model.get_binding(relation)
    
    if network_binding is not None and (bind_address := network_binding.network.bind_address) is not None:
        ip = str(bind_address)
    else:
        log.error("failed to retrieve ip information from juju")
```

tests/unit/test_lib_per_app_requires.py
```diff
@@ -77,27 +77,28 @@ def test_ingress_app_requirer_related(requirer: IngressPerAppRequirer, harness,
 @pytest.mark.parametrize(
     "auto_data, ok",
     (
-        ((True, 42), False),
-        ((10, False), False),
-        ((10, None), False),
-        (("foo", 12), True),
+        ((True, "example.com", 42), False),
+        ((10, "example.com", False), False),
+        ((10, "example.com", None), False),
```
PietroPasotti, line 82: we should probably add a case in which everything is OK except for the IP, such as

`("foo", "not.an.ip", 12)`

to make sure we 're not only testing the other failures

tests/unit/test_lib_per_app_requires.py
```diff
@@ -127,6 +128,7 @@ def test_ipa_events(self):
         # WHEN an ingress relation is formed
         # THEN the ready event is emitted
         with capture(self.harness.charm, IngressPerAppReadyEvent):
+            self.harness.add_network("10.0.0.10")
```
PietroPasotti, line 131: can we add a test where we call provide_ingress_requirements without an IP, and there is no network, and assert that there is an `ip=null` in the databag?

PietroPasotti (changes requested):

lib/charms/traefik_k8s/v2/ingress.py
weiiwang01: Updated, thanks!

tests/unit/test_lib_per_app_requires.py
weiiwang01: I added that case, thanks!

tests/unit/test_lib_per_app_requires.py
weiiwang01: Yes, I have added another test case to ensure the ip field in unit databag is `null`.

simskij (commented):
Once tests are passing and the discussions are closed out, I'm happy having this merged.

weiiwang01:
> Once tests are passing and the discussions are closed out, I'm happy having this merged.

@simskij all fixed, sorry there was a copy paste error in the unit test.

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: @sed-i you too?

PietroPasotti (approved):
Fine by me! Thanks for the patience :)

PR #233, opened 2023-08-31
[IPA] Add scheme to provide_ingress_requirements

sed-i:
## Issue
The IPA library does not have a `scheme` arg in its manual updater, `provide_ingress_requirements`.

## Solution
Repeat #232, this time for IPA.

Addresses https://github.com/canonical/karma-k8s-operator/issues/29.

## Testing Instructions
Relate `ca - trfk - (karma - ca) - catalogue`, and make sure the karma URL displayed in catalogue is the ingress URL and not the k8s fqdn.

In tandem with:
- https://github.com/canonical/karma-k8s-operator/pull/36

## Release Notes
[IPA] Add scheme to provide_ingress_requirements.

PR #234, opened 2023-09-01
Set scheme for traefik_route

sed-i:
## Issue
Grafana, which uses traefik_route, is currently unable to tell if the ingress endpoint is http or https.

## Solution
This PR adds a scheme arg, so that traefik could inform over relation data which scheme it is using.

In tandem with:
- https://github.com/canonical/traefik-route-k8s-operator/pull/41
- https://github.com/canonical/grafana-k8s-operator/pull/257

Drive-by fixes:
- [IPU] Abort provide_ingress_requirements if there is no relation. The same thing was already impl'd for IPA in #233.

## Release Notes
Set scheme for traefik_route.

### DIFF
--- a/lib/charms/traefik_route_k8s/v0/traefik_route.py
+++ b/lib/charms/traefik_route_k8s/v0/traefik_route.py
@@ -88,7 +88,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
 log = logging.getLogger(__name__)
 
@@ -141,7 +141,12 @@ class TraefikRouteProvider(Object):
     _stored = StoredState()
 
     def __init__(
-        self, charm: CharmBase, relation_name: str = "traefik-route", external_host: str = ""
+        self,
+        charm: CharmBase,
+        relation_name: str = "traefik-route",
+        external_host: str = "",
+        *,
+        scheme: str = "http",
     ):
         """Constructor for TraefikRouteProvider.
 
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -85,7 +85,7 @@ allowlist_externals = /usr/bin/env
 description = Scenario tests
 deps =
     pytest
-    ops-scenario>=4.0.4
+    ops-scenario<5
```
sed-i, line 88: @PietroPasotti I tried `>=5`, unpinned ops and ran `tox -re`, but there were `_fs` errors and others.

lib/charms/traefik_route_k8s/v0/traefik_route.py
```diff
@@ -189,30 +200,43 @@ def _update_stored_external_host(self) -> None:
...
+            return
+
+        for relation in self._charm.model.relations[self._relation_name]:
+            if not relation.app:
+                self._stored.external_host = ""
+                self._stored.scheme = ""
+                return
+            external_host = relation.data[relation.app].get("external_host", "")
+            self._stored.external_host = external_host or self._stored.external_host  # type: ignore
+            scheme = relation.data[relation.app].get("scheme", "")
```
Abuelodelanada, line 213: Shouldn't we be explicit about the scheme here?

```suggestion
            scheme = relation.data[relation.app].get("scheme", "http")
```

Abuelodelanada (approved):

lib/charms/traefik_route_k8s/v0/traefik_route.py
sed-i: I think you're right, but I was following the existing pattern for `external_host`.
There are more odd things here that could use a follow-up PR:
- `_stored` gets overwritten every loop iteration.
- Why the provider needs to extract this from relation data? Seems like this was cloned from the requirer. Not sure.

sed-i: https://github.com/canonical/traefik-route-k8s-operator/issues/42

PR #245, opened 2023-09-15
TLS fixes

sed-i:
## Issue
Curling prometheus via ingress url results in `Internal Server Error`.

## Solution
1. Drop incorrect rootCAs section for e2e TLS.
2. On upgrade, make sure to push certs received via the "recv-ca-certs" relation.

## Context
On upgrade, no relation events are emitted (unless you modify relation data as part of the upgrade sequence). So if, for example, you get certificates over relation data, they would all be lost after an upgrade. And you do not want persistent storage because then you would need to figure out which ones you need to delete after an upgrade.

So on upgrade you’d need to iterate over all relations and write those certs to disk. But not quite on upgrade-charm. We need the container to be there, so this must happen of pebble-ready.

## Testing Instructions
Follow testing instructions in https://github.com/canonical/cos-lite-bundle/pull/80 and curl prometheus.

Until we have `cert_transfer` impl'd in prom and grafana, copy traefik's ca cert manually:
```
juju run external-ca/0 get-ca-certificate --format json | jq -r '."external-ca/0".results."ca-certificate"' > external-ca.crt
openssl x509 -noout -text -in external-ca.crt

echo | openssl s_client -showcerts -servername 10.43.8.206 -connect 10.43.8.206:443 | openssl x509 -text -noout

juju scp --container prometheus external-ca.crt prometheus/0:/usr/local/share/ca-certificates
juju ssh --container prometheus prometheus/0 update-ca-certificates --fresh

juju scp --container grafana external-ca.crt grafana/0:/usr/local/share/ca-certificates
juju ssh --container grafana grafana/0 update-ca-certificates --fresh

juju ssh --container grafana grafana/0 curl https://10.43.8.206/trfk-prometheus-0/api/v1/targets | jq | grep -E "scrapeUrl|health"
```

## Release Notes
Fix cert handling.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -70,6 +70,7 @@ from ops.model import (
     WaitingStatus,
 )
 from ops.pebble import APIError, LayerDict, PathError
+from string import Template
 
 logger = logging.getLogger(__name__)
 
@@ -85,7 +86,7 @@ _SERVER_CERT_PATH = f"{_DYNAMIC_CONFIG_DIR}/server.cert"
 _SERVER_KEY_PATH = f"{_DYNAMIC_CONFIG_DIR}/server.key"
 _CA_CERTS_PATH = "/usr/local/share/ca-certificates"
 _CA_CERT_PATH = f"{_CA_CERTS_PATH}/traefik-ca.crt"
-
+_RECV_CA_TEMPLATE = Template(f"{_CA_CERTS_PATH}/receive-ca-cert-$rel_id-ca.crt")
 
 BIN_PATH = "/usr/bin/traefik"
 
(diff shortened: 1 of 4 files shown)


PietroPasotti:
datasources are still broken in grafana after refreshing traefik to use this source
image

sed-i:
> datasources are still broken in grafana after refreshing traefik to use this source image

I think that's because prometheus is announcing its _ingress_ address to grafana. The problem is that grafana cannot validate traefik's cert because it doesn't have the external-ca cert.
- I think if you relate traefik to the local ca it would work (but would be wrong).
- Until https://github.com/canonical/operator/issues/970 is resolved, we could use only fqdns for grafana source urls (this would break datasources over CMRs), or add a "receive-ca-cert" to grafana and relate it to the external-ca :vomiting_face:.

sed-i:
I tested locally with external_hostname for traefik and it works.
After some QA by the team I believe this PR could be merged even if bare IP does not work yet.

PR #250, opened 2023-09-21
feat: add transparent routing mode

DnPlas:
With this mode we are ensuring that no prefix is added when manipulating the external url and generating the configuration segment.
This feature is helpful when we need transparency for handling ingress.

This PR is for discussing the implementation and use case(s).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -14,7 +14,7 @@ options:
       The routing mode allows you to specify how Traefik going to generate
       routes on behalf of the requesters.
 
-      Valid values are "path" and "subdomain".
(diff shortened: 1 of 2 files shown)


PietroPasotti:
@sed-i @simskij wdyt?

DnPlas:
Based on an offline discussion, this proposal will change slightly, so instead of having a "transparent" mode, we will extend the capabilities of the existing "path" mode to accept custom paths from ingress requirers, that way we cover more use cases and avoid adding a third routing mode.
I'm closing this PR as I'll submit a different one for adding the above.

PR #254, opened 2023-09-27
feat: enable custom prefix for ingress requirers

DnPlas:
This feature allows ingress requirers to tell traefik which prefix to use when generating the routing configuration, in cases where the routing rule is required to be other than model-appname.

TODO:
- [ ] Provide test cases for unit and integration(?) tests


src/charm.py
```diff
@@ -912,7 +912,10 @@ def _push_configurations(self, relation: Relation, config: Union[Dict[str, Any],
             self._wipe_ingress_for_relation(relation)
 
     @staticmethod
-    def _get_prefix(data: Dict[str, Any]):
+    def _get_prefix(data: Dict[str, Any]) -> str:
+        """Return a custom prefix if set by the ingress requirer, model-appname otherwise."""
```
PietroPasotti, line 916:
```suggestion
        """Return a custom prefix if set by the ingress requirer, ``[model]-[app name]-[unit id]`` otherwise."""
```

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -657,9 +666,13 @@ def _publish_unit_data(
         if not host:
             host = socket.getfqdn()
 
-        unit_databag = relation.data[self.unit]
+        # TODO: what should be sent to the relation data bag if prefix is None?
```
PietroPasotti, line 669: I think we shouldn't send anything in that case. I think pydantic should be good with accepting a `prefix=None`, only thing we need to make sure is that it doesn't get dumped to databag as a json `null`. If it does, no biggie, it will appear on the other side as None, but I think pydantic models have a config option to omit null values, and that feels a little bit cleaner.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -633,6 +641,7 @@ def provide_ingress_requirements(
             host: Hostname to be used by the ingress provider to address the
              requirer unit; if unspecified, FQDN will be used instead
             port: the port of the service (required)
+            prefix: the prefix to be used for generating the routing configuration.
```
PietroPasotti, line 644:
```suggestion
            prefix: the prefix to be used for generating the ingress route for this application. If omitted, it will default to ``[model name]-[application name]-[unit id]``. This means that by default your application will be exposed at ``http(s)://[traefik url]/mymodel-myapp-0/``, but if you pass ``prefix='foo'``, you will access it at ``http(s)://[traefik url]/foo/``. Special case: if you pass ``prefix=''``, the app will be exposed at ``http(s)://[traefik url]/``.
```

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: Which makes me realize, we need to differentiate between `None`, which means 'use the default prefix', and `""`, which means 'no prefix at all'.

It feels a bit ugly to have None mean something and an empty string mean something else. Should we perhaps go for a more complex type for `prefix`?

```
prefix=None # --> http://traefik-url/model-app-0/
prefix=Prefix(s: str) # -->  http://traefik-url/{s} 
```

This is also more future proof, may we ever need to add further config to the prefix data.

PietroPasotti: @sed-i @simskij wdyt?

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -201,13 +201,19 @@ class IngressRequirerUnitData(DatabagModel):
...
 
     @validator("host", pre=True)
     def validate_host(cls, host):  # noqa: N805  # pydantic wants 'cls' as first arg
         """Validate host."""
         assert isinstance(host, str), type(host)
         return host
 
+    @validator("prefix", pre=True)
+    def validate_prefix(cls, prefix):  # noqa: N805  # pydantic wants 'cls' as first arg
+        """Validate host."""
```
PietroPasotti, line 214:
```suggestion
        """Validate prefix."""
```

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -201,13 +201,19 @@ class IngressRequirerUnitData(DatabagModel):
...
+        assert isinstance(prefix, str), type(prefix)
```
PietroPasotti, line 215: don't think we actually need this. pydantic will cast to string or None if we decide to make it Optional.
should we validate that, if string is nonempty, it has a trailing slash? or will traefik fill that in?

lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -69,7 +69,7 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
```
PietroPasotti, line 72: I would add this feature to v2 only, not v1. V1 is in maintenance mode: no backports unless it's required to fix some outstanding bugs.

PietroPasotti (commented):
going in the right direction! Some docstring suggestions and a few thoughts on how to tell a None from an empty string in the custom prefix specification

lib/charms/traefik_k8s/v2/ingress.py
PietroPasotti: For reference, had a discussion on this in Riga. Conclusions:

- the requirement for a 'transparent proxy' configuration of traefik is no longer applicable. 
- That configuration is possible: using "PathPrefix(`/`)" you can talk 'through' traefik and reach any endpoint behind it without prefix.
- However: doing so will effectively overwrite any of traefik's own endpoints such as /metrics and /debug. So we'll effectively lose self-monitoring.

Thus, for now we drop the 'empty' prefix configuration option and require the custom prefix to be nonempty.

The risk remains that if someone passes `prefix=metrics`, when one asks traefik for its metrics, they'll be in fact accessing (proxied application)'s metrics. 
We have to clearly document this risk and do what's possible to warn against it. Traefik charm will have to include a list of traefik's own internal endpoints such as `/api`, `/metrics`, `/debug` and so on, and warn if anyone passes a custom prefix conflicting with those.

Also, we can and should check the requested custom prefixes for duplicates: if two apps request to be exposed on `/foo`, only one of them will be. In this case the charm should probably block (@simskij ?).

Issue #265, opened 2023-10-06
_update_received_ca_certs method first needs to validate the availability of traefik container


gatici:
### Bug Description

During the integration of CertificateTransfer interface:
If CertificateTransferAvailableEvent is emitted too quick before traefik container become available, _update_received_ca_certs method tries to write to an non-existing path which causes "ops.pebble.ConnectionError: [Errno 2] No such file or directory" error.
In , def _update_received_ca_certs method, first we need to check if container is up or not.

### To Reproduce

juju deploy vault-k8s --channel edge -n 5 --trust
juju deploy traefik-k8s --channel edge --trust
juju integrate vault-k8s:send-ca-cert traefik-k8s:receive-ca-cert

### Environment

juju: 3.1.5
microk8s: v1.27.5
lxd:5.0.2-838e1b2
certificate_transfer interface: 0.4

### Relevant log output

```shell
unit-traefik-0: 2023-10-06 12:03:21 DEBUG jujuc running hook tool "state-get" for traefik/0-receive-ca-cert-relation-changed-577932895375089497
unit-traefik-0: 2023-10-06 12:03:21 DEBUG jujuc running hook tool "juju-log" for traefik/0-receive-ca-cert-relation-changed-577932895375089497
unit-traefik-0: 2023-10-06 12:03:21 DEBUG unit.traefik/0.juju-log receive-ca-cert:4: Emitting custom event <CertificateAvailableEvent via TraefikIngressCharm/CertificateTransferRequires[receive-ca-cert]/on/certificate_available[42]>.
unit-traefik-0: 2023-10-06 12:03:21 DEBUG jujuc running hook tool "juju-log" for traefik/0-receive-ca-cert-relation-changed-577932895375089497
unit-traefik-0: 2023-10-06 12:03:21 ERROR unit.traefik/0.juju-log receive-ca-cert:4: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/usr/lib/python3.8/urllib/request.py", line 1354, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1302, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1251, in endheaders
[...]

PR #266, opened 2023-10-09
Fix Interface and Scenario tests

gatici:
## Issue

Problems in Scenario and interface tests
- Using ops-scenario >= 5.3.1
- https://github.com/canonical/ops-scenario/pull/64 is merged so IndexError does not raise anymore. test_ipa.py needs to be updated.
- https://github.com/canonical/charm-relation-interfaces/pull/85 is merged. No need to use specific ingress-v2-json ingress branch.

## Solution
After merging following PRs, the problems appeared in the tests and the tests are fixed accordingly.
-  https://github.com/canonical/ops-scenario/pull/64 
-  https://github.com/canonical/charm-relation-interfaces/pull/85

## Testing Instructions
Run `tox` to validate that all tests are passing.

## Release Notes
This PR aims to fix the problems which are happened after merging some PRs in the dependent external libraries.

PR #267, opened 2023-10-09
Validate the availability of traefik container before updating CA cer…

gatici:
Validate the availability of traefik container before updating CA certs

## Issue
While updating CA certificates when CertificateTransfer relation joined, traefik container many not be running yet. This causes ConnectionError like https://github.com/canonical/vault-k8s-operator/actions/runs/6430745831/job/17462295379?pr=53. 

## Solution
Availability of traefik container needs to be checked before trying to update CA certs

## Context
The issue is appeared when Traefik is related with Vault-k8s using CertificateTransfer interface.

## Testing Instructions
Run `tox -e integration`.

## Release Notes
This PR aims to fix the bug: https://github.com/canonical/traefik-k8s-operator/issues/265. 
First https://github.com/canonical/traefik-k8s-operator/pull/266 needs to be merged as it fixes some existing test problems  in main branch. Otherwise, some tests are failing regardless of this PR.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -268,6 +268,8 @@ class TraefikIngressCharm(CharmBase):
 
     def _on_recv_ca_cert_available(self, event: CertificateTransferAvailableEvent):
         # Assuming only one cert per relation (this is in line with the original lib design).
+        if not self.container.can_connect():
+            return
         self._update_received_ca_certs(event)
 
     def _update_received_ca_certs(self, event: Optional[CertificateTransferAvailableEvent] = None):
@@ -278,9 +280,6 @@ class TraefikIngressCharm(CharmBase):
         Calling this function from upgrade-charm might be too early though. Pebble-ready is
         preferred.
         """
-        if not self.container.can_connect():
-            return
-
         if event:
             self.container.push(
                 _RECV_CA_TEMPLATE.substitute(rel_id=event.relation_id), event.ca, make_dirs=True
--- a/src/charm.py
+++ b/src/charm.py
@@ -278,6 +278,9 @@ class TraefikIngressCharm(CharmBase):
         Calling this function from upgrade-charm might be too early though. Pebble-ready is
         preferred.
         """
+        if not self.container.can_connect():
+            return
+
         if event:
             self.container.push(
                 _RECV_CA_TEMPLATE.substitute(rel_id=event.relation_id), event.ca, make_dirs=True
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -423,9 +423,9 @@ class TestTraefikCertTransferInterface(unittest.TestCase):
     @patch("ops.model.Container.exec")
     @patch("charm._get_loadbalancer_status", lambda **__: "10.0.0.1")
     @patch("charm.KubernetesServicePatch", lambda *_, **__: None)
-    def test_given_container_can_connect_when_receive_ca_cert_relation_joins_then_ca_certs_are_updated(
-        self, patch_exec
-    ):
+    def test_transferred_ca_certs_are_updated(self, patch_exec):
+        # Given container is ready, when receive-ca-cert relation joins,
+        # then ca certs are updated.
         provider_app = "self-signed-certificates"
         self.harness.set_leader(True)
         self.harness.begin_with_initial_hooks()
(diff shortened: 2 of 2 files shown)


src/charm.py
```diff
@@ -278,6 +278,10 @@ def _update_received_ca_certs(self, event: Optional[CertificateTransferAvailable
         Calling this function from upgrade-charm might be too early though. Pebble-ready is
         preferred.
         """
+        if not self.container.can_connect():
+            event.defer()
```
PietroPasotti, line 282: As a general rule of thumb, we try to avoid using defer whenever we can.
https://juju.is/docs/sdk/deferring-events-details-and-dilemmas

I believe this method is also called from pebble-ready, so if the container is not ready when it is first called, it will be called once pebble-ready fires. So we should be good to go without defer?

PietroPasotti (commented):
Could you verify that we can indeed remove the defer() call and still have the charm work as intended?

src/charm.py
gatici: You are right. PR is updated.

PietroPasotti (approved):
very nice, thanks!

src/charm.py
```diff
@@ -278,6 +278,9 @@ def _update_received_ca_certs(self, event: Optional[CertificateTransferAvailable
         Calling this function from upgrade-charm might be too early though. Pebble-ready is
         preferred.
         """
+        if not self.container.can_connect():
+            return
```
sed-i, line 282: The only caller is `_on_recv_ca_cert_available`.
I think the `can_connect` guard needs to be moved there because it's ok if `_on_recv_ca_cert_available` returns without side-effects, but it's not ok for this method to return without side-effects.

tests/unit/test_charm.py
```diff
@@ -410,6 +410,52 @@ def test_tcp_config(self):
...
+        self.mock_version = patcher.start()
+        self.addCleanup(patcher.stop)
+        self.container_name = "traefik"
+
+    @patch("ops.model.Container.exec")
+    @patch("charm._get_loadbalancer_status", lambda **__: "10.0.0.1")
+    @patch("charm.KubernetesServicePatch", lambda *_, **__: None)
+    def test_given_container_can_connect_when_receive_ca_cert_relation_joins_then_ca_certs_are_updated(
+        self, patch_exec
+    ):
```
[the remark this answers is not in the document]
gatici: Here I followed the structure that  we are using to name a test case. 
- Given conditions (given)
- When something happened ? (when)
- What's the result ? (then)
TBH,  I found that structure more readable.  But if you do not want to use it, I can apply your suggestion.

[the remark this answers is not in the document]
sed-i: Thanks @gatici.
Yes, could you please shorten the function names? That is the main thing.

src/charm.py
gatici: Actually, it's also called by def _on_traefik_pebble_ready, but we do not need a guard there. We can move the guard to _on_recv_ca_cert_available.

tests/unit/test_charm.py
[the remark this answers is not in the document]
gatici: Done

src/charm.py
gatici: Done

tests/unit/test_charm.py
```diff
@@ -410,6 +410,52 @@ def test_tcp_config(self):
...
+            relation_id=certificate_transfer_rel_id, remote_unit_name=f"{provider_app}/0"
+        )
+        patch_exec.assert_called_once_with(["update-ca-certificates", "--fresh"])
+
+    @patch("ops.model.Container.exec")
+    @patch("charm._get_loadbalancer_status", lambda **__: "10.0.0.1")
+    @patch("charm.KubernetesServicePatch", lambda *_, **__: None)
+    def test_given_container_not_ready_when_receive_ca_cert_relation_joins_then_ca_certs_are_not_updated(
+        self, patch_exec
+    ):
```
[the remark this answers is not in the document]
gatici: Done

sed-i (approved):

sed-i:
Thanks @gatici for this important fix!

PietroPasotti:
waiting on one more thumb up before we go: 
@sed-i perhaps?
or @dstathis ?

dstathis (commented):
+1

PR #269, opened 2023-10-10
externalize traefik container and config management

PietroPasotti:
## Issue
The traefik's charm code is somewhat outdated and needed some freshening up. 
Adopting some more modern charming patterns is the start of that process.

## Solution

In this PR we add a Traefik class that takes care of all pebble interactions and all traefik-workload-specific configuration tasks. 
This way the charm can be unaware of:
- the yaml config file structure traefik wants
- the locations at which the traefik files need to be
- how to manage the traefik process (start it, restart it...)

## Context

A next PR will add a manifest-driven filesystem manager, to prevent unnecessary pushes/restarts unless the backing juju state 'really' has changed.

## Testing Instructions

We're going to need careful testing especially in combination with the TLS bundle to catch regressions.


src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+
+
+class RoutingMode(enum.Enum):
+    """Routing mode."""
+
+    path = "path"
+    subdomain = "subdomain"
+
+
+class TraefikError(RuntimeError):
```
sed-i, line 42: Afaik, best practice for custom exceptions is to inherit Exception, not RuntimeError.
```suggestion
class TraefikError(Exception):
```

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+    """Raised if the caller incorrectly assumes that the traefik container is ready."""
+
+
+class Traefik:
+    """Traefik workload representation."""
+
+    diagnostics_port = 8082  # Prometheus metrics, healthcheck/ping
+    _port = 80
+    _tls_port = 443
+    log_path = "/var/log/traefik.log"
```
sed-i, line 56: Now that we have a dedicated module, these vars that are not pertaining to the class itself could become module scope vars (with `_` prefix).
`_layer_name` I'd still keep here.

Not critical.

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+    ):
+        self._service_name = service_name
+        self._container = container
+        self._tcp_entrypoints = tcp_entrypoints
+        self._routing_mode = routing_mode
+        self._tls_enabled = tls_enabled
+
+    @property
+    def _hostname(self) -> str:
+        return socket.getfqdn()
```
sed-i, line 75: I'm on the fence with having this property.
- In ingress requirers it makes more sense to have `_internal_url` but for traefik we don't need it probable.
- Calling this `hostname` is confusing because it's actually "internal hostname"
- Testing is a bit more fragile because what do we mock? _hostname or getfqdn?
- `socket.getfqdn()` is just as easy to type as `self._hostname`

I guess I'm in favor of using getfqdn directly.

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+
+    @property
+    def service_ports(self) -> List[ServicePort]:
+        """Kubernetes service ports to be opened for this workload."""
```
sed-i, line 79:
```suggestion
        """Kubernetes service ports to be opened for this workload.
        
        We cannot use ops unit.open_port here because it gives a ClusterIP
        but for traefik we need LoadBalancer type.
        """
```

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+            [ServicePort(int(port), name=name) for name, port in self._tcp_entrypoints.items()]
+        ]
+
+    @property
+    def scrape_jobs(self) -> list:
+        """List of static configs for prometheus scrape."""
+        return [
+            {
+                "static_configs": [{"targets": [f"{self._hostname}:{self.diagnostics_port}"]}],
+            }
```
sed-i, line 92: Scheme is missing.
But we need to think about the topology we target here:
- for in cluster scrape (by the same prom that is proxied by trfk?), we would need to use `insecureSkipVerify`, as the cert that would be served is for the external hostname.
- for out-of-cluster scrape, we would need to give the external hostname and need prom/gagent to have a `receive-ca-cert` relation (relate to the external ca)

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+                            "defaultCertificate": {
+                                "certFile": SERVER_CERT_PATH,
+                                "keyFile": SERVER_KEY_PATH,
+                            },
+                        },
+                    },
+                }
+            }
+        )
+        self._container.push(DYNAMIC_CERTS_PATH, config, make_dirs=True)
```
sed-i, line 119: Given the push here, perhaps we should name the method more clearly?
"update tls config" or "push tls config"?

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+            self._container.remove_path(SERVER_KEY_PATH, recursive=True)
+
+        if ca:
+            self._container.push(CA_CERT_PATH, ca, make_dirs=True)
+        else:
+            self._container.remove_path(CA_CERT_PATH, recursive=True)
+
+        self.update_ca_certs()
+
+    def add_ca(self, ca: str, uid: str | int):
```
sed-i, line 149: `|` is not supported by py3.8.

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+        prefix: str,
+        scheme: str = "http",
+        host: str,
+        port: int,
+        strip_prefix: bool,
+        redirect_https: bool,
+        external_host: str,
+    ) -> dict:
+        """Generate a config dict for Ingress v1 (PerLeader)."""
+        lb_servers = [{"url": f"http://{host}:{port}"}]
```
sed-i, line 289: Let's avoid default kwargs for mutables.

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+        return self._generate_config_block(
+            prefix=prefix,
+            lb_servers=lb_servers,
+            scheme=scheme,
+            strip_prefix=strip_prefix,
+            redirect_https=redirect_https,
+            external_host=external_host,
+        )
+
+    def _generate_config_block(
```
sed-i, line 299: This and other methods do config + workload concerns.
Totally fine. Just pointing out if you intended otherwise.

src/traefik.py
```diff
@@ -0,0 +1,574 @@
...
+    def add_dynamic_config(self, file_name: str, config: str):
+        """Push a yaml to the dynamic config dir."""
+        self._container.push(Path(DYNAMIC_CONFIG_DIR) / file_name, config, make_dirs=True)
+
+        logger.debug("Updated dynamic configuration file: %s", file_name)
+
+    def clear_certs(self):
+        """Delete the server cert and server key."""
+        self._container.remove_path(SERVER_CERT_PATH)
+        self._container.remove_path(SERVER_KEY_PATH)
```
sed-i, line 525: We should be mindful here of potential race conditions.
Traefik watches these and reloads automatically.
We should make sure the caller first updates traefik config, and only then deletes the certs.
Should document this as well.

tests/scenario/conftest.py
```diff
@@ -54,6 +54,7 @@ def traefik_container(tmp_path):
         layers={"traefik": layer},
         exec_mock={
             ("update-ca-certificates", "--fresh"): ExecOutput(),
+            ("find", "/opt/traefik/juju", "-name", "*.yaml", "-delete"): ExecOutput(),
```
sed-i, line 57: Seems like we need a pebble feature request.

src/traefik.py
PietroPasotti: uhm, interesting. I always used runtimeerror as base class for module-specific error hierarchies.
But I looked into it and I couldn't find a reason why I shouldn't be using Exception instead. So yeah, thanks  :)

src/traefik.py
PietroPasotti: I did not change this code: it's the same as it is right now (https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L1315C1-L1316C1). Are you saying there's an error there?

src/traefik.py
PietroPasotti: mmmh just to make sure we're on the same page, my idea was that the Traefik class should handle the workload configuration (push/pull/remove config yamls), and (mediate, or initiate) workload operations such as starting/restarting.

Are you saying something else?

src/traefik.py
PietroPasotti: I was planning to address more of this in a next PR adding filesystem state manifest-based "as atomic as possible" transactions.
I see the danger but I don't think that's in scope for this PR, unless you think I introduced a bug with my changes? I tried to keep the logic as unchanged as possible, this pr should, if I did things right, be a refactoring only.

PietroPasotti: that said I'm fine adding a todo or comments to point this out to our tomorrow selves

PietroPasotti: beside, I just realized that in `main`, the charm method calling what is now this code [is unreachable](https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L322)

so the `clear_certs` code is effectively unused as it currently stands, and I'll remove it from here (and the two unreachable handlers in charm.py).

The concern you raised still applies, but to the `remove_ca` method. I'll add a note there.

tests/scenario/conftest.py
PietroPasotti: not sure this is generic enough to be useful to many, but we can try :)

PietroPasotti: https://github.com/canonical/operator/issues/1042

src/traefik.py
PietroPasotti: given ^, I'll close this for now but @sed-i if you think there's an issue there we should track this in a separate place

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+        """Update ca certificates and restart traefik."""
+        self._container.exec(["update-ca-certificates", "--fresh"]).wait()
+
+        # Must restart traefik after refreshing certs, otherwise:
+        # - newly added certs will not be loaded and traefik will keep erroring-out with "signed by
+        #   unknown authority".
+        # - old certs will be kept active.
+        self.restart()
+
+    def _update_static_configuration(self, tcp_entrypoints: Dict[str, int]):
```
sed-i, line 179: Isn't `tcp_entrypoints` a constructor arg?

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+                        **({"ca": CA_CERT_PATH} if self._tls_enabled else {"insecure": True}),
+                    }
+                }
+            }
+        )
+        logger.debug(f"dumping tracing config to {DYNAMIC_TRACING_PATH}")
+
+        self._container.push(DYNAMIC_TRACING_PATH, config, make_dirs=True)
+
+    def get_per_unit_http_config(
```
sed-i, line 241: Should this and related methods be moved to a separate class?

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+
+        if not self.is_running:
+            self._container.add_layer(self._layer_name, cast(LayerDict, layer), combine=True)
+            logger.debug(f"replanning {self._service_name!r} after a service update")
+            self._container.replan()
+        else:
+            logger.debug(f"restarting {self._service_name!r}")
+            self._container.restart(self._service_name)
+
+    def clear_dynamic_configs(self):
```
sed-i, line 524: We use "delete" below, so would be clearer without introducing a new term, "clear" :)
```suggestion
    def delete_all_dynamic_configs(self):
```

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+        """Delete **ALL** yamls from the dynamic config dir."""
+        # instead of multiple calls to self._container.remove_path(), delete all files in a swoop
+        self._container.exec(["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"])
+        logger.debug("Deleted all dynamic configuration files.")
+
+    def delete_dynamic_config(self, file_name: str):
+        """Delete a specific yaml from the dynamic config dir."""
+        self._container.remove_path(Path(DYNAMIC_CONFIG_DIR) / file_name)
+        logger.debug("deleted dynamic configuration file: %s", file_name)
```
sed-i, line 533:
```suggestion
        logger.debug("Deleted dynamic configuration file: %s", file_name)
```

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+        """
+        # make_dirs is technically not necessary at runtime, since traefik.configure() should
+        # guarantee that the dynamic config dir exists. However, it simplifies testing as it means
+        # we don't have to worry about setting up manually the traefik container every time we
+        # simulate an event.
+        self._container.push(Path(DYNAMIC_CONFIG_DIR) / file_name, config, make_dirs=True)
+
+        logger.debug("Updated dynamic configuration file: %s", file_name)
+
+    def clear_tracing_config(self):
```
sed-i, line 548:
```suggestion
    def delete_tracing_config(self):
```

src/charm.py
```diff
@@ -174,17 +138,23 @@ def __init__(self, *args):
...
         )
 
-        web = ServicePort(self._port, name=f"{self.app.name}")
-        websecure = ServicePort(self._tls_port, name=f"{self.app.name}-tls")
-        tcp_ports = [ServicePort(int(port), name=name) for name, port in self._tcp_ports.items()]
         self.service_patch = KubernetesServicePatch(
             charm=self,
             service_type="LoadBalancer",
-            ports=[web, websecure] + tcp_ports,
+            ports=self.traefik.service_ports,
```
sed-i, line 157: Per other comment, I think the k8s datastructure should be constructed here in charm.py, not there.
Could still use e.g. `self.traefik.port_x`.

src/charm.py
```diff
@@ -343,7 +294,7 @@ def _on_tracing_endpoint_removed(self, event) -> None:
             # clear anything up. We could defer, but again, we're being torn down and the unit db
             # will
             return
-        self._clear_tracing_config()
+        self.traefik.clear_tracing_config()
```
sed-i, line 297: Not sure if we should keep doing delta-charming here.

src/traefik.py
PietroPasotti: Ok, will try this out in a follow-up PR

src/charm.py
```diff
@@ -174,17 +138,23 @@ def __init__(self, *args):
...
-            charm=self, external_host=self.external_host, scheme=self._scheme  # type: ignore
+            charm=self, external_host=self._external_host, scheme=self._scheme  # type: ignore
+        )
+
+        self.traefik = Traefik(
+            _TRAEFIK_SERVICE_NAME,
+            self.container,
+            routing_mode=self._routing_mode,
+            tcp_entrypoints=self._tcp_entrypoints(),
+            tls_enabled=self._is_tls_enabled(),
```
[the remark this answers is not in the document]
PietroPasotti: `_on_cert_changed` calls `_configure_traefik` calls `Traefik.configure` calls `Traefik._update_tls_configuration()`.

So that path should be covered. Assuming that if cert-changed is firing, `Traefik._tls_enabled` (aka `cert_handler.enabled`) is True, traefik will dump the tls certs.

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+    def remove_ca(self, uid: Union[str, int]):
+        """Remove a ca.
+
+        BEWARE of potential race conditions.
+        Traefik watches the dynamic config dir and reloads automatically on change.
+        So make sure any traefik config depending on the certificates being there is updated
+        **before** the certs are removed, otherwise you might have some downtime.
+        """
+        ca_path = RECV_CA_TEMPLATE.substitute(rel_id=str(uid))
+        self._container.remove_path(ca_path)
```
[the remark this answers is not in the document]
PietroPasotti: same as above: 
image

Although we're not really doing any 'nested loop' ca removals, so we could simplify this flow by making `remove_ca` always automatically call update-ca-certificates, but I think this way it's more symmetric.

Another option is to refactor `add_ca` and `remove_ca` to:
`add_cas(List[CA])`, and make this method call always invoke automatically update-ca-certificates after adding all the CAs. Analogously for remove_ca --> remove_cas.

this would mean the charm is no longer responsible for calling update_ca_certificates after it's done adding/removing the CAs.

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+        container: Container,
+        routing_mode: RoutingMode,
+        tls_enabled: bool,
+        tcp_entrypoints: Dict[str, int],
+    ):
+        self._service_name = service_name
+        self._container = container
+        self._tcp_entrypoints = tcp_entrypoints
+        self._routing_mode = routing_mode
+        self._tls_enabled = tls_enabled
```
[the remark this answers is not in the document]
PietroPasotti: event observers should be defined on the charm only, as per https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619/2

src/traefik.py
PietroPasotti: ha, you're right. forgot to lift that

src/charm.py
PietroPasotti: mm, that does feel like a charm concern, indeed.

src/traefik.py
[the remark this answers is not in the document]
PietroPasotti: in light of more comments below, I went for the refactoring mentioned ^

src/charm.py
PietroPasotti: this PR is not about making traefik fully holistic. 
Happy to consider in a follow-up

src/charm.py
```diff
@@ -352,46 +303,29 @@ def _on_tracing_endpoint_changed(self, event) -> None:
...
     def _update_system_certs(self):
-        self.container.exec(["update-ca-certificates", "--fresh"]).wait()
-
-        # Must restart traefik after refreshing certs, otherwise:
-        # - newly added certs will not be loaded and traefik will keep erroring-out with "signed by
-        #   unknown authority".
-        # - old certs will be kept active.
-        self._restart_traefik()
+        with self.status(MaintenanceStatus("updating ca certificates")):
+            self.traefik.update_ca_certs()
```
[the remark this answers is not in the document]
PietroPasotti: yup, done.

src/charm.py
```diff
@@ -424,140 +358,33 @@ def _tcp_entrypoints(self):
...
-            }
-        }
-
-    def _get_tracing_config(self) -> dict:
-        """Return dictionary with opentelemetry configuration if available."""
+    def _configure_traefik(self):
+        self.traefik.configure()
+        self._configure_tracing()
+
+    def _configure_tracing(self):
```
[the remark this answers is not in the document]
PietroPasotti: mmm, are you sure? I thought pyright was smart enough to derive implicit None returns.

src/charm.py
```diff
@@ -1177,39 +835,9 @@ def _relation_config_file(relation: Relation):
...
-            self.unit.status = MaintenanceStatus(f"creating the {_TRAEFIK_SERVICE_NAME!r} service")
-            self.container.add_layer(
-                _TRAEFIK_LAYER_NAME, typing.cast(LayerDict, layer), combine=True
-            )
-            logger.debug(f"replanning {_TRAEFIK_SERVICE_NAME!r} after a service update")
-            self.container.replan()
-        else:
-            logger.debug(f"restarting {_TRAEFIK_SERVICE_NAME!r}")
-            self.container.restart(_TRAEFIK_SERVICE_NAME)
+        with self.status(MaintenanceStatus("restarting traefik...")):
```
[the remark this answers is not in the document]
PietroPasotti: I think it's useful information, as above

src/charm.py
```diff
@@ -1237,45 +865,48 @@ def _provider_from_relation(self, relation: Relation):
...
+        returns None. Only use this directly when external_host is allowed to be None.
         """
         if external_hostname := self.model.config.get("external_hostname"):
             return external_hostname
 
         return _get_loadbalancer_status(namespace=self.model.name, service_name=self.app.name)
 
     @property
-    def _routing_mode(self) -> _RoutingMode:
+    def external_host(self) -> str:
```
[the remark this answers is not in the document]
PietroPasotti: we need external_host in two kinds of situations: in one of them it's allowed to be None, in the other it's not and we raise an exception. This division makes it explicit.
When you allow a None and don't care: use the private _external_host, when you need it to be a string and want to raise otherwise, use the public variant.

src/charm.py
```diff
@@ -835,13 +670,25 @@ def _get_configs_per_leader(self, relation: Relation) -> Dict[str, Any]:
             return {}
 
         prefix = self._get_prefix(data)  # type: ignore
-        config = self._generate_per_leader_config(prefix, data)  # type: ignore
+        config = self.traefik.get_per_leader_http_config(
```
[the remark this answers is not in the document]
PietroPasotti: Will externalize config in a follow-up pr

src/traefik.py
PietroPasotti: Will externalize config in a follow-up pr

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+
+    @property
+    def service_ports(self) -> List[ServicePort]:
```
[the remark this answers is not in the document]
PietroPasotti: moved to charm

src/traefik.py
```diff
@@ -0,0 +1,600 @@
...
+            self._container.push(CA_CERT_PATH, ca, make_dirs=True)
+        else:
+            self._container.remove_path(CA_CERT_PATH, recursive=True)
+
+        self.update_ca_certs()
+
+    def add_ca(self, ca: str, uid: Union[str, int]):
+        """Add a ca."""
+        ca_path = RECV_CA_TEMPLATE.substitute(rel_id=str(uid))
+        self._container.push(ca_path, ca, make_dirs=True)
```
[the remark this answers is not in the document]
PietroPasotti: fixed (see below)

src/charm.py
```diff
@@ -908,9 +775,7 @@ def _get_configs_per_unit(self, relation: Relation) -> Dict[str, Any]:
     def _push_configurations(self, relation: Relation, config: Union[Dict[str, Any], str]):
```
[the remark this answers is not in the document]
PietroPasotti: Will externalize config in a follow-up pr

src/traefik.py
PietroPasotti: let's track in a separate issue #273

sed-i:
Two yamls for the same app, one with http (relation was previously removed), one with https:

```
juju run external-ca/0 get-ca-certificate --format json | jq -r '."external-ca/0".results."ca-certificate"' > external.cert

$ curl --cacert external.cert https://10.43.8.206/pt-prom-0/-/ready
Client sent an HTTP request to an HTTPS server.

$ jsshc traefik trfk/0 ls /opt/traefik/juju
certificates.yaml			  server.cert
juju_ingress_ingress-per-unit_2_prom.yaml  server.key
juju_ingress_ingress-per-unit_3_prom.yaml  # <--- TWO YAMLS FOR THE SAME APP
```


Bundle:
```
bundle: kubernetes
applications:
  ca:
    charm: self-signed-certificates
    channel: edge
    revision: 40
    scale: 1
    constraints: arch=amd64
  external-ca:
    charm: self-signed-certificates
    channel: edge
    revision: 40
    scale: 1
    constraints: arch=amd64
  prom:
    charm: prometheus-k8s
    channel: edge
    revision: 152
    series: focal
    resources:
      prometheus-image: 130
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  trfk:
    charm: local:traefik-k8s-0
    series: focal
    scale: 1
    constraints: arch=amd64
    storage:
    trust: true
relations:
- - prom:ingress
  - trfk:ingress-per-unit
- - trfk:certificates
  - external-ca:certificates
- - prom:certificates
  - ca:certificates
```

PietroPasotti:
The bug you found is real, but I don't think that's a regression. I've seen that in the past. I think it's because the ipa requirer misses an observer for relation-departed. Adding that to the mix, and let's see if we can make this pass

foo.bndl
sed-i, line 1: Just remembered we have something like this in https://github.com/canonical/traefik-k8s-operator/tree/main/tests/manual. Remove?

sed-i:
> I think it's because the ipa requirer misses an observer for relation-departed.

I think that in this context, a bug in the requirer side of things shouldn't matter to the provider side: the provider gets a relation event from juju that should have resulted in deleting the stale file. I'm guessing it's [888](https://github.com/canonical/operator/issues/888)'s fault.

src/charm.py
```diff
@@ -649,12 +468,14 @@ def _process_status_and_configurations(self):
...
             return
 
-        if not self._traefik_service_running:
+        if not self.traefik.is_running:
             self.unit.status = WaitingStatus(f"waiting for service: '{_TRAEFIK_SERVICE_NAME}'")
             return
 
-        self.unit.status = MaintenanceStatus("updating ingress configurations")
+        with self.status(MaintenanceStatus("updating ingress configurations")):
+            self._update_ingress_configurations()
```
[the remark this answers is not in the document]
sed-i: I get that, but it seems under specced:
- We almost never do this in other charms
- We haven't established what qualifies for a "bleep" status change.
  - pebble operations
  - charm container disk operations
  - client calls to workload HTTP API
  - k8s interactions
  - etc.
- Could this have an undesirable impact on juju load?

sed-i (approved):

foo.bndl
PietroPasotti: oh, committed that by mistake.

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: rolled it back, will do some more specwork before we commit to this solution

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: I checked, it's able to derive None

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: It does, and we do: we have `external_host` that is guaranteed to not raise, but returns Optional[str]. That allows you to code defensively if that's your jam.
All I'm proposing is we **add** api for the other case, the one in which we should raise if external host is not set (because coding defensively is only nice at the toplevel logic layer, you don't want to keep questioning yourself when you're ten levels deep).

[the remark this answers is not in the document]
PietroPasotti: @sed-i is this a blocker we need to discuss more in depth about, or shall I consider you convinced and on-board?

src/traefik.py
```diff
@@ -0,0 +1,616 @@
...
+        We need to generate a different section per middleware type, otherwise traefik complains:
+          "cannot create middleware: multi-types middleware not supported, consider declaring two
+          different pieces of middleware instead"
+        """
+        no_prefix_middleware = {}  # type: Dict[str, Dict[str, Any]]
+        if self._routing_mode is RoutingMode.path:
+            if strip_prefix:
+                no_prefix_middleware[f"juju-sidecar-noprefix-{prefix}"] = {
+                    "stripPrefix": {"prefixes": [f"/{prefix}"], "forceSlash": False}
+                }
```
Abuelodelanada, line 454: I would remove one indentation level:

```suggestion
        if self._routing_mode is RoutingMode.path and strip_prefix:
            no_prefix_middleware[f"juju-sidecar-noprefix-{prefix}"] = {
                "stripPrefix": {"prefixes": [f"/{prefix}"], "forceSlash": False}
            }
```

src/traefik.py
```diff
@@ -0,0 +1,616 @@
...
+                    {
+                        "main": external_host,
+                        "sans": [f"*.{external_host}"],
+                    },
+                ],
+            }
+
+        else:
+            # When the external host is a bare IP, we do not need the 'domains' entry.
+            tls_entry = {}
```
Abuelodelanada, line 486: The `else` clause can be omitted.

```suggestion
        tls_entry = {}

        if is_hostname(external_host):
            # When the external host is not a bare IP, we need the 'domains' entry.
            tls_entry = {
                "domains": [
                    {
                        "main": external_host,
                        "sans": [f"*.{external_host}"],
                    },
                ],
            }
```

Abuelodelanada (approved):
Looks good to me!
I've only added to minor suggestions.

src/traefik.py
PietroPasotti: good thinking

src/traefik.py
PietroPasotti: in this case, I think "explicit is better than implicit" takes precedence. tls is complicated already :)

PR #274, opened 2023-10-19
Subdomain certs

### LINKED ISSUE #244: Subdomain routing mode does not work properly with TLS certificates
### Bug Description

When using Traefik in the `subdomain` routing mode, all Ingress will use Traefik's default certificate.

`CertHandler` manages a single certificate for Traefik, and uses the `external_hostname` as the subject and `sans`. New Ingresses generated hostnames will not be represented in the certificate, and the SNI matching will not work, thus serving the default certificate.

I see 3 ways of fixing this issue:

1. Modify `CertHandler` to support multiple certificates, and when set to the `subdomain` routing mode, request new certificates for each Ingress.
2. Drop `CertHandler` and use the tls library directly to request new certificates for each Ingress when set to the `subdomain` routing mode.
3. Keep `CertHandler` as is and add `f"*.{external_hostname}"` to the list of sans to request a wildcard certificate. This would have negative security implications however and I do not think it should be implemented; I am only mentioning it for completeness' sake.

### To Reproduce

1. juju add-model potato
2. `juju deploy traefik-k8s --trust --channel edge --config routing_mode=subdomain --config external_hostname=potato.com`
3. `juju deploy zinc-k8s --trust --channel edge`
4. `juju deploy self-signed-certificates --trust --channel edge`
5. `juju relate self-signed-certificates:certificates traefik-k8s:certificate`
6. `juju relate zinc-k8s:ingress traefik-k8s:ingress`
7. https://potato-zinc-k8s.potato.com  # this will use Traefik's default certificate

### Environment

This was tested with every charm from the edge channel, running locally on MicroK8s v1.27.5 revision 5892 with Juju 3.1.5-genericlinux-amd64.

### Relevant log output

```shell
There are not relevant logs.
```

### Additional context

_No response_

sed-i:
## Issue
The TLS cert solution was designed with only path routing in mind.
With subdomain routing, traefik is unable to match a domain to a cert.

## Solution
- [ ] Keep using cert_handler as-is for the bare hostname cert. Should keep it even for subdomain routing, to be able to reach traefik api endpoints and metrics endpoint.
- [ ] Refactor to have a cert per app (IPA) or per unit (IPU).
  - [ ] when ingress relation (unit) joins/departs, send CSR/CRR on behalf of the unit
  - [ ] when routing mode changes from subdomain to path, CRR everything
  - [ ] when routing mode changes from path to subdomain, CSR everything

Fixes #244.

### DIFF
--- a/doc/adr/2023-10-18 subdomain routing with cert per unit.md
+++ b/doc/adr/2023-10-18 subdomain routing with cert per unit.md
@@ -29,6 +29,8 @@ the use-case of strict subdomain routing (no wildcards).
 * Bad, because may get rate limited by the CA.
   * IPA: new CSR (CRR) for every added (removed) app
   * IPU: new CSR (CRR) for every added (removed) unit
+* Bad, because if we are not careful, the rate limit could break the only cert
+  in use.
 
 ## Pros and Cons of the Options
 
@@ -67,7 +69,9 @@ SAN example: app1.example.com, app2.example.com
 
 * Good, because no wildcard certs (more secure).
 * Bad, because the number of SANs per cert is limited.
-
+* Bad, because may get rate limited by the CA.
+* Bad, because depends on the order things are done on an ACME server: if the
+  server first revokes the previous certificate, it could create outages.
 
 ### One cert per app (wildcard SANs only for ingress-per-unit)
 ```
@@ -108,3 +112,8 @@ echo | openssl s_client -showcerts -connect google.com:443 | openssl x509 -noout
 - StackExchange projects have wildcard SAN.
 - Government and bank websites have a separate cert per subdomain (and they do not use wildcard SANs)
 - Let's Encrypt limits to 100 SANs per cert ([ref](https://community.letsencrypt.org/t/why-sans-are-limited-to-100-domains-only/154930))
+- Wildcard certs don't allow HTTP challenge, nor ALPN validation using ACME
+  protocol, only DNS challenge is supported
+  (https://letsencrypt.org/docs/challenge-types/).
+- For now, the only `tls-certificates` providers in the ecosystem are using the
+  DNS-01 challenge.
--- a/doc/adr/2023-10-18 subdomain routing with cert per unit.md
+++ b/doc/adr/2023-10-18 subdomain routing with cert per unit.md
@@ -1,0 +1,110 @@
+# Obtain a separate cert per every unit when using subdomain routing
+
+## Context and Problem Statement
+Until now, the TLS cert solution was designed with only path routing in mind.
+When using Traefik in the subdomain routing mode, all Ingress will use
+Traefik's default certificate, because charm code only obtains one cert (for
+the external hostname), and traefik obviously cannot match that cert with
+subdomains.
+
+https://github.com/canonical/traefik-k8s-operator/issues/244
+
+## Considered Options
+
+- One cert per app/unit (no wildcards)
+- Only one cert (as many wildcard SANs as necessary)
+- Only one cert (keep appending subdomains; no wildcards)
+- One cert per app (wildcard SANs only for ingress-per-unit)
+- One cert per app (keep appending IPU subdomains; no wildcards)
+
+## Decision Outcome
+
+Chosen option: "One cert per app/unit (no wildcards)", because we must support
+the use-case of strict subdomain routing (no wildcards).
+
+### Consequences
+
+* Good, because no wildcard certs (more secure).
+* Good, because not limited by SAN count per cert.
+* Bad, because may get rate limited by the CA.
+  * IPA: new CSR (CRR) for every added (removed) app
+  * IPU: new CSR (CRR) for every added (removed) unit
+
+## Pros and Cons of the Options
+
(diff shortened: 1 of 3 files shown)


doc/adr/2023-10-18 subdomain routing with cert per unit.md
sed-i, line 1: Could you please take a look @ghislainbourgeois, @gregory-schiano?

simskij: You'll have to ping them on MM I think, as only collaborators and participants are pinged (or at least I think so)

gregory-schiano: Got the notif (I'm using the Github bot on MM so I get a notif there :)), having a look ! 👍

doc/adr/2023-10-18 subdomain routing with cert per unit.md
```diff
@@ -0,0 +1,110 @@
...
+## Decision Outcome
+
+Chosen option: "One cert per app/unit (no wildcards)", because we must support
+the use-case of strict subdomain routing (no wildcards).
+
+### Consequences
+
+* Good, because no wildcard certs (more secure).
+* Good, because not limited by SAN count per cert.
+* Bad, because may get rate limited by the CA.
```
gregory-schiano, line 29: Let's encrypt rate limit is 50 per domain per week (www.domain.tld and sub.domain.tld count as the same domain in the limit)
Seems pretty reasonable, unless for testing purpose chances are limited that may need more than that amount

gregory-schiano: One drawback I see is that for manual certificate signing (using for example the manual-tls-certificates-operator), the signing is more difficult than a wildcard

doc/adr/2023-10-18 subdomain routing with cert per unit.md
```diff
@@ -0,0 +1,110 @@
...
+IPA: model-app.example.com
+IPU: model-app-0.example.com
+
+SAN example: *.example.com
+```
+
+* Good, because won't get rate limited by the CA.
+* Good, because simple to implement.
+* Bad, because the number of SANs per cert is limited.
+* Bad, because a wildcard cert is too permissive.
```
gregory-schiano, line 58: Also wildcard don't allow HTTP challenge, nor ALPN validation using ACME protocol, only DNS challenge is supported
See https://letsencrypt.org/docs/challenge-types/

doc/adr/2023-10-18 subdomain routing with cert per unit.md
ghislainbourgeois: Also saw this in my list of PRs to review, taking a look!

doc/adr/2023-10-18 subdomain routing with cert per unit.md
ghislainbourgeois: The rate limit seems also reasonable to me. It could be worse if we used `Only one cert` and keep modifying it, because if we are not careful the rate limit could break the only cert in use.

Overall, I think it is something we can live with.

doc/adr/2023-10-18 subdomain routing with cert per unit.md
ghislainbourgeois: Yes, for now the only `tls-certificates` providers in the ecosystem are using the `DNS-01` challenge because they do not have access to the HTTP traffic, but taking that route would limit us in the future (for example if we decide to also support Traefik's Let's Encrypt builtin support).

doc/adr/2023-10-18 subdomain routing with cert per unit.md
```diff
@@ -0,0 +1,110 @@
...
+### Only one cert (keep appending subdomains; no wildcards)
+```
+IPA: model-app.example.com
+IPU: model-app-0.example.com
+
+SAN example: app1.example.com, app2.example.com
+```
+
+* Good, because no wildcard certs (more secure).
+* Bad, because the number of SANs per cert is limited.
```
ghislainbourgeois, line 69: The rate limit would also probably hit us here. Because we are also constantly replacing the same certificate in place, it could also be risky depending on the order things are done on an ACME server for example (if the server first revokes the previous certificate, it could create outages)

ghislainbourgeois (approved):
I added some comments, but I agree with the decision and chosen approach.

src/charm.py
```diff
@@ -863,6 +863,13 @@ def _get_configs_per_app(self, relation: Relation) -> Dict[str, Any]:
 
         return config
 
+    def _get_san_for_relation(self, relation):
+        if self._routing_mode is _RoutingMode.path:
+            return self.external_host
+        else:  # _RoutingMode.subdomain
+            prefix = "somehow, from reldata"
```
sed-i, line 870: @PietroPasotti would you be able to help out obtaining the prefix using the new schema classes?
I imagine [#269](https://github.com/canonical/traefik-k8s-operator/pull/269) may have an impact on this?

dstathis:
This has been here over a year. Feel free to reopen if you want to move forward with it.

PR #275, opened 2023-10-24
Integrate with oathkeeper via `forward-auth` interface

natalian98:
## Issue

This PR adds integration with Oathkeeper - the policy decision point in Identity and Access Proxy architecture, by supporting a new `forwardAuth` middleware.

In order to integrate Traefik into IAP, it must update routing rules for the applications that requested to be protected with the `ForwardAuth` middleware. Its definition must include headers and oathkeeper's decisions address.

## Solution

This PR integrates traefik with oathkeeper with the use of `forward-auth` interface.
Oathkeeper provides Traefik with its decisions address, allowed return headers and application names that requested the IAP protection. This information is processed by Oathkeeper from `auth-proxy` relation.
Traefik updates the relation databag with app names it provides ingress for, so that Oathkeeper can compare the app names and evaluate if they can be protected using forwardAuth.
If an app is related with `ingress-per-app/leader/unit`, Traefik will update its routes and add the middleware.

## Context

[Traefik ForwardAuth upstream docs](https://doc.traefik.io/traefik/middlewares/http/forwardauth/)
[IAP interfaces design](https://docs.google.com/document/d/13a3OEQ_Q4Fwvv9F57wgBYJm4SUFQz4W5M32UVqye4Rg/edit?usp=sharing)

## Testing Instructions

1. Pack and deploy the charms.
As `forward-auth` relies on `auth-proxy` relation data, I'm using zinc-k8s from [this branch](https://github.com/natalian98/zinc-k8s-operator/tree/test-proxy-interfaces) as a dummy example to test that the relation data is then transferred to traefik.
Oathkeeper is built from [this](https://github.com/canonical/oathkeeper-operator/tree/IAM-500-forward-auth-relation-implementation) branch.
```
juju deploy ./traefik-k8s_ubuntu-20.04-amd64.charm traefik-ingress --trust --resource traefik-image=ghcr.io/canonical/traefik:2.10.4
juju deploy ./oathkeeper*.charm --resource oci-image=$(yq eval '.resources.oci-image.upstream-source' metadata.yaml) --trust
juju deploy ./zinc-k8s*.charm --resource zinc-image=ghcr.io/jnsgruk/zinc:0.4.9 --trust
```
2. Add integrations to oathkeeper:
```
juju relate oathkeeper zinc-k8s:auth-proxy
juju relate oathkeeper traefik-ingress:experimental-forward-auth
```
At this point, if zinc was not related to ingress, oathkeeper will get blocked with message `zinc-k8s is not related via ingress`.
After you relate zinc with ingress, oathkeeper will get active:
```
juju relate zinc-k8s traefik-ingress
```
3. Check the traefik routes:
```bash
juju ssh --container traefik traefik-ingress/0 bash
cd /opt/traefik/juju/
```
Zinc route should define and reference the middleware:
```yaml
# cat juju_ingress_ingress_6_zinc-k8s.yaml
http:
  middlewares:
    juju-sidecar-forward-auth-oathkeeper:
      forwardAuth:
        address: https://oathkeeper.test21.svc.cluster.local:4456/decisions
        authResponseHeaders:
        - X-User
    juju-sidecar-noprefix-test21-zinc-k8s:
      stripPrefix:
        forceSlash: false
        prefixes:
        - /test21-zinc-k8s
  routers:
    juju-test21-zinc-k8s-router:
      entryPoints:
      - web
      middlewares:
      - juju-sidecar-forward-auth-oathkeeper
      - juju-sidecar-noprefix-test21-zinc-k8s
      rule: PathPrefix(`/test21-zinc-k8s`)
      service: juju-test21-zinc-k8s-service
    juju-test21-zinc-k8s-router-tls:
      entryPoints:
      - websecure
      middlewares:
      - juju-sidecar-forward-auth-oathkeeper
      - juju-sidecar-noprefix-test21-zinc-k8s
      rule: PathPrefix(`/test21-zinc-k8s`)
      service: juju-test21-zinc-k8s-service
      tls: {}
  services:
    juju-test21-zinc-k8s-service:
      loadBalancer:
        servers:
        - url: http://zinc-k8s.test21.svc.cluster.local:4080
```

## Release Notes

- add `forward-auth` integration with oathkeeper

### DIFF
--- a/lib/charms/oathkeeper/v0/forward_auth.py
+++ b/lib/charms/oathkeeper/v0/forward_auth.py
@@ -1,0 +1,577 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Interface library for providing API Gateways with Identity and Access Proxy information.
+
+It is required to integrate with Oathkeeper (Policy Decision Point).
+
+## Getting Started
+
+To get started using the library, you need to fetch the library using `charmcraft`.
+**Note that you also need to add `jsonschema` to your charm's `requirements.txt`.**
+
+```shell
+cd some-charm
+charmcraft fetch-lib charms.oathkeeper.v0.forward_auth
+```
+
+To use the library from the requirer side, add the following to the `metadata.yaml` of the charm:
+
+```yaml
+requires:
+  forward-auth:
+    interface: forward_auth
+    limit: 1
+```
+
+Then, to initialise the library:
+```python
+from charms.oathkeeper.v0.forward_auth import ForwardAuthConfigChangedEvent, ForwardAuthRequirer
+
+class ApiGatewayCharm(CharmBase):
+    def __init__(self, *args):
+        # ...
+        self.forward_auth = ForwardAuthRequirer(self)
+        self.framework.observe(
+            self.forward_auth.on.forward_auth_config_changed,
+            self.some_event_function
+            )
+
+    def some_event_function(self, event: ForwardAuthConfigChangedEvent):
+        if self.forward_auth.is_ready():
+            # Fetch the relation info
+            forward_auth_data = self.forward_auth.get_forward_auth_data()
+            # update ingress configuration
+            # ...
+```
+"""
+
+import inspect
+import json
+import logging
+from dataclasses import asdict, dataclass, field
+from typing import Dict, List, Mapping, Optional
+
+import jsonschema
+from ops.charm import CharmBase, RelationChangedEvent, RelationCreatedEvent, RelationDepartedEvent
+from ops.framework import EventBase, EventSource, Handle, Object, ObjectEvents
+from ops.model import Relation, TooManyRelatedAppsError
+
+# The unique Charmhub library identifier, never change it
+LIBID = "3fd31fa89da34d7f9ad9b62d5f7e7b48"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+RELATION_NAME = "forward-auth"
+INTERFACE_NAME = "forward_auth"
+
+logger = logging.getLogger(__name__)
+
+FORWARD_AUTH_PROVIDER_JSON_SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://canonical.github.io/charm-relation-interfaces/docs/json_schemas/forward_auth/v0/provider.json",
+    "type": "object",
+    "properties": {
+        "decisions_address": {"type": "string", "default": None},
+        "app_names": {"type": "array", "default": None, "items": {"type": "string"}},
+        "headers": {"type": "array", "default": None, "items": {"type": "string"}},
+    },
+    "required": ["decisions_address", "app_names"],
+}
+
+FORWARD_AUTH_REQUIRER_JSON_SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://canonical.github.io/charm-relation-interfaces/docs/json_schemas/forward_auth/v0/requirer.json",
+    "type": "object",
+    "properties": {
+        "ingress_app_names": {"type": "array", "default": None, "items": {"type": "string"}},
+    },
+    "required": ["ingress_app_names"],
+}
+
+
+class ForwardAuthConfigError(Exception):
+    """Emitted when invalid forward auth config is provided."""
+
+
+class DataValidationError(RuntimeError):
+    """Raised when data validation fails on relation data."""
+
+
+def _load_data(data: Mapping, schema: Optional[Dict] = None) -> Dict:
+    """Parses nested fields and checks whether `data` matches `schema`."""
+    ret = {}
+    for k, v in data.items():
+        try:
+            ret[k] = json.loads(v)
+        except json.JSONDecodeError:
+            ret[k] = v
+
+    if schema:
+        _validate_data(ret, schema)
+    return ret
+
+
+def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
+    if schema:
+        _validate_data(data, schema)
+
+    ret = {}
+    for k, v in data.items():
+        if isinstance(v, (list, dict)):
+            try:
+                ret[k] = json.dumps(v)
+            except json.JSONDecodeError as e:
+                raise DataValidationError(f"Failed to encode relation json: {e}")
+        else:
+            ret[k] = v
+    return ret
+
+
+class ForwardAuthRelation(Object):
+    """A class containing helper methods for forward-auth relation."""
+
+    def _pop_relation_data(self, relation_id: Relation) -> None:
+        if not self.model.unit.is_leader():
+            return
+
+        if len(self.model.relations) == 0:
+            return
+
+        relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        if not relation or not relation.app:
+            return
+
+        try:
+            for data in list(relation.data[self.model.app]):
+                relation.data[self.model.app].pop(data, "")
+        except Exception as e:
+            logger.info(f"Failed to pop the relation data: {e}")
+
+
+def _validate_data(data: Dict, schema: Dict) -> None:
+    """Checks whether `data` matches `schema`.
+
+    Will raise DataValidationError if the data is not valid, else return None.
+    """
+    try:
+        jsonschema.validate(instance=data, schema=schema)
+    except jsonschema.ValidationError as e:
+        raise DataValidationError(data, schema) from e
+
+
+@dataclass
+class ForwardAuthConfig:
+    """Helper class containing configuration required by API Gateway to set up the proxy."""
+
+    decisions_address: str
+    app_names: List[str]
+    headers: List[str] = field(default_factory=lambda: [])
+
+    @classmethod
+    def from_dict(cls, dic: Dict) -> "ForwardAuthConfig":
+        """Generate ForwardAuthConfig instance from dict."""
+        return cls(**{k: v for k, v in dic.items() if k in inspect.signature(cls).parameters})
+
+    def to_dict(self) -> Dict:
+        """Convert object to dict."""
+        return {k: v for k, v in asdict(self).items() if v is not None}
+
+
+@dataclass
+class RequirerConfig:
+    """Helper class containing configuration required by Oathkeeper.
+
+    Its purpose is to evaluate whether apps can be protected by IAP.
+    """
+
+    ingress_app_names: List[str] = field(default_factory=lambda: [])
+
+    def to_dict(self) -> Dict:
+        """Convert object to dict."""
+        return {k: v for k, v in asdict(self).items() if v is not None}
+
+
+class ForwardAuthConfigChangedEvent(EventBase):
+    """Event to notify the requirer charm that the forward-auth config has changed."""
+
+    def __init__(
+        self,
+        handle: Handle,
+        decisions_address: str,
+        app_names: List[str],
+        headers: List[str],
+        relation_id: int,
+        relation_app_name: str,
+    ) -> None:
+        super().__init__(handle)
+        self.decisions_address = decisions_address
+        self.app_names = app_names
+        self.headers = headers
+        self.relation_id = relation_id
+        self.relation_app_name = relation_app_name
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {
+            "decisions_address": self.decisions_address,
+            "app_names": self.app_names,
+            "headers": self.headers,
+            "relation_id": self.relation_id,
+            "relation_app_name": self.relation_app_name,
+        }
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        self.decisions_address = snapshot["decisions_address"]
+        self.app_names = snapshot["app_names"]
+        self.headers = snapshot["headers"]
+        self.relation_id = snapshot["relation_id"]
+        self.relation_app_name = snapshot["relation_app_name"]
+
+
+class ForwardAuthConfigRemovedEvent(EventBase):
+    """Event to notify the requirer charm that the forward-auth config was removed."""
+
+    def __init__(
+        self,
+        handle: Handle,
+        relation_id: int,
+    ) -> None:
+        super().__init__(handle)
+        self.relation_id = relation_id
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {"relation_id": self.relation_id}
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        self.relation_id = snapshot["relation_id"]
+
+
+class ForwardAuthRequirerEvents(ObjectEvents):
+    """Event descriptor for events raised by `ForwardAuthRequirer`."""
+
+    forward_auth_config_changed = EventSource(ForwardAuthConfigChangedEvent)
+    forward_auth_config_removed = EventSource(ForwardAuthConfigRemovedEvent)
+
+
+class ForwardAuthRequirer(ForwardAuthRelation):
+    """Requirer side of the forward-auth relation."""
+
+    on = ForwardAuthRequirerEvents()
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str = RELATION_NAME,
+        forward_auth_requirer_config: Optional[RequirerConfig] = None,
+    ):
+        super().__init__(charm, relation_name)
+
+        self._charm = charm
+        self._relation_name = relation_name
+        self._forward_auth_requirer_config = forward_auth_requirer_config
+
+        events = self._charm.on[relation_name]
+        self.framework.observe(events.relation_created, self._on_relation_created_event)
+        self.framework.observe(events.relation_changed, self._on_relation_changed_event)
+        self.framework.observe(events.relation_departed, self._on_relation_departed_event)
+
+    def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
+        """Update the relation with requirer data when a relation is created."""
+        if not self.model.unit.is_leader():
+            return
+
+        self.update_requirer_relation_data(self._forward_auth_requirer_config, event.relation.id)
+
+    def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
+        """Get the forward-auth config and emit a custom config-changed event."""
+        if not self.model.unit.is_leader():
+            return
+
+        data = event.relation.data[event.app]
+        if not data:
+            logger.info("No provider relation data available.")
+            return
+
+        try:
+            forward_auth_data = _load_data(data, FORWARD_AUTH_PROVIDER_JSON_SCHEMA)
+        except DataValidationError as e:
+            logger.error(
+                f"Received invalid config from the provider: {e}. Config-changed will not be emitted."
+            )
+            return
+
+        decisions_address = forward_auth_data.get("decisions_address")
+        app_names = forward_auth_data.get("app_names")
+        headers = forward_auth_data.get("headers")
+
+        relation_id = event.relation.id
+        relation_app_name = event.relation.app.name
+
+        # Notify Traefik to update the routes
+        self.on.forward_auth_config_changed.emit(
+            decisions_address, app_names, headers, relation_id, relation_app_name
+        )
+
+    def _on_relation_departed_event(self, event: RelationDepartedEvent) -> None:
+        """Notify the requirer that the relation has departed."""
+        self.on.forward_auth_config_removed.emit(event.relation.id)
+
+    def update_requirer_relation_data(
+        self, ingress_app_names: Optional[RequirerConfig], relation_id: Optional[int] = None
+    ) -> None:
+        """Update the relation databag with app names that can get IAP protection."""
+        if not self.model.unit.is_leader():
+            return
+
+        if not ingress_app_names:
+            logger.error("Ingress-related app names are missing")
+            return
+
+        if not isinstance(ingress_app_names, RequirerConfig):
+            raise ValueError(f"Unexpected type: {type(ingress_app_names)}")
+
+        try:
+            relation = self.model.get_relation(
+                relation_name=self._relation_name, relation_id=relation_id
+            )
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
+
+        if not relation or not relation.app:
+            return
+
+        data = _dump_data(ingress_app_names.to_dict(), FORWARD_AUTH_REQUIRER_JSON_SCHEMA)
+        relation.data[self.model.app].update(data)
+
+    def get_provider_info(self, relation_id: Optional[int] = None) -> Optional[ForwardAuthConfig]:
+        """Get the provider information from the databag."""
+        if len(self.model.relations) == 0:
+            return None
+        try:
+            relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
+        if not relation or not relation.app:
+            return None
+
+        data = relation.data[relation.app]
+        if not data:
+            logger.info("No relation data available.")
+            return
+
+        data = _load_data(data, FORWARD_AUTH_PROVIDER_JSON_SCHEMA)
+        forward_auth_config = ForwardAuthConfig.from_dict(data)
+        logger.info(f"ForwardAuthConfig: {forward_auth_config}")
+
+        return forward_auth_config
+
+    def get_remote_app_name(self, relation_id: Optional[int] = None) -> Optional[str]:
+        """Get the remote app name."""
+        if len(self.model.relations) == 0:
+            return None
+        try:
+            relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
+        if not relation or not relation.app:
+            return None
+
+        return relation.app.name
+
+    def is_ready(self, relation_id: Optional[int] = None) -> Optional[bool]:
+        """Checks whether ForwardAuth is ready on this relation.
+
+        Returns True when Oathkeeper shared the config; False otherwise.
+        """
+        if len(self.model.relations) == 0:
+            return None
+        try:
+            relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
+
+        if not relation or not relation.app:
+            return None
+
+        return (
+            "decisions_address" in relation.data[relation.app]
+            and "app_names" in relation.data[relation.app]
+        )
+
+
+class ForwardAuthProxySet(EventBase):
+    """Event to notify the charm that the proxy was set successfully."""
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {}
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        pass
+
+
+class InvalidForwardAuthConfigEvent(EventBase):
+    """Event to notify the charm that the forward-auth configuration is invalid."""
+
+    def __init__(self, handle: Handle, error: str):
+        super().__init__(handle)
+        self.error = error
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {
+            "error": self.error,
+        }
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        self.error = snapshot["error"]
+
+
+class ForwardAuthRelationRemovedEvent(EventBase):
+    """Event to notify the charm that the relation was removed."""
+
+    def __init__(
+        self,
+        handle: Handle,
+        relation_id: int,
+    ) -> None:
+        super().__init__(handle)
+        self.relation_id = relation_id
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {"relation_id": self.relation_id}
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        self.relation_id = snapshot["relation_id"]
+
+
+class ForwardAuthProviderEvents(ObjectEvents):
+    """Event descriptor for events raised by `ForwardAuthProvider`."""
+
+    forward_auth_proxy_set = EventSource(ForwardAuthProxySet)
+    invalid_forward_auth_config = EventSource(InvalidForwardAuthConfigEvent)
+    forward_auth_relation_removed = EventSource(ForwardAuthRelationRemovedEvent)
+
+
+class ForwardAuthProvider(ForwardAuthRelation):
+    """Provider side of the forward-auth relation."""
+
+    on = ForwardAuthProviderEvents()
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str = RELATION_NAME,
+        forward_auth_config: Optional[ForwardAuthConfig] = None,
+    ) -> None:
+        super().__init__(charm, relation_name)
+        self.charm = charm
+        self._relation_name = relation_name
+        self.forward_auth_config = forward_auth_config
+
+        events = self.charm.on[relation_name]
+        self.framework.observe(events.relation_created, self._on_relation_created_event)
+        self.framework.observe(events.relation_changed, self._on_relation_changed_event)
+        self.framework.observe(events.relation_departed, self._on_relation_departed_event)
+
+    def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
+        """Update the relation with provider data when a relation is created."""
+        if not self.model.unit.is_leader():
+            return
+
+        try:
+            self._update_relation_data(self.forward_auth_config, event.relation.id)
+        except ForwardAuthConfigError as e:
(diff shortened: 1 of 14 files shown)


PietroPasotti:
Thanks Natalia, will review when I have time after/during the sprint (or probably next week we can go through it together if you like).

PietroPasotti:
In order to make it clear that the feature is untested and might be unstable/experimental, we should rename the endpoint to `experimental_forward_auth`. The description should make it clear what it does and that it's insecure and that you're supposed to have traefik enabled as TLS termination endpoints.

For now you'll have to do `juju relate traefik:experimental_forward_auth oathkeeper`, and then when the feature becomes stable and oathkeeper has TLS, we can rename the interface to `forward_auth`.

Secondly, we will only enable the experimental forward auth feature behind a feature flag.
You'll need to `juju config traefik enable_experimental_forward_auth=True` in order to use it.
All code blocks introduced by this PR should be behind a:
`if self.config["enable_experimental_forward_auth"]:` guard so that we can easily filter out any buggy behaviour due to this feature.

If you relate traefik to oathkeeper but `enable_experimental_forward_auth` is False, Traefik will not enable the middleware.

natalian98:
> In order to make it clear that the feature is untested and might be unstable/experimental, we should rename the endpoint to `experimental_forward_auth`. The description should make it clear what it does and that it's insecure and that you're supposed to have traefik enabled as TLS termination endpoints.
> 
> For now you'll have to do `juju relate traefik:experimental_forward_auth oathkeeper`, and then when the feature becomes stable and oathkeeper has TLS, we can rename the interface to `forward_auth`.
> 
> Secondly, we will only enable the experimental forward auth feature behind a feature flag. You'll need to `juju config traefik enable_experimental_forward_auth=True` in order to use it. All code blocks introduced by this PR should be behind a: `if self.config["enable_experimental_forward_auth"]:` guard so that we can easily filter out any buggy behaviour due to this feature.
> 
> If you relate traefik to oathkeeper but `enable_experimental_forward_auth` is False, Traefik will not enable the middleware.

Wouldn't it be enough to have the feature flag as a juju config? I'd prefer to avoid renaming the relation.

tests/unit/test_charm.py
```diff
@@ -95,26 +95,6 @@ def _render_middlewares(*, strip_prefix: bool = False, redirect_https: bool = Fa
...
-                    "prefixes": ["/test-model-remote-0"],
-                    "forceSlash": False,
-                }
-            }
-        )
-    return (
-        {"middlewares": {"juju-sidecar-noprefix-test-model-remote-0": middlewares}}
-        if middlewares
-        else {}
-    )
```
natalian98, line 115: this method was duplicated

src/charm.py
```diff
@@ -930,6 +974,20 @@ def _generate_middleware_config(
           "cannot create middleware: multi-types middleware not supported, consider declaring two
           different pieces of middleware instead"
         """
+        forwardauth_middleware = {}
+        if self.forward_auth.is_ready():
+            forward_auth_config = self.forward_auth.get_provider_info()
+            if data.get("name") in forward_auth_config.app_names:  # type: ignore
```
sed-i, line 980: Don't quite follow what this condition does. Could you please explain?

tests/unit/test_charm.py
```diff
@@ -409,6 +389,71 @@ def test_tcp_config(self):
...
+
+        _ = self.setup_forward_auth_relation()
+
+        self.assertTrue(self.harness.charm.forward_auth.is_ready())
+
+        expected_provider_info = self.harness.charm.forward_auth.get_provider_info()
+
+        assert expected_provider_info.decisions_address == provider_info["decisions_address"]
+        assert expected_provider_info.app_names == provider_info["app_names"]
+        assert expected_provider_info.headers == provider_info["headers"]
```
sed-i, line 427: **Question:** Do these asserts belong here or in forwardauth?

It would be useful to assert that the decisions_address is present in the rendered traefik config yaml.

PietroPasotti:
> > In order to make it clear that the feature is untested and might be unstable/experimental, we should rename the endpoint to `experimental_forward_auth`. The description should make it clear what it does and that it's insecure and that you're supposed to have traefik enabled as TLS termination endpoints.
> > For now you'll have to do `juju relate traefik:experimental_forward_auth oathkeeper`, and then when the feature becomes stable and oathkeeper has TLS, we can rename the interface to `forward_auth`.
> > Secondly, we will only enable the experimental forward auth feature behind a feature flag. You'll need to `juju config traefik enable_experimental_forward_auth=True` in order to use it. All code blocks introduced by this PR should be behind a: `if self.config["enable_experimental_forward_auth"]:` guard so that we can easily filter out any buggy behaviour due to this feature.
> > If you relate traefik to oathkeeper but `enable_experimental_forward_auth` is False, Traefik will not enable the middleware.
> 
> Wouldn't it be enough to have the feature flag as a juju config? I'd prefer to avoid renaming the relation.

We don't have to rename the interface, only the endpoint. That seems like a minor change.
We want to avoid the impression that traefik offers stable support for forward_auth. A config-toggled feature flag is a bit hidden: someone might skip reading the config, deploy traefik, relate forward-auth, and be surprised when it doesn't work.

src/charm.py
natalian98: Sure. The `_generate_middleware_config` method generates middlewares for each charm related with one of the ingress interfaces.
In this block I want to check if the app name, for which the middlewares are currently generated (`data.get("name")`), requested IAP protection (`forward_auth_config.app_names`). If so, the `forwardAuth` definition should be added.
Let me know if it's clearer now.

src/charm.py
```diff
@@ -930,6 +974,20 @@ def _generate_middleware_config(
...
+                policy_decision_point_app = self.forward_auth.get_remote_app_name()
```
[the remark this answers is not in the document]
natalian98: `app_name` could be confused with `forward_auth_config.app_names`. I think the original variable name is more descriptive because it points to oathkeeper as the policy decision point (which is what forwardAuth talks to in IAP architecture, and thus we want to include it in the middleware name).

src/charm.py
```diff
@@ -266,6 +284,26 @@ def __init__(self, *args):
         # Action handlers
         observe(self.on.show_proxied_endpoints_action, self._on_show_proxied_endpoints)  # type: ignore
 
+    @property
+    def _forward_auth_requirer_config(self) -> RequirerConfig:
```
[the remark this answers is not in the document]
natalian98: Since the provider has a dataclass, I'd prefer the requirer to have it as well for it to be coherent. I also don't think the requirer config will be subject to change once this PR is finalized.

Would `ForwardAuthRequirerConfig` be better?

src/charm.py
```diff
@@ -221,6 +227,11 @@ def __init__(self, *args):
                 self.on.update_status,  # type: ignore
             ],
         )
+
+        self.forward_auth = ForwardAuthRequirer(
+            self, forward_auth_requirer_config=self._forward_auth_requirer_config
```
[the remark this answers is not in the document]
natalian98: The intention was to differentiate between forward-auth's config provider (oathkeeper) and traefik as the requirer.
Since traefik only provides app names related via ingress, perhaps `ingress_app_names` would sound better? What do you think?

src/charm.py
```diff
@@ -254,6 +265,13 @@ def __init__(self, *args):
             self._on_recv_ca_cert_removed,
         )
 
+        observe(
+            self.forward_auth.on.forward_auth_config_changed, self._on_forward_auth_config_changed
```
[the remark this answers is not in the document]
natalian98: I agree that it would look ambiguous if we named a custom event similarly to a built-in one.
Changed to `auth_config_changed` and `auth_config_removed`.

src/charm.py
```diff
@@ -730,6 +771,9 @@ def _handle_ingress_data_removed(self, event: RelationEvent):
             event.relation, wipe_rel_data=not isinstance(event, RelationBrokenEvent)
         )
 
+        if self.forward_auth.is_ready():
+            self.forward_auth.update_requirer_relation_data(self._forward_auth_requirer_config)
```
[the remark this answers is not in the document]
natalian98: My understanding was that the ingress relation is first wiped in `_wipe_ingress_for_relation()`, then the forward-auth relation is updated with a "fresh" list of app names related via ingress. Is that not the case?

src/charm.py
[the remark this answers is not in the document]
sed-i: Not sure what you mean but seems fine 😅

src/charm.py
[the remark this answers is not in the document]
sed-i: `ForwardAuthRequirerConfig` is clearer, but before we take that path: it's fine for the lib to use dataclasses. The question is should they leak outside of the lib. It's very pythonic to have kwargs.

src/charm.py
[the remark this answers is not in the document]
sed-i: I didn't look into code ordering. Code ordering here or any other patch to ops/888 are not ideal, so if it works - it works :)

src/charm.py
sed-i: Thanks!
Should this logic (including `is_ready`?) be owned by forward_auth lib? E.g.
```python
if self.forward_auth.is_protected_app(data.get("name")):
```

metadata.yaml
```diff
@@ -72,6 +72,12 @@ requires:
     limit: 1
     description: |
       Send a CSR to- and obtain a signed certificate from an external CA.
+  forward-auth:
+    interface: forward_auth
+    limit: 1
+    description: |
+      Receives forwardAuth middleware config from Oathkeeper.
+      The same middleware will be applied to all charms that requested Identity and Access Proxy protection.
```
sed-i, line 80: Just updating with my understanding, which could be wrong!
```suggestion
      Receive config from e.g. oathkeeper, for rendering the forwardAuth middleware.
      The same middleware is applied to all proxied endpoints that requested Identity and Access Proxy (IAP) protection.
      For this reason we set a relation count limit of 1.
```

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+    def _on_relation_departed_event(self, event: RelationDepartedEvent) -> None:
+        """Notify the requirer that the relation has departed."""
+        self.on.auth_config_removed.emit(event.relation.id)
+
+    def update_requirer_relation_data(
+        self, ingress_app_names: Optional[RequirerConfig], relation_id: Optional[int] = None
+    ) -> None:
+        """Update the relation databag with app names that can get IAP protection."""
+        if not self.model.unit.is_leader():
+            return
```
sed-i, line 334: We have a `limit: 1` set so app data is somewhat equivalent to unit data, but I wonder if it makes more sense to use unit data for this. Perhaps it would help avoid some edge cases with leadership status?

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+
+        if not isinstance(ingress_app_names, RequirerConfig):
+            raise ValueError(f"Unexpected type: {type(ingress_app_names)}")
+
+        try:
+            relation = self.model.get_relation(
+                relation_name=self._relation_name, relation_id=relation_id
+            )
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
```
sed-i, line 348: `TooManyRelatedAppsError` is native to ops so it's fine to let it be raised.
Converting it to RuntimeError seems a bit counter intuitive.

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+        """Update the relation databag with app names that can get IAP protection."""
+        if not self.model.unit.is_leader():
+            return
+
+        if not ingress_app_names:
+            logger.error("Ingress-related app names are missing")
+            return
+
+        if not isinstance(ingress_app_names, RequirerConfig):
+            raise ValueError(f"Unexpected type: {type(ingress_app_names)}")
```
sed-i, line 341: [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError)?

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+        try:
+            relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        except TooManyRelatedAppsError:
+            raise RuntimeError("More than one relation is defined. Please provide a relation_id")
+        if not relation or not relation.app:
+            return None
+
+        data = relation.data[relation.app]
+        if not data:
+            logger.info("No relation data available.")
```
sed-i, line 369: I'd drop this, or change to debug level.

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+            return
+
+        data = _load_data(data, FORWARD_AUTH_PROVIDER_JSON_SCHEMA)
+        forward_auth_config = ForwardAuthConfig.from_dict(data)
+        logger.info(f"ForwardAuthConfig: {forward_auth_config}")
```
sed-i, line 374: Debug level?

lib/charms/oathkeeper/v0/forward_auth.py
natalian98: There should always be a leader which will write to the databag. What edge cases do you have in mind?

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,577 @@
...
+        data = event.relation.data[event.app]
+        if not data:
+            logger.info("No provider relation data available.")
+            return
+
+        try:
+            forward_auth_data = _load_data(data, FORWARD_AUTH_PROVIDER_JSON_SCHEMA)
+        except DataValidationError as e:
+            logger.error(
+                f"Received invalid config from the provider: {e}. Config-changed will not be emitted."
```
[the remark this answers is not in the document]
natalian98: This is handled in oathkeeper status. If the config was invalid, it will get blocked

src/charm.py
natalian98: I'm not sure we should move all the logic to the lib, especially given that the `forward_auth.is_ready():` check is required in a few blocks anyways

sed-i:
Hi @natalian98 

I deployed the following bundle (you can add it to `./tests/manual` if you'd like):
```yaml
bundle: kubernetes
applications:
  ok:
    # From https://github.com/canonical/oathkeeper-operator
    charm: ./oathkeeper-operator/oathkeeper_ubuntu-22.04-amd64.charm
    scale: 1
    constraints: arch=amd64
    trust: true
    resources:
      oci-image: "ghcr.io/canonical/oathkeeper:0.40.6"
  trfk:
    # From https://github.com/canonical/traefik-k8s-operator/pull/275
    charm: ./traefik-k8s-operator/traefik-k8s_ubuntu-20.04-amd64.charm
    series: focal
    scale: 1
    constraints: arch=amd64
    storage:
    trust: true
    resources:
      traefik-image: "docker.io/ubuntu/traefik:2-22.04"  
zinc:
    # From https://github.com/natalian98/zinc-k8s-operator/tree/test-proxy-interfaces
    charm: ./zinc-k8s-operator/zinc-k8s_ubuntu-22.04-amd64.charm
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
    trust: true
    resources:
      zinc-image: "ghcr.io/jnsgruk/zinc:0.4.9"
relations:
- - ok:forward-auth
  - trfk:experimental-forward-auth
- - ok:auth-proxy
  - zinc:auth-proxy
- - zinc:ingress
  - trfk:ingress
```

Then oathkeeper is stuck in 
```
Unit     Workload  Agent      Address       Ports  Message
ok/0*    waiting   executing  10.1.166.110         Waiting for Oathkeeper service
```

and debug-log has a recurring error:

```
unit-ok-0: 14:55:12.963 ERROR unit.ok/0.juju-log Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/lightkube/core/generic_client.py", line 188, in raise_for_status
    resp.raise_for_status()
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/httpx/_models.py", line 758, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://10.152.183.1/apis/apps/v1/namespaces/auth2/statefulsets/oathkeeper?fieldManager=ok'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

During handling of the above exception, another exception occurred:

  File "/var/lib/juju/agents/unit-ok-0/charm/./src/charm.py", line 555, in <module>
    main(OathkeeperCharm)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/ops/main.py", line 436, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/ops/framework.py", line 340, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/ops/framework.py", line 842, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/ops/framework.py", line 931, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-ok-0/charm/./src/charm.py", line 329, in _on_oathkeeper_pebble_ready
    self._patch_statefulset()
  File "/var/lib/juju/agents/unit-ok-0/charm/./src/charm.py", line 318, in _patch_statefulset
    self.client.patch(StatefulSet, name=self.meta.name, namespace=self.model.name, obj=patch)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/lightkube/core/client.py", line 325, in patch
    return self._client.request("patch", res=res, name=name, namespace=namespace, obj=obj,
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/lightkube/core/generic_client.py", line 245, in request
    return self.handle_response(method, resp, br)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/lightkube/core/generic_client.py", line 196, in handle_response
    self.raise_for_status(resp)
  File "/var/lib/juju/agents/unit-ok-0/charm/venv/lightkube/core/generic_client.py", line 190, in raise_for_status
    raise transform_exception(e)
lightkube.core.exceptions.ApiError: statefulsets.apps "oathkeeper" not found
unit-ok-0: 14:55:13.190 ERROR juju.worker.uniter.operation hook "oathkeeper-pebble-ready" (via hook dispatching script: dispatch) failed: exit status 1
unit-ok-0: 14:55:13.191 ERROR juju.worker.uniter pebble poll failed for container "oathkeeper": failed to send pebble-ready event: hook failed
```

src/charm.py
sed-i: It's fine to keep `is_ready` public.
It seems that the flow `is_ready -> get_provider_info -> app_name in config.app_names` requires some know-how about how the lib is built, and could potentially be accomplished by a lib method?

lib/charms/oathkeeper/v0/forward_auth.py
sed-i: Haven't examined all code paths but when we sprinkle leader guards we potentially need to make sure the logic is executed on [leader elected](https://discourse.charmhub.io/t/event-leader-elected/6470) too. Moreover, on [upgrade](https://discourse.charmhub.io/t/event-upgrade-charm/6485), we don't get relation events nor leader elected (only leader-settings-changed). So if after upgrade we need to update reldata, then we suddenly need to be mindful of that too.

If using unit data is possible, we won't need leader guards anymore.

tests/integration/test_forward_auth.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@retry(
+    wait=wait_exponential(multiplier=3, min=1, max=10),
+    stop=stop_after_attempt(10),
+    reraise=True,
+)
+def assert_anonymous_response(url):
+    resp = requests.get(url, verify=False)
+    assert resp.status_code == 200
```
shipperizer, line 148: if we used an async client (aiohttp) we could drop the retry bit in this case
we'd still be blocking here though https://github.com/canonical/traefik-k8s-operator/pull/275/commits/13f93cae41764bab15e7e5ad48f228b5482e8322#diff-03198ed59d11f58b5b2b25d32d96a8726c03c0fe4b25df59e1cc7cbefad31e8eR160

src/charm.py
```diff
@@ -730,6 +776,9 @@ def _handle_ingress_data_removed(self, event: RelationEvent):
             event.relation, wipe_rel_data=not isinstance(event, RelationBrokenEvent)
         )
 
+        if self.forward_auth.is_ready():
+            self.forward_auth.update_requirer_relation_data(self._forward_auth_config)
```
nsklikas, line 780: Won't the relation still exist? This means that this call will not remove the app_name from the `forward_auth` databag. This should probably be handled by `_process_status_and_configurations`.

src/charm.py
```diff
@@ -266,6 +279,32 @@ def __init__(self, *args):
...
+                self.ingress_per_appv2.relations,
+                self.ingress_per_unit.relations,
+                self.traefik_route.relations,
+            )
+        ]
+        return ForwardAuthRequirerConfig(ingress_app_names)
+
+    def _on_forward_auth_config_changed(self, event: AuthConfigChangedEvent):
+        if self.config["enable_experimental_forward_auth"]:
+            self.forward_auth.update_requirer_relation_data(self._forward_auth_config)
```
nsklikas, line 297: Do we need to update the app_names when the provider's databag changes? I think that these should be independent actions (i.e. when the provider's databag changes, the only action needed is that traefik updates its ingress routes config).

I think that traefik should update the requirer databag when the relation is created and then on `_process_status_and_configurations`

src/charm.py
```diff
@@ -721,6 +764,9 @@ def _handle_ingress_data_provided(self, event: RelationEvent):
         # update-status.
         self._process_status_and_configurations()
 
+        if self.forward_auth.is_ready():
```
nsklikas, line 767: Does it matter if the `forward_auth` provider side is ready? I think that we only need to check if `enable_experimental_forward_auth` is true.

src/charm.py
```diff
@@ -930,6 +979,21 @@ def _generate_middleware_config(
...
           different pieces of middleware instead"
         """
+        forwardauth_middleware = {}
+        if self.config["enable_experimental_forward_auth"]:
+            if self.forward_auth.is_ready():
+                forward_auth_config = self.forward_auth.get_provider_info()
+                if data.get("name") in forward_auth_config.app_names:  # type: ignore
+                    policy_decision_point_app = self.forward_auth.get_remote_app_name()
+                    forwardauth_middleware[
+                        f"juju-sidecar-forward-auth-{policy_decision_point_app}"
```
nsklikas, line 989: I think we need to suffix the middleware with the `prefix`
```suggestion
                        f"juju-sidecar-forward-auth-{prefix}"
```

lib/charms/oathkeeper/v0/forward_auth.py
natalian98: Aren't unit databags supposed to be used when each unit's data differs? In this case all units would write the same data. Having each unit write to its own databag could also introduce race conditions.
For edge cases, we can rely on `update-status` where the databag will be updated from `_process_status_and_configurations`.

sed-i: That's fair.
Not sure how the races would differ between app/unit data, but scaling traefik is kinda hypothetical.
As this is internal to the lib, could always change it later per need.

tests/unit/test_charm.py
natalian98: This is covered in integration tests

src/charm.py
```diff
@@ -602,6 +640,10 @@ def _on_config_changed(self, _: ConfigChangedEvent):
         # reconsider all data sent over the relations and all configs
         new_external_host = self.external_host
         new_routing_mode = self.config["routing_mode"]
+
+        if self.config["enable_experimental_forward_auth"]:
+            self.forward_auth.update_requirer_relation_data(self._forward_auth_config)
```
nsklikas, line 645: Is this needed? Changes in the URL should not affect the `forward_auth` relation, also `_process_status_and_configurations` will be called if something changed.

lib/charms/oathkeeper/v0/forward_auth.py
```diff
@@ -0,0 +1,598 @@
...
+        )
+
+    def is_protected_app(self, app: str) -> bool:
+        if self.is_ready():
+            forward_auth_config = self.get_provider_info()
+            if forward_auth_config and app in forward_auth_config.app_names:
+                return True
+            return False
+
+        return False
```
nsklikas, line 427: Reminder: This (and maybe some other changes) will need to be pushed to oathkeeper before merging this PR.

src/charm.py
natalian98: Right, I removed that part and added a check to call `_process_status_and_configurations` if the forward-auth config flag has changed.

lib/charms/oathkeeper/v0/forward_auth.py
natalian98: Ref PR: https://github.com/canonical/oathkeeper-operator/pull/38

src/charm.py
```diff
@@ -498,6 +547,9 @@ def _update_ingress_configurations(self):
 
         errors = False
 
+        if self._is_forward_auth_enabled:
```
PietroPasotti, line 550: don't we need to check forward_auth.ready() here?

src/charm.py
```diff
@@ -688,6 +740,9 @@ def _get_configs_per_leader(self, relation: Relation) -> Dict[str, Any]:
             redirect_https=data.get("redirect-https", False),
             strip_prefix=data.get("strip-prefix", False),
             external_host=self.external_host,
+            enable_experimental_forward_auth=self._is_forward_auth_enabled,
```
PietroPasotti, line 743: I'd make `experimental_forward_auth` an init argument to `self.traefik`, since we seem to be passing it in multiple times.

PietroPasotti: (next to `tls_enabled: same pattern there)

PietroPasotti (changes requested):
A couple of points regarding the API changes made to the Traefik class, but the rest looks good to me!

PietroPasotti (approved):

sed-i:
Tested successfully with the bundle below.

```
$ juju run trfk/0 show-proxied-endpoints
proxied-endpoints: '{"zinc": {"url": "http://10.43.8.206/auth2-zinc"}}'

$ curl http://10.43.8.206/auth2-zinc/version
{"version":"'0.4.9'","build":"0","commit_hash":"'249a843897ade8ca3f7e80ac6036b8ac0a27154a'","branch":"0","build_date":"'2023-09-13_09:50:04AM-GMT'"}%

$ curl http://10.43.8.206/auth2-zinc/versionn
{"error":{"code":401,"status":"Unauthorized","message":"The request could not be authorized"}}
```

@natalian98 do we have issues open to capture the gaps currently addressed by:
- config option: `enable_experimental_forward_auth: true`
- config option: `dev: true`
- relation: `trfk:experimental-forward-auth`

---

```yaml
bundle: kubernetes
applications:
  ok:
    # From https://github.com/canonical/oathkeeper-operator
    # 1e23771 - (HEAD -> address-forward-auth-comments, origin/address-forward-auth-comments) refactor
    charm: ./oathkeeper-operator/oathkeeper_ubuntu-22.04-amd64.charm
    scale: 1
    constraints: arch=amd64
    trust: true
    resources:
      oci-image: "ghcr.io/canonical/oathkeeper:0.40.6"
    options:
      # until tls is impl'd dev mode would resort to http decisions url
      dev: true
  trfk:
    # From https://github.com/canonical/traefik-k8s-operator/pull/275
    # cc35a85 - (HEAD -> IAM-500-forward-auth-relation-implementation, natalian98/IAM-500-forward-auth-relation-implementation) refactor: add experimental_forward_auth_enabled as an init arg
    charm: ./traefik-k8s-operator/traefik-k8s_ubuntu-20.04-amd64.charm
    series: focal
    scale: 1
    constraints: arch=amd64
    storage:
    trust: true
    resources:
      traefik-image: "docker.io/ubuntu/traefik:2-22.04"
    options:
      enable_experimental_forward_auth: true
  zinc:
    # From https://github.com/natalian98/zinc-k8s-operator
    # 0c19865 - (HEAD -> test-proxy-interfaces, origin/test-proxy-interfaces) fix: auth-proxy
    charm: ./zinc-k8s-operator/zinc-k8s_ubuntu-22.04-amd64.charm
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
    trust: true
    resources:
      zinc-image: "ghcr.io/jnsgruk/zinc:0.4.9"
relations:
- - ok:forward-auth
  - trfk:experimental-forward-auth
- - ok:auth-proxy
  - zinc:auth-proxy
- - zinc:ingress
  - trfk:ingress
```

src/charm.py
```diff
@@ -250,6 +265,37 @@ def _service_ports(self) -> List[ServicePort]:
...
+    @property
+    def _is_forward_auth_enabled(self) -> bool:
+        if self.config["enable_experimental_forward_auth"]:
+            return True
+        return False
+
+    def _on_forward_auth_config_changed(self, event: AuthConfigChangedEvent):
+        if self._is_forward_auth_enabled:
+            if self.forward_auth.is_ready():
+                self._process_status_and_configurations()
```
sed-i, line 290: **Question**: In case `is_ready` is False during this hook, it will be addressed on update-statue the latest. Is that ok?

Would be useful to have this situation captured in a unit test.

src/charm.py
```diff
@@ -434,15 +480,19 @@ def _on_config_changed(self, _: ConfigChangedEvent):
...
 
         if (
             self._stored.current_external_host != new_external_host  # type: ignore
             or self._stored.current_routing_mode != new_routing_mode  # type: ignore
+            or self._stored.current_forward_auth_mode != new_forward_auth_mode  # type: ignore
         ):
             self._stored.current_external_host = new_external_host  # type: ignore
             self._stored.current_routing_mode = new_routing_mode  # type: ignore
+            self._stored.current_forward_auth_mode = new_forward_auth_mode  # type: ignore
             self._process_status_and_configurations()
```
sed-i, line 496: **Suggestion**: update stored state only after calling `_process_status_and_configurations`.

We had an [issue](https://github.com/canonical/cos-proxy-operator/pull/89) recently when `relation-set` failed, and upon hook retry the relation update logic wasn't executed because stored state was already updated from before.

Since `_process_status_and_configurations` updates reldata, we could call it every time regardless of the stored state, or before we update stored state (and pass args directly instead of relying on stored state?).

tests/integration/testers/forward-auth/lib/charms/oathkeeper/v0/auth_proxy.py
sed-i, line 1: **Suggestion**: This file should probably not be committed? We will likely forget to update it.
See [`copy_traefik_library_into_tester_charms`](https://github.com/canonical/traefik-k8s-operator/blob/816a33727b79161ad7eeb3b11022088bd25626a5/tests/integration/conftest.py#L69)

tests/integration/test_forward_auth.py
```diff
@@ -0,0 +1,188 @@
...
+    await ops_test.model.integrate(f"{IAP_REQUIRER_CHARM}:auth-proxy", OATHKEEPER_CHARM)
+
+    await ops_test.model.integrate(f"{TRAEFIK_CHARM}:experimental-forward-auth", OATHKEEPER_CHARM)
+
+    await ops_test.model.wait_for_idle(
+        [TRAEFIK_CHARM, OATHKEEPER_CHARM, IAP_REQUIRER_CHARM], status="active", timeout=1000
+    )
+
+
+@retry(
```
sed-i, line 83: Could you please add a comment why the retry is needed after the `wait_for_idle` above?
Can we use `block_until` some condition flips, e.g. oathkeeper endpoint?

tests/integration/test_forward_auth.py
```diff
@@ -0,0 +1,188 @@
...
+    headers = json.loads(resp.content).get("headers")
+    assert headers["X-User"] == "anonymous"
+
+
+@retry(
+    wait=wait_exponential(multiplier=3, min=1, max=10),
+    stop=stop_after_attempt(5),
+    reraise=True,
+)
+def update_access_rules_configmap(ops_test: OpsTest, lightkube_client: Client, rule):
```
sed-i, line 160: Could you please add a comment why direct k8s mods are needed for this itest?

src/charm.py
natalian98: As mentioned, we rely on `update-status` for edge cases. Do you think we should defer the event instead?

tests/integration/testers/forward-auth/lib/charms/oathkeeper/v0/auth_proxy.py
natalian98: That lib is needed by the iap-tester charm. Should I place it in traefik's `lib` dir and copy from there even though traefik won't directly use it?

sed-i: Hmm good point. Is there a charm on charmhub we could use instead of the tester charm?
We can keep things as is until there is one I imagine.

src/charm.py
sed-i: We should avoid both 😅 
But I see that `self._process_status_and_configurations()` is traefik's "common exit hook" so no issue here at all!

tests/integration/testers/forward-auth/lib/charms/oathkeeper/v0/auth_proxy.py
natalian98: https://github.com/canonical/oathkeeper-operator/issues/47

PR #279, opened 2023-12-01
deprecate ingress v1

PietroPasotti:
On module import, warn users of v1 ingress that the lib is deprecated and give instructions to bump to v2.

Only do that in the requirer wrapper's `__init__` because traefik still supports v1 and therefore also uses the module.

### DIFF
--- a/lib/charms/traefik_k8s/v1/ingress.py
+++ b/lib/charms/traefik_k8s/v1/ingress.py
@@ -1,54 +1,18 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-r"""# Interface Library for ingress.
+r"""# [DEPRECATED!] Interface Library for ingress.
+This is a DEPRECATED version of the Ingress interface library.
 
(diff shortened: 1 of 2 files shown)


lib/charms/traefik_k8s/v1/ingress.py
```diff
@@ -450,6 +415,13 @@ def __init__(
         Request Args:
             port: the port of the service
         """
+        log.warning(
+            "The ``ingress v1`` library is DEPRECATED in favour of ``ingress v2`` "
+            "and no longer maintained. This library does NOT in fact implement the "
+            "``ingress`` interface, but, instead, the ``ingress-per-leader`` one."
+            "Please bump with ``charmcraft fetch-lib chars.traefik_k8s.v2.ingress``."
```
lucabello, line 422:
```suggestion
            "Please bump with ``charmcraft fetch-lib charms.traefik_k8s.v2.ingress``."
```

lucabello (approved):

sed-i (approved):

PR #282, opened 2024-01-26
fixed some issues when removing an ingress relation

PietroPasotti:
## Issue
When removing a relation between traefik and, say, alertmanager, traefik goes into error state because it fails to wipe the relation databags due to some permission error.

Solution: be more careful with databag cleanup.


src/charm.py
```diff
@@ -890,7 +900,15 @@ def _wipe_ingress_for_relation(self, relation: Relation, *, wipe_rel_data=True):
         # Skip this for traefik-route because it doesn't have a `wipe_ingress_data` method.
         provider = self._provider_from_relation(relation)
         if wipe_rel_data and self.unit.is_leader() and provider != self.traefik_route:
-            provider.wipe_ingress_data(relation)  # type: ignore  # this is an ingress-type relation
+            try:
+                provider.wipe_ingress_data(relation)  # type: ignore  # this is an ingress-type relation
```
sed-i, line 904: Would it work if we do nothing for relation broken?
Juju takes care of the databag - we probably shouldn't worry about cleanup.

Seems wrong to log an ERROR with a reservation.

PietroPasotti: mmm, true, not sure what's the goal semantically of wiping the databags. historically traefik's been doing it forever, it's just habit now. 
@simskij perhaps has some context?

PietroPasotti: perhaps it's a way for the charm to mark for itself that the relation is going away and should not be used.

dstathis:
This appears to be stale. Please reopen.

PR #286, opened 2024-02-02
ingress lib support for pydantic v2

PietroPasotti:
This PR makes the ingress library compatible with pydantic v1 and pydantic v2.

### DIFF
--- a/lib/charms/traefik_k8s/v2/ingress.py
+++ b/lib/charms/traefik_k8s/v2/ingress.py
@@ -13,7 +13,7 @@ To get started using the library, you just need to fetch the library using `char
 
 ```shell
 cd some-charm
-charmcraft fetch-lib charms.traefik_k8s.v2.ingress
+charmcraft fetch-lib charms.traefik_k8s.v1.ingress
 ```
 
 In the `metadata.yaml` of the charm, add the following:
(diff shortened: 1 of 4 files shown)


lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -13,7 +13,7 @@
 
 ```shell
 cd some-charm
-charmcraft fetch-lib charms.traefik_k8s.v2.ingress
+charmcraft fetch-lib charms.traefik_k8s.v1.ingress
```
mmkay, line 16: This should stay at v2, I think?

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -72,69 +72,131 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
...
+            if nest_under:
+                return cls.model_validate(json.loads(databag[nest_under]))
 
-        if databag is None:
-            databag = {}
+            try:
+                data = {k: json.loads(v) for k, v in databag.items() if k not in BUILTIN_JUJU_KEYS}
+            except json.JSONDecodeError as e:
+                msg = f"invalid databag contents: expecting json. {databag}"
+                logger.error(msg)
```
mmkay, line 163: Most likely should be `log` - logger in this lib is under a `log` reference.

mmkay (approved):
There's a couple of failing pipelines that might need updating, but overall I think it's better than creating a `v3` ingress just for pydantic upgrade. A couple of minor comments left.

PietroPasotti:
note the unittest failures are due to what's being fixed in #287

PR #298, opened 2024-02-28
http->https redirect

### LINKED ISSUE #177: HTTP to HTTPS redirect

PietroPasotti:
Configures traefik to redirect by default http to https if tls is enabled.

Replaces #246 
Fixes #177

### DIFF
--- a/src/traefik.py
+++ b/src/traefik.py
@@ -199,11 +199,14 @@ class Traefik:
         tcp_entrypoints = self._tcp_entrypoints
         logger.debug(f"Statically configuring traefik with tcp entrypoints: {tcp_entrypoints}.")
 
+
         web_config = {
             "address": f":{self.port}",
-            # enable by default http -> https redirect
-            "http": {"redirections": {"entryPoint": {"to": "websecure", "scheme": "https"}}},
         }
+
+        if self._tls_enabled:
+            # enable http -> https redirect
+            web_config["http"] = {"redirections": {"entryPoint": {"to": "websecure", "scheme": "https"}}},
 
         # TODO Disable static config with telemetry and check new version
         raw_config = {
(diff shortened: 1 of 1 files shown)


src/traefik.py
```diff
@@ -199,14 +199,20 @@ def _update_static_configuration(self):
         tcp_entrypoints = self._tcp_entrypoints
         logger.debug(f"Statically configuring traefik with tcp entrypoints: {tcp_entrypoints}.")
 
+        web_config = {
+            "address": f":{self.port}",
+            # enable by default http -> https redirect
+            "http": {"redirections": {"entryPoint": {"to": "websecure", "scheme": "https"}}},
```
mmkay, line 205: Might be missing something, but should we add a redirection scheme only if there is a certificate? I could imagine a tricky redirect look for a HTTP-only setup if traefik doesn't take into account whether certificate exists.

PietroPasotti: that might explain why the itests are failing though :D

mmkay (approved):

PietroPasotti:
this entails that traefik won't route http requests anymore without upgrading them to https.
@simskij we seem to recall that this change was rejected in the past on the grounds that some applications might actually prefer to be routed over http even if https is available. Can you confirm?

simskij:
> this entails that traefik won't route http requests anymore without upgrading them to https. @simskij we seem to recall that this change was rejected in the past on the grounds that some applications might actually prefer to be routed over http even if https is available. Can you confirm?

I think that was expressed in the context of end-to-end TLS, where we can't necessarily guarantee that everything being routed by traefik supports TLS for the communication between traefik itself and the workload. As for the ingress, upgrading to TLS when configured is in my eyes a reasonable decision.

PR #317, opened 2024-03-26
Use entire chain for the server cert

sed-i:
## Issue
Certificates generated by the charm don't include intermediates: https://github.com/canonical/httprequest-lego-k8s-operator/issues/93.

## Solution
Fetch-lib and use the new `.chain_as_pem` method.

In tandem with:
- https://github.com/canonical/observability-libs/pull/82

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -19,7 +19,7 @@ parts:
       # From PYDEPS
       - "importlib-metadata==6.0.0"
       - "opentelemetry-exporter-otlp-proto-http==1.21.0"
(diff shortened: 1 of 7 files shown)


PietroPasotti:
re the static error: issue is that we can't tell the static checker that in that code block we're working with pydantic v1.
So let's just ignore any error in there.

Abuelodelanada (approved):

PietroPasotti:
Given the specific upgrade instructions, should we take precautions as we publish this to edge, or do we go ahead and merge?

Issue #294, opened 2024-02-16
add config options for providing certificates and cas through config


simskij:
### Enhancement Proposal

Some users provide their own TLS certificates, and don't want it generated through CSRs. To support this we'd need to support the following config options **in addition to the ca relation**:

- `ssl-ca`
- `ssl-cert`
- `ssl-key`

See https://gitea.arif-ali.co.uk/canonical/cpe-deployments/src/branch/main/config/overlays/ssl.yaml for an example of how this works in OpenStack.

mmkay:
Closing as completed by #321 .

PR #321, opened 2024-04-11
Tls config values

dstathis:
## Issue
#294

## Solution
Added config options for TLS.

## Testing Instructions
* Deploy cos-lite
* Set TLS config options in traefik
* Observe that TLS is working properly

(A good way to get working certs is to first relate to self-signed-certificates operator, then use the ones it generates)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -45,15 +45,15 @@ options:
     type: string
     default: path
 
-  ssl-ca:
+  tls-ca:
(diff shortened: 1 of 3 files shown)


lucabello:
Looking good, but I'd rather have another look after CI is green :)

lucabello (approved):

PR #334, opened 2024-04-24
Open ServicePort for entrypoints passed via static traefik_route

mmkay:
## Issue
After #325 we started to add provided `entryPoints` to the static config so that it was possible to route to traefik on custom entrypoints. However, trying to call the same port on juju application ip:${PORT} failed with `Connection refused`. Both in automated and manual tests we verified it by calling the unit's IP, not the external IP (which is later passed in tracing databags).

## Solution
Open ServicePort for entrypoints that are provided via `traefik_route`.

## Testing Instructions
I updated integration tests for `route`, you can run them with:
```
tox -re integration -- -k test_charm_route --keep-models
````
in order to have a working model you can test manually too. Remember though that the last test removes the relation so you'll need to re-relate tester charm with traefik.

### DIFF
--- a/tests/unit/test_route.py
+++ b/tests/unit/test_route.py
@@ -1,6 +1,7 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 """Helpers for unit testing charms which use this library."""
+import os
 from unittest.mock import Mock, patch
 
 import ops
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/unit/test_route.py [hunk 1]
```diff
@@ -280,3 +280,41 @@ def test_static_config_partially_broken(harness: Harness[TraefikIngressCharm]):
...
+
+    charm.traefik_route.on.ready.emit(charm.model.get_relation("traefik-route"))
+
+    # THEN Traefik can list the provided entrypoints
+    tcp_entrypoints = charm._tcp_entrypoints()
+    assert tcp_entrypoints["shondaland"] == "6767"
+
+    # AND open one more service port next to web, websecure
+    service_ports = charm._service_ports
+    assert len(service_ports) == 3
```
PietroPasotti: this is a bit brittle, will adjust a bit

PietroPasotti (approved):

PR #344, opened 2024-04-30
fix issue with cert not present on charm container

PietroPasotti:
## Issue

`server_cert` returns `SERVER_CERT_PATH`, but the cert is only ever pushed to the workload container, never written to the charm container. Consequently charm tracing breaks with tls because the path doesn't exist.

Also, removed a pointless warning log.

## Solution
when you push the cert to the workload container, also write it to the charm container

## Testing Instructions
deploy with tls enabled and watch it burn because other things are broken with tls. But this is definitely an issue too.

### DIFF
--- a/src/traefik.py
+++ b/src/traefik.py
@@ -162,6 +162,8 @@ class Traefik:
     ):
         """Update the server cert, ca, and key configuration files."""
         if cert:
+            # write it to the charm container too, for charm tracing.
+            Path(SERVER_CERT_PATH).write_text(cert)
             self._container.push(SERVER_CERT_PATH, cert, make_dirs=True)
         else:
             self._container.remove_path(SERVER_CERT_PATH, recursive=True)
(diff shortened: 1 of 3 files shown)


PietroPasotti:
TODO @mmkay double-check that the same exact issue doesn't happen in all other charms where we added cert-path for charm tracing + tls

mmkay (approved):

src/traefik.py
```diff
@@ -162,6 +162,8 @@ def update_cert_configuration(
     ):
         """Update the server cert, ca, and key configuration files."""
         if cert:
+            # write it to the charm container too, for charm tracing.
+            Path(SERVER_CERT_PATH).write_text(cert)
```
mmkay, line 166: Looking at itests, this path can be not present in charm container. How about doing `Path.mkdir(DYNAMIC_CONFIG_DIR, parents=True, exist_ok=True)` first?

PietroPasotti: ah yes, good catch

PR #354, opened 2024-05-07
Refactor `_get_loadbalancer_status`

Abuelodelanada:
## Issue

This PR refactors `_get_loadbalancer_status` function. 

## Context

This PR:
- Removes `return ingress_address.hostname` since the hostname (external hostname) is [configured through juju config](https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L1064) so it is not necessary to retrieve that from K8s
- Remove unnecessary indentation.

## Testing Instructions

Run tests

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1141,7 +1141,6 @@ def _get_loadbalancer_status(namespace: str, service_name: str) -> Optional[str]
     return ingress_address.hostname or ingress_address.ip
 
 
-
 def _get_relation_type(relation: Relation) -> _IngressRelationType:
     if relation.name == "ingress":
         return _IngressRelationType.per_app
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -1129,12 +1129,16 @@ def _get_loadbalancer_status(namespace: str, service_name: str) -> Optional[str]
...
+    if not (status := traefik_service.status):  # type: ignore
+        return None
+    if not (load_balancer_status := status.loadBalancer):
+        return None
+    if not (ingress_addresses := load_balancer_status.ingress):
+        return None
+    if not (ingress_address := ingress_addresses[0]):
+        return None
+
+    return ingress_address.ip
```
sed-i, line 1141: I'm guessing that the only thing we'd be losing here is when `ingress_address.hostname` evaluates to something sensible that was not set as a config option. Seems reasonable.

Perhaps just add a comment saying we intentionally disregard it?

Abuelodelanada: done

PietroPasotti (approved):
This looks fine but I'd take the chance to refactor where it's being used as well. That whole codepath is unnecessarily complex

PR #355, opened 2024-05-09
refactor get k8s service ip function

PietroPasotti:
big IMHO for @Abuelodelanada


PietroPasotti:
not going to happen it seems

PR #356, opened 2024-05-10
Create new LoadBalancer svc instead of modifying the Juju ClusterIP svc

IbraAoad:
In tandem with https://github.com/canonical/observability-libs/pull/90

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,3 +1,4 @@
-show-proxied-endpoints:
+get-endpoints:
   description: |
-    Returns a list of endpoints proxied by this Traefik proxy.
+    Returns a list of endpoints, 
+    including both the endpoints proxied by this Traefik proxy and Traefik's own endpoint.
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,4 +1,4 @@
-get-endpoints:
+show-proxied-endpoints:
   description: |
     Returns a list of endpoints, 
     including both the endpoints proxied by this Traefik proxy and Traefik's own endpoint.
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -205,11 +205,13 @@ class KubernetesServicePatch(Object):
         """
         super().__init__(charm, "kubernetes-service-patch")
         self.charm = charm
-        self.service_name = service_name if service_name else self._app
+        self.service_name = service_name or self._app
+        if self.service_name == self._app and service_type == "LoadBalancer":
+            self.service_name= f'{self._app}-lb'
         self.service_type = service_type
         self.service = self._service_object(
             ports,
-            service_name,
+            self.service_name,
             service_type,
             additional_labels,
             additional_selectors,
@@ -222,7 +224,7 @@ class KubernetesServicePatch(Object):
         self.framework.observe(charm.on.install, self._patch)
         self.framework.observe(charm.on.upgrade_charm, self._patch)
         self.framework.observe(charm.on.update_status, self._patch)
-        self.framework.observe(charm.on.remove, self._remove_service)
+        self.framework.observe(charm.on.stop, self._remove_service)
 
         # apply user defined events
         if refresh_event:
@@ -258,8 +260,6 @@ class KubernetesServicePatch(Object):
         Returns:
             Service: A valid representation of a Kubernetes Service with the correct ports.
         """
-        if not service_name:
-            service_name = self._app
         labels = {"app.kubernetes.io/name": self._app}
         if additional_labels:
             labels.update(additional_labels)
@@ -301,12 +301,7 @@ class KubernetesServicePatch(Object):
                 if not self.service_type == "LoadBalancer":
                     self._delete_and_create_service(client)
                 else:
-                    self._delete_and_create_lb_service(client)
-            elif self.service_type == "LoadBalancer":
-                logger.error(
-                    "Failed to create service: cannot create a lb with the same name as the default service."
-                )
-                return
+                    self._create_lb_service(client)
             client.patch(Service, self.service_name, self.service, patch_type=PatchType.MERGE)
         except ApiError as e:
             if e.status.code == 403:
@@ -323,13 +318,9 @@ class KubernetesServicePatch(Object):
         client.delete(Service, self._app, namespace=self._namespace)
         client.create(service)
 
-    def _delete_and_create_lb_service(self, client: Client):
+    def _create_lb_service(self, client: Client):
         try:
-            service = client.get(Service, self.service_name, namespace=self._namespace)
-            service.metadata.name = self.service_name  # type: ignore[attr-defined]
-            service.metadata.resourceVersion = service.metadata.uid = None  # type: ignore[attr-defined]   # noqa: E501
-            client.delete(Service, self.service_name, namespace=self._namespace)
-            client.create(service)
+            client.get(Service, self.service_name, namespace=self._namespace)
         except ApiError:
             client.create(self.service)
 
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -109,6 +109,25 @@ class SomeCharm(CharmBase):
     # ...
 ```
 
+Creating a new k8s service instead of patching the one created by juju
+Providing a service name is mandatory as it shouldn't be the same as default one, i.e., `app_name`.
+```python
+from charms.observability_libs.v1.kubernetes_service_patch import KubernetesServicePatch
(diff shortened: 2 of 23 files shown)


tests/integration/helpers.py
```diff
@@ -5,11 +5,47 @@
...
 from typing import Optional
 
+from lightkube import AsyncClient
+from lightkube.resources.core_v1 import Service
 from pytest_operator.plugin import OpsTest
 
 logger = logging.getLogger(__name__)
 
 
+async def get_k8s_service_address(ops_test: OpsTest, service_name: str) -> Optional[str]:
```
sed-i, line 15: Specifically for traefik, because we already know it must be loadBalancer, wdyt about replacing this func with a one-liner such as:
```python
sh.kubectl(*"-n test-bundle-iwfn get service/trfk -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'".split())
```

src/charm.py
```diff
@@ -1099,7 +1105,9 @@ def _external_host(self) -> Optional[str]:
         if external_hostname := self.model.config.get("external_hostname"):
             return cast(str, external_hostname)
 
-        return _get_loadbalancer_status(namespace=self.model.name, service_name=self.app.name)
+        return _get_loadbalancer_status(
+            namespace=self.model.name, service_name=f"{self.app.name}-lb"
```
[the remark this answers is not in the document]
IbraAoad: this will break the charm and is a bit tricky cause `_get_loadbalancer_status` is called in `traefik_route` prior to initializing `service_patch` and `service_patch` itself depends on `traefik_route`.

actions.yaml
```diff
@@ -1,3 +1,4 @@
-show-proxied-endpoints:
+get-endpoints:
```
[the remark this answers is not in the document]
sed-i: If that's the only reason then it sounds like we should not rename it.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -187,6 +206,7 @@ def __init__(
         super().__init__(charm, "kubernetes-service-patch")
         self.charm = charm
         self.service_name = service_name if service_name else self._app
```
sed-i, line 208: If we already decide on the service name ourselves, might as well append `-lb` to avoid the error below.

```suggestion
        self.service_name = service_name if service_name else self._app if service_type == "LoadBalancer" else f"{self._app}-lb"
```

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -277,7 +298,15 @@ def _patch(self, _) -> None:
             if self._is_patched(client):
                 return
             if self.service_name != self._app:
-                self._delete_and_create_service(client)
+                if not self.service_type == "LoadBalancer":
+                    self._delete_and_create_service(client)
+                else:
+                    self._delete_and_create_lb_service(client)
+            elif self.service_type == "LoadBalancer":
+                logger.error(
```
sed-i, line 306: Should we `raise ValueError` here instead?

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -202,6 +222,7 @@ def __init__(
         self.framework.observe(charm.on.install, self._patch)
         self.framework.observe(charm.on.upgrade_charm, self._patch)
         self.framework.observe(charm.on.update_status, self._patch)
+        self.framework.observe(charm.on.remove, self._remove_service)
```
sed-i, line 225: Iiuc, we're not addressing the scenario of type change across an upgrade. I.e. a charm uses ClusterIP and then switches to LoadBalancer. Or vice versa.

Removing the service on stop would probably address this. Any reason to prefer remove over stop here?

src/charm.py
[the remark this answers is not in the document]
sed-i: I don't follow. I meant something like:
```suggestion
            namespace=self.model.name, service_name=self.service_patch.service_name"
```

src/charm.py
```diff
@@ -1162,7 +1170,10 @@ def server_cert_sans_dns(self) -> List[str]:
 @functools.lru_cache
 def _get_loadbalancer_status(namespace: str, service_name: str) -> Optional[str]:
     client = Client()  # type: ignore
-    traefik_service = client.get(Service, name=service_name, namespace=namespace)
+    try:
+        traefik_service = client.get(Service, name=service_name, namespace=namespace)
+    except ApiError:
+        return None
```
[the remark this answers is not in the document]
sed-i: I don't follow.
Seems like returning None for ApiError is semantically inconsistent with the rest of the function.
Shouldn't the ApiError be bubbled up? I.e. why catch it at all?

lib/charms/observability_libs/v1/kubernetes_service_patch.py
IbraAoad: done but not as condensed just for readability

lib/charms/observability_libs/v1/kubernetes_service_patch.py
IbraAoad: removed altogether in favor of pre-setting the service name

lib/charms/observability_libs/v1/kubernetes_service_patch.py
IbraAoad: yeah this was just an oversight, thought it was the remove hook that comes before upgrade, thanks for noticing!

src/charm.py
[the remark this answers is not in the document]
IbraAoad: So the `service_patch` object includes a param `_service_ports` that requires the `TraefikRouteProvider` and `Traefik` objects to be instantiated prior to `service_patch` and because `TraefikRouteProvider` needs to be aware of `_external_host` which itself calls the `_get_loadbalancer_status` it fails with `AttributeError: 'TraefikIngressCharm' object has no attribute 'service_patch'`

below would work to overcome this but i'm not sure tbh if this actually looks any better than just hard-coding it

```
        try:
            return _get_loadbalancer_status(
                namespace=self.model.name, service_name=self.service_patch.service_name
            )
        except AttributeError:
            return None
```

src/charm.py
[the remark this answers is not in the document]
IbraAoad: so same answer as in here https://github.com/canonical/traefik-k8s-operator/pull/356#discussion_r1612910713 `TraefikRouteProvider` calls this method before the `service_object` is instantiated so service isn't created yet, before this change it used to work cause you get the juju default services prior to any charm events, returning None here indicates that service isn't yet created

actions.yaml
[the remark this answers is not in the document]
simskij: I'm ok with either. I don't think it's **that** weird to keep the name, but also not sure whether I believe we should extend our notion of what a breaking change is to also include charm actions, given that they're not meant to be used programmatically.

Let's keep the name for now.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -186,10 +205,13 @@ def __init__(
         """
         super().__init__(charm, "kubernetes-service-patch")
         self.charm = charm
-        self.service_name = service_name if service_name else self._app
+        self.service_name = service_name or self._app
+        if self.service_name == self._app and service_type == "LoadBalancer":
```
sed-i, line 209: Add comment? E.g.
```suggestion
        # The LB service cannot have the same name as the app, because...
        if self.service_name == self._app and service_type == "LoadBalancer":
```

lib/charms/observability_libs/v1/kubernetes_service_patch.py
sed-i, line 1: Is this in sync with the tandem PR?

sed-i (commented):
Few last bits.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
IbraAoad: one commit was missing, should be in sync now

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -321,6 +351,26 @@ def _is_patched(self, client: Client) -> bool:
...
+        Raises:
+            ApiError: If there is an issue with the Kubernetes API request.
+        """
+        client = Client()  # pyright: ignore
+
+        try:
+            client.get(Service, self.service_name, namespace=self._namespace)
+            client.delete(Service, self.service_name, namespace=self._namespace)
+        except ApiError:
+            return
```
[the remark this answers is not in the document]
IbraAoad: Yes it's broader, it should be updated now to only `pass` if not found `status.code 404`

lucabello (approved):

PR #371, opened 2024-06-14
adapted charm and tracing test code ready for new tracing and charm_tracing libs

PietroPasotti:
Currently the CI is bork because the new tracing lib has a new tracing protocol interface version.
Also charm_tracing has some changes we want to try out.

This will need to wait for tempo to release the new charm_tracing and tracing libraries.

https://github.com/canonical/tempo-k8s-operator/pull/135
https://github.com/canonical/tempo-k8s-operator/pull/136

Once that is done, we can fetch_lib and this branch should work.


mmkay:
We now have the new libs and we also use `charm_tracing_config` so this can be closed.

PR #373, opened 2024-06-17
tests: wait 1m before retry to reduce flakiness

natalian98:
## Issue

The forward-auth integration [tests](https://github.com/canonical/traefik-k8s-operator/blob/6f2895fffbcb3cf1a92523c3034dbf88eef01081/tests/integration/test_forward_auth.py#L83) are flaky due to latency in mounting the configmap as a volume inside the container.

## Solution

The latency depends on kubelet's `syncFrequency` config (see more reference [here](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/)). The default is 1m, so to reduce the tests flakiness we can wait 1m before retrying a request.

## Context

n/a

## Testing Instructions

n/a

## Upgrade Notes

tests: wait 1m before retry to reduce flakiness


PietroPasotti:
@natalian98 we'll delete this branch as it seems to be stale, or do you intend to drive it to completion?

natalian98:
Yes, you can delete this branch. Thanks

PR #387, opened 2024-07-25
Use 5m update-status hook interval for the auth test

sed-i:
## Issue
forward-auth itest fails about 50% of the time.
It relies on event deferral, and by extension on the update-status interval.

## Solution
Override the "inf" interval we set in our autouse fixture just for this test.
h/t @natalian98


natalian98 (approved):

natalian98:
fixes https://github.com/canonical/traefik-k8s-operator/pull/387

PietroPasotti (approved):

Issue #382, opened 2024-07-12
pydantic deprecation warnings


dimaqq:
### Bug Description

validators generate these warnings when tests are run

### To Reproduce

- add the ingress lib
- run pytest / tox -e unit

### Environment

any

### Relevant log output

```shell
lib/charms/traefik_k8s/v2/ingress.py:255
  /code/hexanator/lib/charms/traefik_k8s/v2/ingress.py:255: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
    @validator("scheme", pre=True)

lib/charms/traefik_k8s/v2/ingress.py:262
  /code/hexanator/lib/charms/traefik_k8s/v2/ingress.py:262: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
    @validator("port", pre=True)

lib/charms/traefik_k8s/v2/ingress.py:280
  /code/hexanator/lib/charms/traefik_k8s/v2/ingress.py:280: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
    @validator("host", pre=True)

lib/charms/traefik_k8s/v2/ingress.py:286
  /code/hexanator/lib/charms/traefik_k8s/v2/ingress.py:286: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
    @validator("ip", pre=True)
```

lucabello:
@PietroPasotti

lucabello:
[...]

PR #388, opened 2024-07-30
Fix Pydantic warnings.

pedro-avalos:
Pydantic has deprecated validator and should be replaced by field_validator.

See https://docs.pydantic.dev/2.8/migration/#validator-and-root_validator-are-deprecated

Let me know if you need anything else from me.

## Issue

This PR addresses #382.

## Solution

As documented by https://docs.pydantic.dev/2.8/migration/#validator-and-root_validator-are-deprecated, `validator` is now deprecated and should be replaced with `field_validator`.

`pre=True` is no longer a parameter, but `mode="before"` should behave in the same way if I properly understood the documentation.

## Context

- https://docs.pydantic.dev/2.8/migration/#validator-and-root_validator-are-deprecated
- https://docs.pydantic.dev/latest/concepts/validators/#before-after-wrap-and-plain-validators

## Testing Instructions

- add the ingress lib
- run pytest / tox -e unit

## Upgrade Notes

N/A

### DIFF
--- a/lib/charms/traefik_k8s/v2/ingress.py
+++ b/lib/charms/traefik_k8s/v2/ingress.py
@@ -62,7 +62,7 @@ import pydantic
 from ops.charm import CharmBase, RelationBrokenEvent, RelationEvent
 from ops.framework import EventSource, Object, ObjectEvents, StoredState
 from ops.model import ModelError, Relation, Unit
-from pydantic import AnyHttpUrl, BaseModel, Field, validator
+from pydantic import AnyHttpUrl, BaseModel, Field, field_validator
 
 # The unique Charmhub library identifier, never change it
 LIBID = "e6de2a5cd5b34422a204668f3b8f90d2"
@@ -252,14 +252,14 @@ class IngressRequirerAppData(DatabagModel):
         default="http", description="What scheme to use in the generated ingress url"
     )
 
-    @validator("scheme", pre=True)
+    @field_validator("scheme", mode="before")
     def validate_scheme(cls, scheme):  # noqa: N805  # pydantic wants 'cls' as first arg
         """Validate scheme arg."""
         if scheme not in {"http", "https", "h2c"}:
             raise ValueError("invalid scheme: should be one of `http|https|h2c`")
         return scheme
 
-    @validator("port", pre=True)
+    @field_validator("port", mode="before")
     def validate_port(cls, port):  # noqa: N805  # pydantic wants 'cls' as first arg
         """Validate port."""
         assert isinstance(port, int), type(port)
@@ -277,13 +277,13 @@ class IngressRequirerUnitData(DatabagModel):
         "IP can only be None if the IP information can't be retrieved from juju.",
     )
 
-    @validator("host", pre=True)
+    @field_validator("host", mode="before")
     def validate_host(cls, host):  # noqa: N805  # pydantic wants 'cls' as first arg
         """Validate host."""
         assert isinstance(host, str), type(host)
         return host
 
-    @validator("ip", pre=True)
+    @field_validator("ip", mode="before")
     def validate_ip(cls, ip):  # noqa: N805  # pydantic wants 'cls' as first arg
         """Validate ip."""
         if ip is None:


PietroPasotti:
Hi! Thanks a lot for the contribution. There is one important complication that we have to keep in mind.
Traefik's ingress libraries are used by charms that use pydantic v1 as well as charms that use v2. So we have to make sure all changes are backwards compatible with pydantic v1 (TODO for us: set up CI to verify that it is!).

If field_validator is not available in V1 as I suspect, you should do a little bit of trickery in the existing 
`if PYDANTIC_IS_V1:` block to define a uniform validator for both versions.

PietroPasotti (changes requested):
Let's make sure we're backwards compatible with pydantic v1.

pedro-avalos:
Hi @PietroPasotti, thank you for your feedback. Would something along these lines be what you requested? There are some parameters that are no longer parameters in pydantic's `field_validator` of v2. I noticed that they weren't used anyways in `ingress.py` anyways, so that *shouldn't* be an issue. Let me know what you think.

```py
    from pydantic import validator

    def field_validator(
        field: str,
        /,
        *fields: str,
        mode: str,
        check_fields: Optional[bool] = None,
    ) -> Callable[[Any], Any]:
    """Wraps the pydantic v1 validator decorator with a v2-compatible interface.

    See https://docs.pydantic.dev/latest/migration/#changes-to-config.
    """
        # Maps mode to value for `pre`
        # See https://docs.pydantic.dev/latest/concepts/validators/#before-after-wrap-and-plain-validators
        modes = {
            "after": False,
            "before": True,
            "plain": True,
            "wrap": False,  # I am honestly not sure what this one should map to...
        }

        pre = modes.get(mode)
        if pre is None:
            raise ValueError(f"Unrecognized mode '{mode}'")

        # The following arguments are no longer arguments for field_validator:
        # each_item, always, whole, and allow_reuse
        return validator(
            field,
            *fields,
            pre=pre,
            check_fields=check_fields,
        )
```

PietroPasotti:
> Hi @PietroPasotti, thank you for your feedback. Would something along these lines be what you requested? There are some parameters that are no longer parameters in pydantic's `field_validator` of v2. I noticed that they weren't used in `ingress.py` anyways, so that _shouldn't_ be an issue. Let me know what you think.

I would start by doing things in the simplest possible way:

```python
if PYDANTIC_IS_V1:
    # right now we don't need any arguments that are incompatible between v1's validator and v2's field_validator, so we can use them transparently.
    from pydantic import validator as field_validator
else:
    from pydantic import field_validator
```

or will this not work?

pedro-avalos:
That will not work. The parameter that each of them take is completely different.

This is v2's `field_validator` signature:
```py
field_validator(
    field: str,
    /,
    *fields: str,
    mode: FieldValidatorModes = "after",
    check_fields: bool | None = None,
) -> Callable[[Any], Any]
```

And this is v1's `validator` signature:
```py
def validator(
    *fields: str,
    pre: bool = False,
    each_item: bool = False,
    always: bool = False,
    check_fields: bool = True,
    whole: Optional[bool] = None,
    allow_reuse: bool = False,
) -> Callable[[AnyCallable], 'AnyClassMethod']:
```

If the rest of the file only used the `fields` part of the validator, then your proposed solution would work. However, the `pre=True` parameter that was used is incompatible with the v2 way of doing things (`mode='before'`).

PietroPasotti:
Thanks for the clarification.
So it seems the concrete problem now is the 'pre' param. That maps to the `mode="before"` arg in v2. 

So what I'd do is

```python
if PYDANTIC_IS_V1:
    from pydantic import validator
    input_validator = partial(validator, pre=True)
else:
    from pydantic import field_validator
    input_validator = partial(field_validator, mode="before")
```

which addresses the issues we have right now?

pedro-avalos:
Yes that would work. I'll make those changes.

pedro-avalos:
I pushed the fix using `partial`, as suggested. I noticed that there was another part of the code that could be updated for pydantic v2. These **should** be all the changes within this file (and they address the warnings specifically mentioned by #382). However, there are other warnings that may be coming from other files, so I can look into those and play some whack-a-mole with the deprecation warnings if that should be part of this PR.

pedro-avalos:
I'm not sure why the integration test is failing, it seems to be failing across the board for all the PRs and branches, so I assume it doesn't have to do with my PR.

@PietroPasotti, should I edit other files to completely get rid of the deprecation warnings? I have found it in 2 other places (`lib/charms/tempo_k8s/v2/tracing.py:869` and `src/charm.py:933`, the latter seems to be the one specifically triggering the last deprecation warning when running `tox -e unit`)

https://github.com/pedro-avalos/traefik-k8s-operator/blob/2fe1698ec697c77a41a7a9b5ff521ddb09c22d83/lib/charms/tempo_k8s/v2/tracing.py#L868

https://github.com/pedro-avalos/traefik-k8s-operator/blob/2fe1698ec697c77a41a7a9b5ff521ddb09c22d83/src/charm.py#L933

PietroPasotti (changes requested):
Hi! 
This looks good :) 
We're missing two things for this to be accepted from my side:
1 - you need to bump the LIBPATCH in both `tracing` and `ingress`, so that once we merge this the CI will release the new revisions to charmhub.
2 - not a blocker, but it would be great if you added a test to verify that the warnings are gone! a simple one using `pytest`'s `caplog` fixture should do the trick. I think given the fact that the warnings happen on import, it might be a little tricky to trigger them inside the scope of a test, so perhaps you'll need a fixture to 'unimport' the packages beforehand? Let me know if you need help with that, I did something similar just a few days ago.

PietroPasotti (approved):
Thank you!

PR #404, opened 2024-09-18
Fix `traefik_route` library ID

michaeldmitry:
## Issue
Previous PR tried to publish a new library `traefik_k8s.v0.traefik_route`, but using an ID for an existing library `traefik_route_k8s.v0.traefik_route`.

## Solution
Retrieve unique ID for this new lib using `charmcraft create-lib traefik_route`

## Context
https://github.com/canonical/traefik-k8s-operator/pull/403


PietroPasotti (approved):
you'll also need to reset libpatch to 0 or 1 as it's effectively a new library

michaeldmitry:
> you'll also need to reset libpatch to 0 or 1 as it's effectively a new library

I did that last PR! But it seems CI has overwritten it after the merge (auto lib update?)
Good catch!

PR #407, opened 2024-09-27
Implement V1 of certificate_transfer

saltiyazan:
## Issue
The new version (V1) of the certificate_transfer_interface is incompatible of its V0 version. Applications that implement V1 of cert transfer won't be able to share their CAs with Traefik.

## Solution

This change is to allow application the implement the V1 version of the interface to integrate with Traefik to provide their CAs that Traefik needs to trust.
**Note**: The solution might seem a bit hacky, I tried to follow the same solution with `IngressPerAppV0` and `IngressPerAppV1`.

## Testing Instructions
- Deploy Traefik from this branch
- Deploy `notary-k8s` from [this](https://github.com/canonical/notary-k8s-operator/pull/31) branch
- Deploy self-signed-certificates
- Integrate Notary and self-signed-certificates using the `certificates` interface
- Integrate Notary and Traefik using the `receive-ca-cert-v1` interface
- Observe the CA file being created in the correct directory in the Traefik container
- Remove the integration and observe the file being removed


saltiyazan:
Hi, @sed-i @simskij 
Could someone please take a look at this? happy to re-work it if needed.

metadata.yaml
```diff
@@ -100,6 +100,14 @@ requires:
...
       This relation can be used with a local CA to obtain the CA cert that was used to sign proxied
       endpoints.
+      This is meant for applications that use the certificate-transfer-interface v0.
+  receive-ca-cert-v1:
+    interface: certificate_transfer
+    description: |
+      Receive a CA certs for traefik to trust.
+      This relation can be used with a local CA to obtain the CA certs that was used to sign proxied
+      endpoints.
+      This is meant for applications that use the certificate-transfer-interface v1.
```
ca-scribner, line 110: My uninformed first impression is that I dislike adding a new relation to handle a different version of the same interface.  This feels like a change that should be transparent to the user if at all possible.  Ideally, the existing `receive-ca-cert` relation should just work for both v0 and v1.  Is that possible?

dstathis: If the interface is the same, we shouldn't use a new relation. The promise is supposed to be that any provider of an interface works with any consumer.

saltiyazan: V0 and V1 are not compatible. There was breaking change in the interface too not only the library.
Here you can see examples of the data expected in [V0](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/certificate_transfer/v0/README.md#example) and the data expected in [V1](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/certificate_transfer/v1/README.md#example).

Isn't it the same with Ingress V1 and V2?

IbraAoad: It's the same for ingress but handled differently not to require a new relation, traefik [only has one relation for ingress](https://github.com/canonical/traefik-k8s-operator/blob/main/metadata.yaml#L46) and we do our checks internally within the charm code to [validate if client is talking v0/v1](https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L890)

saltiyazan: I see your point thank you.
I'll update the PR.

saltiyazan: I updated the PR to follow the example of Ingress.
Could you please take a look @IbraAoad ?

src/charm.py
```diff
@@ -474,21 +502,31 @@ def _update_received_ca_certs(self, event: Optional[CertificateTransferAvailable
         preferred.
         """
         cas = []
-        if event:
+        if event and isinstance(event, CertificateTransferAvailableEventV0):
```
ca-scribner, line 505: This is a minor suggestion, but I like refactoring this function instead to something like:

```python
        cas = []
        if event and isinstance(event, CertificateTransferAvailableEventV0):
            cas.append(CA(event.ca, uid=event.relation_id))
        else:
            for relation in self.model.relations.get(_RECV_CA_CERT_RELATION_NAME, []):
                recv_ca_cert_requirer = self._recv_ca_cert_requirer_from_relation(relation)
                relation_certificates = []
                if recv_ca_cert_requirer is self.recv_ca_cert_v0:
                    # For some reason, relation.units includes our unit and app. Need to exclude them.
                    for unit in set(relation.units).difference([self.app, self.unit]):
                        # Note: this nested loop handles the case of multi-unit CA, each unit providing
                        # a different ca cert, but that is not currently supported by the lib itself.
                        if ca := relation.data[unit].get("ca"):
                            relation_certificates.append(ca)
                elif recv_ca_cert_requirer is self.recv_ca_cert_v1:
                    relation_certificates.extend(self.recv_ca_cert_v1.get_all_certificates(relation.id))
                else:
                    # todo: handle this error as mentioned in the other comment
                    pass
                cas.extend(CA(ca, uid=f"{relation.id}-{i}") for i, ca in enumerate(relation_certificates))
        self.traefik.add_cas(cas)
```

The only reason is that it feels a bit odd that traefik treats the v0 and v1 with different naming schemes.  If I understand correctly, the difference is that v0 conveyed always one CA and v1 can convey N CAs?  If so, I'd name anything from v0 as if its just the 0th element of the list.  If that's ok, then the code gets slightly less duplicated and feels more readable imo

src/traefik.py
```diff
@@ -217,8 +217,9 @@ def remove_cas(self, uids: Iterable[Union[str, int]]):
         Calls update-ca-certificates once done.
         """
         for uid in uids:
-            ca_path = RECV_CA_TEMPLATE.substitute(rel_id=str(uid))
-            self._container.remove_path(ca_path)
+            files = self._container.list_files(CA_CERTS_PATH, pattern=f"receive-ca-cert-{uid}-*")
```
ca-scribner, line 220: I think files can be:
1. receive-ca-cert-UID
2. receive-ca-cert-UID-I

The pattern cannot have the last "-" or it'll miss format (1)

```suggestion
            files = self._container.list_files(CA_CERTS_PATH, pattern=f"receive-ca-cert-{uid}*")
```

Note that elsewhere I suggest we should only have one file name pattern rather than two, but either way this suggestion here would work

ca-scribner: (just noting - it isn't this PRs fault) I think there must not be test coverage for this deletion atm?

ca-scribner (commented):
I think this looks close.  A few suggestions and minor bugs commented

I don't have a great sense of how this charm is supposed to function so I might be missing something, but afaict the direction of this pr lgtm

src/charm.py
saltiyazan: Thanks, that's a good point.
Refactored.

src/traefik.py
saltiyazan: Suggestion applied anyway.

saltiyazan:
The failing test `test_charm_trace_collection` seems to be failing on other open prs and on main (at least locally for me).
I haven't looked into it that much, but I'm happy to do it if you want me to in the context of this PR.

IbraAoad:
#442 fixed the broken tests, updating this to main to see if CI will pass

IbraAoad (approved):

saltiyazan:
> #442 fixed the broken tests, updating this to main to see if CI will pass

Hi @IbraAoad Thanks for the CI fix :)
Looks like it passes here too now.

PR #414, opened 2024-11-06
Change alert rule to be for 5 minutes

dstathis:
## Issue
The TraefikIngressUnitIsUnavailable is very unreliable and triggers constantly on COS installs. It should wait for 5 minutes before firing the alert.

## Solution
Add `for: 5m` to the alert rule.

## Testing Instructions
None really

## Upgrade Notes
Traefik alert rules are now more stable.

### DIFF
--- a/src/prometheus_alert_rules/unit_unavailable.rule
+++ b/src/prometheus_alert_rules/unit_unavailable.rule
@@ -1,6 +1,6 @@
 alert: TraefikIngressUnitIsUnavailable
 expr: up < 1
-for: 0m
+for: 5m
 labels:
   severity: critical
 annotations:


src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -1,6 +1,6 @@
 alert: TraefikIngressUnitIsUnavailable
 expr: up < 1
-for: 0m
+for: 5m
```
sed-i, line 3: @dstathis is this to address the case of first install, or do you see these fire all the time?

dstathis: All the time. A few times a day. I investigated what might be wrong but I could not find any issues.

PR #428, opened 2024-11-27
Add a charm config to add custom annotations to loadbalancers created by traefik

### LINKED ISSUE #423: add a charm config to add custom annotations to loadbalancers created by traefik
### Enhancement Proposal

In the use case of deploying the COS stack with multiple traefiks for finer control over routing between the various components of COS, it will be super helpful if we could add custom annotations to loadbalancers.

IbraAoad:
## Issue
Fixes #423 

## Solution
1- Remove the [`kubernetes_service_patch`](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v1/kubernetes_service_patch.py) dependency from traefik and implement an lb reconciler using [`kubernetes_resource_manager`](https://github.com/canonical/lightkube-extensions/blob/main/src/lightkube_extensions/batch/_kubernetes_resource_manager.py)
2- Expose a juju config named `loadbalancer_annotations` that will allow users to provide custom annotations.

Annotations validation is based on [Kubernetes' annotation validation](https://github.com/kubernetes/apimachinery/blob/v0.31.3/pkg/api/validation/objectmeta.go#L44) and [Kubernetes' syntax and character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set)

## Context
Cloud providers use annotations on load balancers to control their behavior. This change enables support for custom annotations on Traefik's load balancers

## Testing Instructions
1 - Juju deploy traefik-k8s
2 - juju config traefik-k8s loadbalancer_annotations="key1=value1"
3 - kubectl describe traefik lb service and validate annotations


config.yaml
```diff
@@ -8,6 +8,19 @@ options:
...
+      A comma-separated list of annotations to apply to the LoadBalancer service.
+      The format should be: `key1=value1,key2=value2,key3=value3`.
+      These annotations are passed directly to the Kubernetes LoadBalancer service, 
+      enabling customization for specific cloud provider settings or integrations.
+      
+      Example:
+        "external-dns.alpha.kubernetes.io/hostname=example.com,service.beta.kubernetes.io/aws-load-balancer-type=nlb"
+
+      Ensure the annotations are correctly formatted and adhere to Kubernetes' syntax and character set : https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+      invalid values will result in the annotations being reset to None, removed from the LoadBalancer, and all previously set annotations will be lost.
```
ca-scribner, line 22: I worry about the failure case here.  Annotations used on these loadbalancers are sometimes security controls (for example, to make an LB internal-facing only or provide certs).  If someone does this:

```
juju config traefik loadbalancer_annotations="important.security.setting=x,something@malformed"
```

The charm will configure the loadbalancer without the `important.security.setting`.  Yes the charm will complain, but until someone notices and can rectify the issue, the security setting is missing.  

Ideally, I think this should fail in a way that is fail safe, such as removing the LoadBalancer if possible

src/charm.py
```diff
@@ -94,6 +94,34 @@
 
 PYDANTIC_IS_V1 = int(pydantic.version.VERSION.split(".")[0]) < 2
 
+# Regex for Kubernetes annotation values:
```
ca-scribner, line 97: Do you have links where these come from?  If yes, that's a good thing to add in the comments

src/charm.py
```diff
@@ -94,6 +94,34 @@
...
+# - Allows alphanumeric characters, dots (.), dashes (-), and underscores (_)
+# - Matches the entire string
+# - Does not allow empty strings
+# - Example valid: "value1", "my-value", "value.name", "value_name"
+# - Example invalid: "value@", "value#", "value space"
+ANNOTATION_VALUE_PATTERN = re.compile(r"^[\w.\-_]+$")
+
+
```
ca-scribner, line 105: nit: some of these are double spaced, some single

src/charm.py
```diff
@@ -340,6 +370,79 @@ def _basic_auth_user(self) -> Optional[str]:
...
+                "Expected format: key1=value1,key2=value2."
+            )
+            return None
+
+        # Validate each key-value pair
+        for key, value in parsed_annotations.items():
+            if not self._validate_annotation_key(key) or not self._validate_annotation_value(
+                value
+            ):
+                return None
```
ca-scribner, line 429: extra context: my comment in `config.yaml` about whether to fail open or closed

imo this failure behaviour is too quiet.  Misconfiguring the LoadBalancer feels like something that should surface as a charm status.  Almost certainly if the user has provided annotations and we discard them, the LoadBalancer is not operating as they'd expect from a charm that says it is healthy.

`_validate_annotation_*()` do log errors (`logger.error(msg)`), but that is easily missed.

ca-scribner: Probably the reason we're discussing this though is because, like in #420, the current architecture of this charm makes handling an error that should block the whole charm really hard.  Ideally imo, this sort of misconfiguration would sit as a BlockedStatus until it was resolved (although maybe some parts of the charm would still function separately)

Given that we cannot easily achieve that atm, I think the LoadBalancer needs to break more obviously.  Maybe if we cannot validate the annotations, we should change the LoadBalancer service_type back to ClusterIP?

src/charm.py
```diff
@@ -340,6 +370,79 @@ def _basic_auth_user(self) -> Optional[str]:
...
+
+    def _validate_annotation_value(self, value: str) -> bool:
+        """Validate the annotation value."""
+        if not ANNOTATION_VALUE_PATTERN.match(value):
+            logger.error(
+                f"Invalid annotation value: '{value}'. Must follow Kubernetes annotation syntax."
+            )
+            return False
+
+        return True
```
ca-scribner, line 399: nit: these are static and could be helpers outside the class

tests/unit/test_charm.py
```diff
@@ -598,3 +598,52 @@ def test_when_external_hostname_is_invalid_go_into_blocked_status(self):
                     self.harness.charm.unit.status, BlockedStatus, invalid_hostname
                 )
                 self.assertEqual(requirer.urls, {})
+
+    @patch("charm._get_loadbalancer_status", lambda **_: "10.0.0.1")
+    @patch("charm.KubernetesServicePatch", lambda **_: None)
+    def test_lb_annotations(self):
+        test_cases = [
+            ("key1=value1,key2=value2", {"key1": "value1", "key2": "value2"}),
+            ("key1=value1,key2", None),
```
ca-scribner, line 607: should this be down with the other failure cases?

ca-scribner (commented):
This PR to me is correctly implemented, but personally I think the failure mode designed here should be different (see comments for more details)

src/k8s_lb_manager.py
```diff
@@ -0,0 +1,237 @@
...
+                logger.info(f"Patched LoadBalancer {self.name} in namespace {self.namespace}")
+            else:
+                logger.debug(f"No changes for LoadBalancer {self.name}")
+        except ApiError as e:
+            # Create the service if it doesn't exist
+            if e.status.code == 404:
+                self.client.create(service)
+                logger.info(f"Created LoadBalancer {self.name} in namespace {self.namespace}")
+            else:
+                logger.info(f"Failed to create LoadBalancer {self.name}: {e}")
```
ca-scribner, line 127: What about using the [kubernetes resource manager](https://github.com/canonical/lightkube-extensions/blob/main/src/lightkube_extensions/batch/_kubernetes_resource_manager.py) here?  That way you get reconcile and deletion logic for free

src/k8s_lb_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        else:
+            self.remove_lb()
+
+    def remove_lb(self):
+        """Removes the LoadBalancer."""
+        try:
+            self.client.delete(Service, name=self.name, namespace=self.namespace)
+            logger.info(f"Deleted LoadBalancer {self.name} in namespace {self.namespace}")
+        except ApiError as e:
+            logger.info(f"Failed to delete LoadBalancer {self.name}: {e}")
```
ca-scribner, line 92: What about using the [kubernetes resource manager](https://github.com/canonical/lightkube-extensions/blob/main/src/lightkube_extensions/batch/_kubernetes_resource_manager.py) here?  That way you get reconcile and deletion logic for free.

src/charm.py
```diff
@@ -238,24 +209,16 @@ def __init__(self, *args):
...
-                ipa_v2.on.data_removed,  # type: ignore
-                self.ingress_per_unit.on.data_provided,  # type: ignore
-                self.ingress_per_unit.on.data_removed,  # type: ignore
-                self.traefik_route.on.ready,  # type: ignore
-                self.traefik_route.on.data_removed,  # type: ignore
-                self.on.traefik_pebble_ready,  # type: ignore
-                self.on.config_changed,  # type: ignore
-            ],
+            additional_labels={"app.kubernetes.io/name": self.app.name},
+            additional_selectors={"app.kubernetes.io/name": self.app.name},
```
ca-scribner, line 219: Now that it isn't a charm library that needs to observe its own charm events, should this be in `__init__`?  I bet it fits more naturally somewhere else

Putting this somewhere else will also make a lot of the mocking in tests unnecessary as we no longer make lightkube calls on any charm invocation

src/k8s_lb_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        self.additional_selectors = additional_selectors or {}
+        self.additional_annotations = additional_annotations or {}
+        self.additional_annotations = parse_annotations(additional_annotations)
+        # Initialize Kubernetes client
+        self.client = Client(namespace=self.namespace, field_manager=self.field_manager)
+
+        if self._annotations_valid():
+            self.reconcile()
+        else:
+            self.remove_lb()
```
ca-scribner, line 84: what about removing these from `__init__` and just expecting the user to call `lb.reconcile()`?  It'll make testing and mocking a lot easier, and feels a bit more natural imo.  Its true that then doing `x = KubernetesLoadBalancer(...)` doesn't actually produce the loadbalancer, but I don't think that's a huge negative.

ca-scribner, line 84: in particular, `self.remove_lb()` feels like an unexpected thing to call in a constructor

src/k8s_lb_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        self.ports = ports
+        self.field_manager = field_manager
+        self.additional_labels = additional_labels or {}
+        self.additional_selectors = additional_selectors or {}
+        self.additional_annotations = additional_annotations or {}
+        self.additional_annotations = parse_annotations(additional_annotations)
+        # Initialize Kubernetes client
+        self.client = Client(namespace=self.namespace, field_manager=self.field_manager)
+
+        if self._annotations_valid():
```
ca-scribner, line 81: Could we just put this "if valid, reconcile; else, remove" logic into `reconcile`?  I think that feels more natural.  Then `reconcile` could also provide a `force` or a `delete_if_annotations_are_invalid` to control whether we respect that check.

src/k8s_lb_manager.py
```diff
@@ -0,0 +1,237 @@
...
+                type="LoadBalancer",
+            ),
+        )
+
+        try:
+            # Check if the service exists
+            existing_service = self.client.get(Service, name=self.name, namespace=self.namespace)
+
+            # Patch if differences exist
+            if not self._is_patched(old_service=existing_service, new_service=service):
```
ca-scribner, line 116: Kubernetes is idempotent, so if we do `client.patch()` when the service is unchanged, nothing will happen.  Why don't we remove this `_is_patched` code entirely and just always patch?

ca-scribner (commented):
Focused mainly on the kubernetes load balancer class.  Still need to do a full pass later

ca-scribner (approved):
A few minor things otherwise lgtm

src/charm.py
```diff
@@ -543,6 +652,7 @@ def _config_hash(self) -> int:
             (
                 self._external_host,
                 self.config["routing_mode"],
+                self.config.get("loadbalancer_annotations", None),
```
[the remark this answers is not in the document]
IbraAoad: not really, traefik uses the hash function to compare if `old configs == new` and then call `_process_status_and_configurations()` which is the only entrypoint to reconciling the lb, I changed this into a more aggressive call-on-every-hook cause apparently `_process_status_and_configurations()`  isn't really called on every hook as I thought :cry:

PR #431, opened 2024-12-04
Remove deprecated call to ops.main

sed-i:
## Issue
Charm logs have the following for every hook:
```
./src/charm.py:1255: DeprecationWarning: Calling `ops.main.main()` is deprecated, call `ops.main()` instead
  main(TraefikIngressCharm, use_juju_for_storage=True)
```

## Solution
Use `ops.main` instead of `ops.main.main`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -65,7 +65,7 @@ from ops.charm import (
     UpdateStatusEvent,
 )
 from ops.framework import StoredState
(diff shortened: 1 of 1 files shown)


lucabello:
Done in https://github.com/canonical/traefik-k8s-operator/pull/428

Issue #347, opened 2024-05-02
Traefik's server_cert_sans_dns has unexpected contents


PietroPasotti:
### Bug Description

TLDR: "It works on my virtual machine"

Traefik issues a CSR for "`Creating CSR for darkstar.kpn with DNS ['darkstar.kpn'] and IPs [] `" because `Traefik.server_cert_sans_dns` is returning `['darkstar.kpn']` i.e. 

Supposing the app IP is  "10.64.140.43"
```python
import socket
socket.gethostbyaddr("10.64.140.43")   
('darkstar.kpn', [], ['10.64.140.43'])     
```

Note that `darkstar` is the name of my machine.

In a VM, you get

```python
>>> socket.gethostbyaddr("10.233.117.142")
('10-233-117-142.kubernetes.default.svc.cluster.local', [], ['10.233.117.142'])
```

Traefik doesn't expect 'darkstar.kpn' and therefore we have a mismatch between the SANs in the certs issued to traefik by self-signed-certificates.

Note: in another case, (the VM's network was bridged to the host with e.g. `--network enp1s0`) we got:
```python
>>> socket.gethostbyaddr("192.168.1.250")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
socket.herror: [Errno 1] Unknown host
```

![image](https://github.com/canonical/traefik-k8s-operator/assets/6230162/43e7bb5b-be7a-4314-8c55-f81f513be946)

### To Reproduce

n/a

### Environment

n/a

### Relevant log output

```shell
n/a
```

mmkay:
We could try setting traefik's external hostname to `darkstar.kpn` to see if that helps. If this is the issue, we could try to document it or detect this scenario and change the logic around it.

PR #446, opened 2025-02-07
Add the IP address of the resolved hostname to the certificate SANS

michaeldmitry:
## Issue
Potentially fix the issue where TLS communication to Traefik fails in environments where the Loadbalancer IP can be resolved back to a hostname on the machine running the workloads.

Potentially fix https://github.com/canonical/traefik-k8s-operator/issues/347
Fixes #461
@PietroPasotti to confirm.

## Solution
If an IP address can be resolved back to a hostname, return both the hostname as well as the IP address to be used in the certificate SANS. The reason is that charms related to Traefik over ingress will communicate with Traefik via this IP address and thus it should be contained in the server certificate SANS.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -3,6 +3,7 @@
 # See LICENSE file for licensing details.
 
 """Charmed traefik operator."""
+
 import contextlib
 import enum
 import itertools
@@ -197,7 +198,9 @@ class TraefikIngressCharm(CharmBase):
         self.ingress_per_unit = IngressPerUnitProvider(charm=self)
 
         self.traefik_route = TraefikRouteProvider(
-            charm=self, external_host=self._external_host, scheme=self._scheme  # type: ignore
+            charm=self,
+            external_host=self._external_host,  # type: ignore
+            scheme=self._scheme,  # type: ignore
         )
 
         self._topology = JujuTopology.from_charm(self)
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -1285,7 +1288,9 @@ def server_cert_sans_dns(self) -> List[str]:
             name, _, _ = socket.gethostbyaddr(target)  # type: ignore
             # Do not return "hostname" like '10-43-8-149.kubernetes.default.svc.cluster.local'
             if is_hostname(name) and not name.endswith(".svc.cluster.local"):
-                return [name]
+                # In case we can do a DNS lookup on that IP address,
+                #  return the resolved hostname as well as the IP address to be both included in the certificate SANS.
+                return [name, target] if target else [name]
```
dstathis, line 1293: I think it is rather strange that we are doing a reverse DNS lookup here. I think if hostname is not set we should just use the IP address. In other words the cert should match the address in the "Serving at \<address\>" status message.

So I think the best course of action is to just delete this block altogether.

dstathis (changes requested):

src/charm.py
dstathis: On second thought, it is probably just safer to keep this.

dstathis (approved):

PR #450, opened 2025-02-12
Traefik appends a `tls` route even when `tls` routes are explicitly provided

### LINKED ISSUE #430: Traefik appends a `tls` route even when `tls` routes are explicitly provided
Summarizing my understanding of what's going on here... please correct any mistakes

For every route, the previous code created a duplicate route `ROUTENAME-tls` for the same `rule` and `service` but with additional TLS configuration.  Both applied to the same `rule` and thus conflict, and the new `ROUTENAME-tls` route takes precedence over the original `ROUTENAME` because its name is longer.  Prior to canonical/traefik-k8s-operator#425 this worked fine because we wanted to use `ROUTENAME-tls` in all cases.

The updated code here avoids making the second `-tls` route in the specific case where the route includes `tls: passthrough: true`.

Have we really solved the broader problem here, or narrowly fixed canonical/traefik-k8s-operator#425?  Two questions come to mind:
1) are there other TLS configurations that someone might pass to traefik via `submit_to_traefik`?  If so, the same issue as canonical/traefik-k8s-operator#425 will occur
2) Should the charm be creating these duplicate routes that mask the original routes in the first place?  That might be out of scope for the fix here, but we should at least document it as an issue if it is actually an issue.

_Originally posted by @ca-scribner in https://github.com/canonical/traefik-k8s-operator/pull/429#discussion_r1863903261_

IbraAoad:
## Issue
Fixes #430

## Context
For every route, `traefik` created a duplicate route `ROUTENAME-tls` for the same rule and service but with additional TLS configuration. Both applied to the same rule and thus conflict, and the new `ROUTENAME-tls` route takes precedence over the original `ROUTENAME` because its name is longer. 

## Solution
- **New Parameter:** Introduce a new boolean argument `raw` in the interface.
- **Conditional Behavior:**  
  - When `raw` is set to `True`, Traefik will **only** append TLS routes to HTTP routes.
  - TLS routes for non-HTTP protocols will not be generated, preventing the duplicate `ROUTENAME-tls` routes.

## Testing
**Tandem PR**: https://github.com/canonical/grafana-k8s-operator/pull/391

### DIFF
--- a/lib/charms/traefik_k8s/v0/traefik_route.py
+++ b/lib/charms/traefik_k8s/v0/traefik_route.py
@@ -66,7 +66,7 @@ class TraefikRouteCharm(CharmBase):
     traefik_route = TraefikRouteRequirer(
         self, self.model.relations.get("traefik-route"),
         "traefik-route",
-        raw_config=False  # Default: Traefik will append TLS configs
+        raw=False  # Default: Traefik will append TLS configs
     )
     if traefik_route.is_ready():
         traefik_route.submit_to_traefik(
@@ -86,7 +86,7 @@ class TraefikRouteCharm(CharmBase):
     traefik_route = TraefikRouteRequirer(
         self, self.model.relations.get("traefik-route"),
         "traefik-route",
-        raw_config=True  # Traefik will not modify TLS settings on non HTTP routes
+        raw=True  # Traefik will not modify TLS settings on non HTTP routes
     )
     if self.traefik_route.is_ready():
         self.traefik_route.submit_to_traefik(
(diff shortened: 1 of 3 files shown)


dstathis (approved):
Love all the red in this diff!

mmkay:
Are we expecting any charms to break due to 

> and rely on charm authors to provide the full traefik config when relating, including any TLS configs.

IbraAoad:
> Are we expecting any charms to break due to
> 
> > and rely on charm authors to provide the full traefik config when relating, including any TLS configs.
> 
> ?

Unfortunately yes, it seems Tempo is relying to Traefik upgrading the connection, We'll have to keep the functionality and introduce a backward compatible way to allow charm authors to opt out of getting automated TLS routes and at the same time not breaking charms that depend on it.

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -274,29 +312,50 @@ def get_static_config(self, relation: Relation) -> Optional[str]:
...
+        self._raw = raw
+
+        if self._raw:
+            log.warning(
+                "WARNING: 'raw' flag is set.\n"
+                "This will prevent Traefik from generating TLS routes for non-HTTP protocols,\n"
+                "even if a TLS provider is configured.\n"
+                "Charm authors should only use this option if they fully understand the implications\n"
+                "and intentionally do not require the appended TLS routes."
+            )
```
dstathis, line 358: I would not include this warning. It will only show up if raw was intentionally set, so it will not really ever result in action.

dstathis (commented):
LGTM

dstathis (approved):

Issue #262, opened 2023-10-04
Traefik forward traffic to backends which are down


gnuoy:
### Bug Description

Traefik does not seem to do an aliveness check on the backends it is forwarding traffic to. This causes client requests to fail if a backend is down.

Perhaps charms.traefik_k8s.v2.ingress should support the requirer passing a health check url ?

### To Reproduce

1) Deploy this bundle: https://opendev.org/openstack/charm-keystone-k8s/src/branch/main/tests/bundles/smoke.yaml
2) Add a keystone unit: juju add-unit keystone
3) Wait for unit to be ready
4) URL=$(juju run  keystone/leader get-admin-account | awk 'BEGIN {FS="="} /OS_AUTH_URL/ {print $NF}')
5) curl $URL   (repeat multiple times to check both backends are alive as traefik will round-robin the backends)
6) juju ssh --container keystone keystone/1 "pebble stop wsgi-keystone"
7) Repeat step 5 and every other request will be a bad gateway

Example output: https://paste.ubuntu.com/p/zHrmjFrQ7g/

### Environment

juju 3.2.3-genericlinux-amd64
Controller in microk8s
Traefick charm: 1.0/candidate  r148

### Relevant log output

```shell
2023-10-04T09:04:43.426Z [traefik] time="2023-10-04T09:04:43Z" level=debug msg="'502 Bad Gateway' caused by: dial tcp 10.1.188.231:5000: connect: connection refused"
2023-10-04T09:04:44.446Z [traefik] time="2023-10-04T09:04:44Z" level=debug msg="'502 Bad Gateway' caused by: dial tcp 10.1.188.231:5000: connect: connection refused"
```

PietroPasotti:
could definitely wrap around https://doc.traefik.io/traefik/routing/services/#health-check
we'll discuss prioritization in the next backlog refinement

sed-i:
Path to a health check seems like a reasonable addition to the reldata schema.

gboutry:
@PietroPasotti We are still hitting this issue, is there any news?

PietroPasotti:
@dstathis @IbraAoad has this been prioritised?

sed-i:
Perhaps this means adding health check info to the relation schema, so we could render "healthcheck" etnries in the services section of the traefik config, e.g.

```yaml
        healthCheck:
          path: /status
          interval: 10s
          timeout: 3s
```

https://doc.traefik.io/traefik/routing/services/#health-check_1

[...]

PR #460, opened 2025-02-28
Add support for healthcheck

MylesJP:
## Issue
This commit adds support for an optional dictionary 'healthcheck_params' containing healthcheck path, interval, and timeout to be passed via an ingress relation. This allows Traefik to detect down nodes and remove them from the loadbalancer rotation. If 'healthcheck_params' is provided, the 'path' field is required from the charm while 'interval' and 'timeout' can be omitted and will be set to Traefik default values.

## Solution
To implement healthcheck, an optional dictionary is passed via the ingress relation and populates in the Traefik dynamic configuration file found at `/opt/traefik/juju/`. This piece of configuration requires `path` according to the [documentation for healthcheck](https://doc.traefik.io/traefik/routing/services/#health-check) while other fields are optional or have default values within Traefik.

## Testing Instructions
Unit and integration tests are a WIP and will be added once the reviewing team is in agreement on the solution. PR is left in draft state until then. This solution has been tested locally using multiple nodes of keystone-k8s and can be confirmed to detect nodes in a down state. To view raw API data from Traefik, `insecure: True` was temporarily added to the static configuration file as [described in the documentation](https://doc.traefik.io/traefik/operations/api/#insecure).

[#LP 2077269](https://bugs.launchpad.net/snap-openstack/+bug/2077269)
Related to: #262


lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -663,6 +666,7 @@ def __init__(
         host: Optional[str] = None,
         ip: Optional[str] = None,
         port: Optional[int] = None,
+        healthcheck_params: Optional[dict] = None,
```
hemanthnakkina, line 669: Add validation on keys allowed in healthcheck_params

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -252,6 +252,9 @@ class IngressRequirerAppData(DatabagModel):
     model: str = Field(description="The model the application is in.")
     name: str = Field(description="the name of the app requesting ingress.")
     port: int = Field(description="The port the app wishes to be exposed.")
+    healthcheck_params: Optional[dict] = Field(
```
hemanthnakkina, line 255: Update LIBPATCH version in this file

hemanthnakkina (changes requested):

PR #500, opened 2025-07-08
Use persistent hash

### LINKED ISSUE #499: Incorrect config hash logic
question (unrelated to this PR): does this actually work? AFAIK `hash` uses a pseudorandom salt that is calculated when the interpreter starts, since the charm code runs every time an event is fired, I would expect that the `hash` would change on every single run even if the config didn't change (because the salt would be different). If you want the hash to be the same across different runs, you should use `hashlib`. I hope I am not missing something. 

e.g.
image

_Originally posted by @nsklikas in https://github.com/canonical/traefik-k8s-operator/pull/496#discussion_r2185374556_

nsklikas:
## Issue
Using the python builtin `hash` function is not consistent across interpreter
runs, as it relies on a pseudo-random salt that is calculated that differs
on every run.

See https://github.com/canonical/traefik-k8s-operator/issues/499

## Solution
Use `hashlib` instead

Closes #499


Thanhphan1147 (approved):
LGTM, thanks!

nsklikas:
@Thanhphan1147 there is a bug we discovered in one of our charms where if the pod gets deleted, the charm storage will continue to exist (because it is on a volume) BUT the config will not be there (because it is not on a volume). This caused the charm to try to start the workload with invalid config (see relevant [issue](https://github.com/canonical/hydra-operator/issues/422)) . There reason is that the storedState is an emptyDir volume and (from the [emptyDir docs](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir)):

> A container crashing does not remove a Pod from a node. The data in an emptyDir volume is safe across container crashes.

So for our charms we decided to ditch the hash approach and we just compare the files

DeeKay3 (approved):

PR #550, opened 2025-09-24
ISD-4275 add mypy to lint checks 

Thanhphan1147:
As part of the handover of traefik to @canonical/platform-engineering, I'd like to add `mypy` as a lint step in the relevant tox environment:

* Add `mypy` as an optional dependency
* Run `mypy` checks against the `src/` directory for now, we'll cover libraries owned by traefik in `lib/` in a follow-up PR
* Fix any issues that comes with adding `mypy`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -359,7 +359,7 @@ class TraefikIngressCharm(CharmBase):
         observe(self.on.show_external_endpoints_action, self._on_show_external_endpoints)  # type: ignore
 
         # Hook hollistic method
-        observe(self.on.traefik_pebble_ready, self.cleanup_tls_configuration)  # type: ignore
+        observe(self.on.traefik_pebble_ready, self.cleanup_tls_configuration)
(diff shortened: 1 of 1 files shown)


dstathis (approved):

src/charm.py
```diff
@@ -359,7 +359,7 @@ def __init__(self, *args):
         observe(self.on.show_external_endpoints_action, self._on_show_external_endpoints)  # type: ignore
 
         # Hook hollistic method
-        observe(self.on.traefik_pebble_ready, self.cleanup_tls_configuration)  # type: ignore
+        observe(self.on.traefik_pebble_ready, self.cleanup_tls_configuration)
         observe(self.on.start, self.cleanup_tls_configuration)
         observe(self.on.update_status, self.cleanup_tls_configuration)
         observe(self.on.config_changed, self.cleanup_tls_configuration)
```
arturo-seijas, line 365: Do you need to run this on update status too? Also, isn't start always preceded by config changed?

PR #551, opened 2025-09-25
Isd 4275 fix lint in traefik libs

Thanhphan1147:
Follow-up of #550.  As part of the handover of traefik to @canonical/platform-engineering, I'd like to add mypy as a lint step in the relevant tox environment, This PR fixes the `mypy` issues identified in the libraries owned by the charm.


Thanhphan1147:
In draft until #550 is merged

PR #557, opened 2025-09-26
fix: restart pebble in the correct order

### LINKED ISSUE #556: Traefik restarts Pebble service before the new static config is pushed to the container
### Bug Description

TLDR: It seems like there is a race condition (kind of?) in Traefik where on `certs_changed` (i.e. when a relation over `certificates` is joined, such as relating to `self-signed certificates`) we restart the Traefik Pebble service before the new static config file is pushed to the container. Hence, for the changes to the static config to take effect, a manual restart becomes necessary. 

# Description of Bug
This issue builds on top of another related [issue](https://github.com/canonical/traefik-k8s-operator/issues/503).
Assume you have Traefik related to another charm (e.g. Grafana) and that Grafana is related to Traefik over `traefik_route` and Grafana is related to self-signed certificates over `certificates`. You would need to now relate SSC to Traefik. When you do this, the following happens:

1. First, the `_on_cert_changed` method is called inside charm.py:

https://github.com/canonical/traefik-k8s-operator/blob/9ddaa8cba1bebb93a9cbb9832cd32176cf5db68c/src/charm.py#L325-L327

2. That method called the `_configure` method in [charm.py](https://github.com/canonical/traefik-k8s-operator/blob/9ddaa8cba1bebb93a9cbb9832cd32176cf5db68c/src/charm.py#L645C5-L647C26)

3. The `_configure` method does a few things. Most importantly as far this issue is concerned is that it calls `_update_cert_configs` **before** calling `_configure_traefik`.
https://github.com/canonical/traefik-k8s-operator/blob/9ddaa8cba1bebb93a9cbb9832cd32176cf5db68c/src/charm.py#L879-L886

4. `_update_cert_configs` makes calls to `traefik.py` and eventually restarts the Pebble service.
5. `_configure_traefik` recreates the static config file and pushes it to the container.
6. The effect of this order is that when on step 4 the restart happens, it starts the service running the old config file. Thus, if now you call `http://<Traefik-IP>/<model-name>-<related_app_to_traefik>`, you don't get a 200 (which is the expected behaviour of adding the `certificates` relation). A manual restart becomes necessary for the effect of the added relation to kick in.

It seems like the the order of the calls needs to be reversed, such that we configure first (i.e. we push the config to the container first) and then we restart the service.

### To Reproduce

Please use the bundle below:
```
bundle: kubernetes
applications:
  graf:
    charm: grafana-k8s
    channel: 2/edge
    revision: 162
    resources:
      grafana-image: 71
      litestream-image: 46
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  ssc:
    charm: self-signed-certificates
    channel: 1/edge
    revision: 360
    scale: 1
    constraints: arch=amd64
  trfk:
    charm: traefik-k8s
    channel: latest/edge
    revision: 246
    base: ubuntu@20.04/stable
    resources:
      traefik-image: 162
    scale: 1
    constraints: arch=amd64
    storage:
    trust: true
relations:
- - graf:ingress
  - trfk:traefik-route
- - ssc:certificates
  - graf:certificates
```
Once you have deployed the bundle, then (and only then) do `juju relate ssc:certificates trfk`.

Now, call http://Traefik-IP/model-name-graf. You should see a 404. 
Perform, `juju ssh -c traefik trfk/0 pebble restart traefik` and carry out the HTTP call again. This time, you'll get a 200 because the restart occurred with the new static config file (i.e. with the redirectHTTPS block).

In other words, the race condition caused a restart of Pebble before the new static config file could be pushed to the container and that's why we needed a manual restart for it to take effect.

### Environment

Running inside Multipass VM. Using Juju 3.6.10 and Ck8s `1.33-classic/stable`. For the charm versions, please refer to the bundle above.

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

sinapah:
## Issue

Fixes #556 

This PR fixes the issue by removing the logic to restart Pebble from inside `traefik.py` to inside `charm.py` when needed and in the correct order. Hence, `charm.py` executes the logic to restart Pebble which was previously handled by `traefik.py`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -883,8 +883,8 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         if not self.container.can_connect():
             return
 
-        self._configure_traefik()
         self._update_cert_configs()
+        self._configure_traefik()
         self._restart_traefik()
         self._process_status_and_configurations()
 
@@ -901,8 +901,9 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
             if self._is_tls_enabled():
                 # we keep this nested under the hash-check because, unless the tls config has
                 # changed, we don't need to redo this.
-                self._configure_traefik()
                 self._update_cert_configs()
+                self._configure_traefik()
+
                 self._restart_traefik()
 
             self._process_status_and_configurations()
--- a/tests/integration/test_pebble_restart_after_cert_relation_joined.py
+++ b/tests/integration/test_pebble_restart_after_cert_relation_joined.py
@@ -82,4 +82,7 @@ async def test_can_route_ingress_using_tls(ops_test: OpsTest):
     # Ensure we are able to get a 200 when calling AM. The helper asserts the status code.
     fetch_with_retry(alertmanager_address)
 
-
+    alertmanager_address_http = alertmanager_address.replace("https://", "http://")
+    # Ensure we are able to get a 200 when calling AM. The helper asserts the status code.
+    # This should also work with HTTP.
+    fetch_with_retry(alertmanager_address_http)
--- a/tests/integration/test_upstream_ingress.py
+++ b/tests/integration/test_upstream_ingress.py
@@ -201,7 +201,6 @@ async def test_ipa_ingressed_through_upstream_ingress_with_tls(ops_test: OpsTest
         200,
     )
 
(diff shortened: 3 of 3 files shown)


Thanhphan1147 (approved):
LGTM, thanks

Issue #612, opened 2026-03-10
Traefik unit took long time to settle


hemanthnakkina:
### Bug Description

This is a sunbeam deployment with 2 traefiks - traefik-public and traefik. Both integrated with 15 applications (each having 3 units) over ingress/traefik-route. When traefik gets certificates from manual-tls-certificates operator, the traefik took around 30-45 minutes to settle.

Relevant snap-openstack bug: https://bugs.launchpad.net/snap-openstack/+bug/2129152

juju status for traefik-public
```
$ juju status -m openstack traefik-public --relations
Model Controller Cloud/Region Version SLA Timestamp
openstack sqa-dh1-j6-controller sqa-dh1-j6-k8s/localhost 3.6.14 unsupported 11:57:12Z

SAAS Status Store URL
cinder-volume active local admin/openstack-machines.cinder-volume
grafana-dashboards active local admin/observability.grafana-dashboards
loki-logging active local admin/observability.loki-logging
microceph active local admin/openstack-machines.microceph
prometheus-receive-remote-write active local admin/observability.prometheus-receive-remote-write

App Version Status Scale Charm Channel Rev Address Exposed Message
traefik-public 2.11.0 active 3 traefik-k8s latest/stable 272 10.152.183.234 no Serving at https://10.242.10.10/

Unit Workload Agent Address Ports Message
traefik-public/0* active idle 10.1.2.240 Serving at https://10.242.10.10/
traefik-public/1 active idle 10.1.0.7 Serving at https://10.242.10.10/
traefik-public/2 active idle 10.1.1.102 Serving at https://10.242.10.10/

Offer Application Charm Rev Connected Endpoint Interface Role
ceilometer ceilometer ceilometer-k8s 155 1/1 ceilometer-service ceilometer provider
cert-distributor keystone keystone-k8s 332 2/2 send-ca-cert certificate_transfer provider
certificate-authority certificate-authority self-signed-certificates 586 1/1 certificates tls-certificates provider
cinder-volume-mysql-router cinder-volume-mysql-router mysql-router-k8s 871 1/1 database mysql_client provider
keystone-credentials keystone keystone-k8s 332 2/2 identity-credentials keystone-credentials provider
keystone-endpoints keystone keystone-k8s 332 1/1 identity-service keystone provider
keystone-ops keystone keystone-k8s 332 0/0 identity-ops keystone-resources provider
nova nova nova-k8s 217 1/1 nova-service nova provider
[...]

PR #649, opened 2026-04-15
Optimize traefik code

swetha1654:
Applicable spec: <link>

### Overview

1. Merged dynamic config configuration files from multiple relations into a single config file to avoid several container pushes. 
2. Introduced caching of properties that i don't expect to change over a hook lifecycle
3. Removed call to `update-ca-certificates`. We are now using `rootCAs` option provided by traefik instead.
4. Removed `reconcile-lb` where it doesn't make sense. 
5. Removed `restart_traefik` wherever its already done by the code later.

### Checklist

- [ ] The [charm style guide](https://documentation.ubuntu.com/juju/3.6/reference/charm/charm-development-best-practices/) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The documentation is updated
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`, `documentation`)
- [ ] The `docs/changelog.md` is updated with user-relevant changes.
- [ ] The `LIBAPI` and `LIBPATCH` values have been incremented for charm libraries owned by this project (if they were modified).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -517,7 +517,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         """
         return cast(Optional[str], self.config.get("basic_auth_user", None))
 
-    @property
+    @functools.cached_property
     def _loadbalancer_annotations(self) -> Optional[Dict[str, str]]:
         """Parses and returns annotations to apply to the LoadBalancer service.
 
@@ -573,7 +573,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
             resources_list.append(self._construct_lb())
         klm.reconcile(resources_list)
 
-    @property
+    @functools.cached_property
     def _get_loadbalancer_status(self) -> Optional[str]:
         try:
             traefik_service = self.lightkube_client.get(
@@ -594,7 +594,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
 
         return ingress_address.hostname or ingress_address.ip
 
-    @property
+    @functools.cached_property
     def _traefik_loadbalancer_ip(self) -> Optional[str]:
         try:
             traefik_service = self.lightkube_client.get(
@@ -649,7 +649,8 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         if not self.container.can_connect():
             return
         self._update_received_ca_certs(event)
-        # We need to restart Traefik now
+        # Regenerate static config so rootCAs picks up the new CA, then restart.
+        self.traefik.configure()
         self._restart_traefik()
 
     def _update_received_ca_certs(
@@ -681,7 +682,8 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
     def _on_recv_ca_cert_removed(self, event: CertificateTransferRemovedEvent) -> None:
         # Assuming only one cert per relation (this is in line with the original lib design).
         self.traefik.remove_ca(str(event.relation_id))
-        # Since remove_ca will call update_ca_certs in traefik, a restart is needed.
+        # Regenerate static config so rootCAs reflects the removed CA, then restart.
+        self.traefik.configure()
         self._restart_traefik()
 
     def _is_tls_enabled(self) -> bool:
@@ -1098,10 +1100,13 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
     def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
-        self._clear_all_configs_and_restart_traefik()
+        # Wipe stale dynamic configs that may have survived on the storage volume.
+        self.traefik.delete_dynamic_config()
+        # Restore CA certs from relations before configuring, so that
+        # generate_static_config() picks them up in rootCAs.
+        self._update_received_ca_certs()
         # push the (fresh new) configs.
         self._configure()
-        self._update_received_ca_certs()
         self._set_workload_version()
 
     def _clear_all_configs_and_restart_traefik(self) -> None:
@@ -1742,7 +1747,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
             "ip": self._traefik_loadbalancer_ip,
         }
 
-    @property
+    @functools.cached_property
     def _traefik_external_address(self) -> Optional[str]:
         """Return the address used to ingress directly through this Traefik's gateway.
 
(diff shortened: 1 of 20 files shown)


src/traefik.py
```diff
@@ -291,15 +297,39 @@ def remove_ca(self, uid: str) -> None:
...
+        hash_before = self._ca_certs_hash()
         try:
             self._container.remove_path(ca_path)
         except PathError:
             logger.exception("Error removing cert file %s.", ca_path)
-        self.update_ca_certs()
+        self.update_ca_certs(hash_before=hash_before)
+
+    def _ca_certs_hash(self) -> str:
+        """Compute a hash of all CA cert files in the CA certs directory."""
```
javierdelapuente: This does not look to be correct, as the hash is computed with metadata not with the files.

src/traefik.py
```diff
@@ -291,15 +297,39 @@ def remove_ca(self, uid: str) -> None:
...
+        hasher = hashlib.md5()  # noqa: S324
+        try:
+            for f in sorted(self._container.list_files(CA_CERTS_DIR), key=lambda x: x.path):
+                hasher.update(f.path.encode())
+                hasher.update(str(f.size).encode())
+                hasher.update(str(f.last_modified).encode())
+        except Exception:  # noqa: BLE001
```
javierdelapuente: Why this? Could we catch the real exceptions?

src/traefik.py
```diff
@@ -725,29 +755,48 @@ def restart(self) -> None:
...
+    def flush_dynamic_configs(self) -> None:
+        """Merge all buffered configs and push a single file to the container."""
+        merged: Dict[str, Any] = {}
+        for config in self._dynamic_configs.values():
+            for protocol in ("http", "tcp", "udp"):
+                if protocol not in config:
+                    continue
+                merged.setdefault(protocol, {})
+                for section, entries in config[protocol].items():
+                    if isinstance(entries, dict):
```
javierdelapuente: Out of curiosity, if it is not a dict, what can it be (and contain)?

src/traefik.py
```diff
@@ -291,15 +297,39 @@ def remove_ca(self, uid: str) -> None:
...
         ca_path = RECV_CA_TEMPLATE.substitute(rel_id=uid)
+        hash_before = self._ca_certs_hash()
         try:
             self._container.remove_path(ca_path)
         except PathError:
             logger.exception("Error removing cert file %s.", ca_path)
-        self.update_ca_certs()
+        self.update_ca_certs(hash_before=hash_before)
+
+    def _ca_certs_hash(self) -> str:
```
gregory-schiano: Certificates already have hashes, why not getting each cert hash in a set and use that set for comparison, it'll avoid `hexdigest`
Not sure which one will be more efficient as reading the hash in the CA file will require reading the file with certificate manipulation libraries, so it might be less efficient

src/traefik.py
```diff
@@ -291,15 +297,39 @@ def remove_ca(self, uid: str) -> None:
...
+        try:
+            for f in sorted(self._container.list_files(CA_CERTS_DIR), key=lambda x: x.path):
+                hasher.update(f.path.encode())
+                hasher.update(str(f.size).encode())
+                hasher.update(str(f.last_modified).encode())
+        except Exception:  # noqa: BLE001
+            pass
+        return hasher.hexdigest()
+
+    def update_ca_certs(self, hash_before: Optional[str] = None) -> None:
```
gregory-schiano: I checked and Traefik 3 has a `rootCAs` directive, not sure if Traefik 2 has a similar one. Because we're doing that only to add the Root CA as a trusted CA for Traefik and I would be surprised if there's no way to inject extra CA directly in traefik bypassing the system CA which requires `update-ca-certificates` command

src/charm.py
```diff
@@ -6,7 +6,9 @@
 """Charmed traefik operator."""
 
 import contextlib
+import cProfile
```
seb4stien, line 9: It should be removed for the production charm I guess. Or is there a way to disable it to not impact performances?

tests/unit/test_charm.py
```diff
@@ -99,6 +99,22 @@ def _render_middlewares(*, strip_prefix: bool = False, redirect_https: bool = Fa
     )
 
 
+# In production, ops creates a fresh charm instance per hook, so cached_property works correctly.
+# Harness reuses the same instance across events, requiring manual cache invalidation.
+_CACHED_PROPERTIES = (
+    "_loadbalancer_annotations",
```
javierdelapuente, line 105: do we use cached_property? I do not see the change in the code.

javierdelapuente (commented):
A few comments, but looks good!

src/traefik.py
swetha1654: removed this logic in favour of `rootCA` now

src/traefik.py
swetha1654: v2.11 also has the rootCA directive. I've used that now

swetha1654: I'm only wondering if there would be any side effects around TCP servers. The static `serversTransport.rootCAs` only applies to the default@internal HTTP serversTransport.

src/charm.py
swetha1654: Yes i will take care of removing this before my PR is ready for review.

src/traefik.py
swetha1654: removed this logic altogether in favour of rootCAs

src/traefik.py
swetha1654: removed this logic altogether in favour of rootCAs

tests/unit/test_charm.py
swetha1654: pushed the changes now

src/traefik.py
swetha1654: it can contain lists, for eg TLS certificates, in which case i extend the certificates list. Updated the merging logic.

src/charm.py
```diff
@@ -1069,28 +1067,30 @@ def _udp_entrypoints(self) -> Dict[str, Any]:
...
-        self.traefik.configure()
-
     def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
+        if not self.container.can_connect():
+            return
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
-        self._clear_all_configs_and_restart_traefik()
+        # Wipe stale dynamic configs that may have survived on the storage volume.
+        self.traefik.delete_dynamic_config()
```
hemanthnakkina, line 1076: This will delete and recreate in-memory config in traefik workload which may create problems, see https://github.com/canonical/traefik-k8s-operator/issues/665

swetha1654:
All the changes in this PR have been applied over a series of 5 PRs mentioned in #612

Issue #647, opened 2026-04-13
traefik-k8s validation of the loadbalancer_annotations config incorrectly returns an error if the value is a url


mostafaCamel:
### Bug Description

Sometimes, I need to pass specific annotations to the laodbalancer, especially when k8s is integrated with openstack.
One of these annotations is `loadbalancer.openstack.org/default-tls-container-ref` (see [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#service-annotations)) whose value is a url

It looks like the traefik k8s config incorrectly erros when one of the values is a url (I guess the `:` and `/` characters are problematic in the charm validation code)

### Impact

Medium (functionality degraded, workaround exists)

### Impact Rationale

I can always omit this value from the config then run `kubectl patch` to update directly in the cluster

### To Reproduce

Deploy a traefik-k8s charm and put a dummy loadbalancer annotation whose value is a url

### Environment

traefik-k8s latest/stable   revision 270  

juju --version
3.6.2-genericlinux-amd64

traefik in cos running on top of canonicalk8s (k8s charm) . The k8s nodes and the juju controllers are openstack VMs.

k8s integrated with opesntack via OCCM (deployed via helm) and not the openstack-integrator charm. So the loadbalancers are created via Octavia in openstack (see [reference](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md))

### Relevant log output

```shell
juju config -m cos traefik loadbalancer_annotations="loadbalancer.openstack.org/port-id=0f21631e-4f41-4a9f-bc0e-fa8d2a3a8eac,loadbalancer.openstack.org/flavor-id=9685d152-fb1c-4a7b-9913-79a47d59ac8f,loadbalancer.openstack.org/default-tls-container-ref=https://brbcn.iapi.rsos.hexacloud.sa:9312/v1/secrets/17890c19-3adf-4592-90eb-b7328f4ba761"

juju debug-log --include traefik/leader -m cos
[...]
unit-traefik-0: 02:05:08 ERROR unit.traefik/0.juju-log Invalid annotation value: 'https://brbcn.iapi.rsos.hexacloud.sa:9312/v1/secrets/17890c19-3adf-4592-90eb-b7328f4ba761'. Must follow Kubernetes annotation syntax
```

[...]

PR #653, opened 2026-04-17
Remove unnecessary annotation value validation

mostafaCamel:
Applicable spec: <link>

### Overview

Remove unnnecessary validation on k8s annotation values.

k8s does not do do any validation on the content of the annotation value . See `for k := range annotations {` (which means the validation inside this block is only on the key and not on the value ) in the validation package ([here](https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L46) and [here](https://github.com/kubernetes/apimachinery/blob/72791e98891abe6955b7bcc709f9904c8434c8b0/pkg/api/validation/objectmeta.go#L46))

Also, in the k8s documentation, there is an [example](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) of a url as an annotation value , which is not allowed in the pattern currently in src/charm.py

### Rationale

Being able to pass annotation values that are pefectly fine from a k8s perspective

### Juju Events Changes

None

### Module Changes

None

### Library Changes

None

### Checklist

- [ x ] The [charm style guide](https://documentation.ubuntu.com/juju/3.6/reference/charm/charm-development-best-practices/) was applied
- [ x ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ x ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ x ] The documentation is updated
- [ x ] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`, `documentation`)
- [ x ] The `docs/changelog.md` is updated with user-relevant changes.
- [ x ] The `LIBAPI` and `LIBPATCH` values have been incremented for charm libraries owned by this project (if they were modified).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -105,15 +105,6 @@ logger = logging.getLogger(__name__)
 
 _TRAEFIK_CONTAINER_NAME = "traefik"
 
-
-# Regex for Kubernetes annotation values:
-# - Allows alphanumeric characters, dots (.), dashes (-), and underscores (_)
-# - Matches the entire string
-# - Does not allow empty strings
-# - Example valid: "value1", "my-value", "value.name", "value_name"
-# - Example invalid: "value@", "value#", "value space"
-ANNOTATION_VALUE_PATTERN = re.compile(r"^[\w.\-_]+$")
-
 # Based on https://github.com/kubernetes/apimachinery/blob/v0.31.3/pkg/util/validation/validation.go#L204  # noqa  # pylint: disable=line-too-long
 # Regex for DNS1123 subdomains:
 # - Starts with a lowercase letter or number ([a-z0-9])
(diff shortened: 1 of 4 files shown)


mostafaCamel:
https://github.com/canonical/traefik-k8s-operator/issues/647

src/charm.py
swetha1654, line 1: It seems like there's another issue with the annotation validation too. It seems like we're checking if the key <= 253 chars but we should we checking the prefix only. [[reference]](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set:~:text=If%20specified%2C%20the%20prefix%20must%20be%20a%20DNS%20subdomain%3A%20a%20series%20of%20DNS%20labels%20separated%20by%20dots%20(.)%2C%20not%20longer%20than%20253%20characters%20in%20total%2C%20followed%20by%20a%20slash%20(/).)

Would you be able to change that too?

swetha1654 (approved):
Thank you so much for your contribution. One small comment from my side, otherwise all good :)

src/charm.py
mostafaCamel: Created a new issue https://github.com/canonical/traefik-k8s-operator/issues/659

PR #658, opened 2026-04-21
Fix missing wait

yhaliaw:
Applicable spec: <link>

### Overview

Add `wait()` to a container `exec()`.

### Rationale

The `wait()` call waits for the config file to be deleted before the next `update_tls_configuration` pushes a config file to the same path.

There can be a race condition between the delete and push file if they are executed without waiting for one to complete first.

### Checklist

- [x] The [charm style guide](https://documentation.ubuntu.com/juju/3.6/reference/charm/charm-development-best-practices/) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`, `documentation`)
- [ ] The `docs/changelog.md` is updated with user-relevant changes.
- [ ] The `LIBAPI` and `LIBPATCH` values have been incremented for charm libraries owned by this project (if they were modified).

### DIFF
--- a/src/traefik.py
+++ b/src/traefik.py
@@ -728,7 +728,7 @@ class Traefik:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     def delete_dynamic_configs(self) -> None:
         """Delete **ALL** yamls from the dynamic config dir."""
         # instead of multiple calls to self._container.remove_path(), delete all files in a swoop
-        self._container.exec(["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"])
+        self._container.exec(["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"]).wait()
         logger.debug("Deleted all dynamic configuration files.")
 
     def delete_dynamic_config(self, file_name: str) -> None:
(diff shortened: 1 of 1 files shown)


src/traefik.py
```diff
@@ -728,7 +728,7 @@ def restart(self) -> None:
     def delete_dynamic_configs(self) -> None:
         """Delete **ALL** yamls from the dynamic config dir."""
         # instead of multiple calls to self._container.remove_path(), delete all files in a swoop
-        self._container.exec(["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"])
+        self._container.exec(["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"]).wait()
```
swetha1654: would it be better to specify wait with a timeout? We don't want it to hang indefinitely.

swetha1654 (approved):
Out of curiosity, were you able to simulate the race condition? Could i see the logs if you did?

yhaliaw:
> Out of curiosity, were you able to simulate the race condition? Could i see the logs if you did?

I did not got a run that has this issue. Just thought it is technical possible.

src/traefik.py
yhaliaw: Added a timeout.

btw, the `self._container.exec(["update-ca-certificates", "--fresh"]).wait()` does not have a timeout in the code now. I am not sure what would be a reasonable timeout value for that one.

swetha1654: I wouldn't worry about it since i am planning to raise a PR to remove that altogether and replace it with a config option in traefik to read a specific path containing CAs.

PR #675, opened 2026-05-08
fix: Rewrite the logic to push dynamic configuration files (3 of 5)

swetha1654:
#### What this PR does

- Rewrote the logic in `src/traefik.py` to accumulate all per-relation dynamic configs in memory and push them together as a tar archive, extracting them in the container and ensuring stale configs are cleaned up before new ones are written. This replaces the previous approach of pushing each file individually. (`[src/traefik.pyR741-R820](diffhunk://#diff-2f7766a4273c9c0294d8b46a7436e65af8f8666147e88b9f52a12426fb967479R741-R820)`)
- Updated the usage in `src/charm.py` to use the new batching mechanism: configs are now added via `add_dynamic_config` and flushed with `flush_dynamic_configs`.

#### Why we need it

Optimization for #612 

#### Checklist


src/charm.py
```diff
@@ -715,10 +715,6 @@ def _update_cert_configs(self) -> None:
         self._sync_certs_to_peer_databag()
         self.traefik.update_cert_configuration(self._get_certs())
 
-        # update_cert_configuration relies on traefik.update_ca_certs.
-        # Thus, we should restart traefik with the new CA certs.
-        self._restart_traefik()
```
swetha1654, line 718: every function that calls _update_cert_configs already calls restart_traefik, this reduces the number of restarts

src/charm.py
```diff
@@ -1068,7 +1064,8 @@ def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
         if not self.container.can_connect():
             logger.debug("Container is unreachable")
             return
-        self._clear_all_configs_and_restart_traefik()
+        # Wipe stale dynamic configs that may have survived on the storage volume.
+        self.traefik.delete_dynamic_configs()
```
swetha1654, line 1066: `_clear_all_configs_and_restart_traefik` basically calls:
```
L1: self.traefik.delete_dynamic_configs()
L2: self._update_cert_configs()
L3: self._configure_traefik()
L4: self._restart_traefik()
```

but L2-L4 is already called by `self._configure()` which is right after this, causing duplicate code calls every time pebble_ready is called. Hence i have simply extracted `delete_dynamic_configs` to this method and removed `_clear_all_configs_and_restart_traefik` from this method.

src/charm.py
```diff
@@ -1346,24 +1345,6 @@ def _handle_traefik_route_ready(self, event: TraefikRouteRequirerReadyEvent) ->
...
             event.defer()
             return
-        if self._static_config_changed:
-            # This will regenerate the static configs and reevaluate all dynamic configs,
-            # including this one.
-            self._update_ingress_configurations()
-
-        else:
-            try:
-                self._process_ingress_relation(event.relation)
```
swetha1654, line 1352: I would have to call `self.traefik.flush_dynamic_configs()` after this. But instead of doing that I deleted this entire section and called `._process_status_and_configurations()` in an attempt to make it more holistic. 
IngressSetupError is already called in that method. Only difference now is that if IngressSetupError is raised, Traefik will go to blocked status instead of `ActiveStatus("traefik-route relation degraded")`. Either ways traefik will still work.

src/traefik.py
```diff
@@ -734,33 +738,86 @@ def delete_dynamic_configs(self) -> None:
...
             ["find", DYNAMIC_CONFIG_DIR, "-name", "*.yaml", "-delete"],
             timeout=60,
         ).wait()
         logger.debug("Deleted all dynamic configuration files.")
 
-    def delete_dynamic_config(self, file_name: str) -> None:
-        """Delete a specific yaml from the dynamic config dir."""
+    def delete_dynamic_config(self, key: str) -> None:
+        """Delete a specific yaml from the dynamic config dir and remove it from internal state."""
+        self._dynamic_configs.pop(key, None)
```
seb4stien, line 750: Shouldn't we do this at the end of the function just in case something goes wrong with the container interactions?

src/traefik.py
```diff
@@ -734,33 +738,86 @@ def delete_dynamic_configs(self) -> None:
...
-        # make_dirs is technically not necessary at runtime, since traefik.configure() should
-        # guarantee that the dynamic config dir exists. However, it simplifies testing as it means
-        # we don't have to worry about setting up manually the traefik container every time we
-        # simulate an event.
-        self._container.push(Path(DYNAMIC_CONFIG_DIR) / file_name, config, make_dirs=True)
+        # Remove all existing per-relation config files first; the archive
+        # extraction below will recreate the wanted ones.
+        try:
+            self._container.exec(
+                ["find", DYNAMIC_CONFIG_DIR, "-name", f"{INGRESS_CONFIG_PREFIX}*.yaml", "-delete"]
```
seb4stien: It feels a bit dangerous to delete everything at the beginning, if anything fails after that we're left with no config.
Ideally we should look for something atomic: the deletion of the files and the write of the new files should happen in a single command (if the DYNAMIC_CONFIG_DIR only contains these files, this may be done by creating a new folder, and replace the old one).
Plan B: push the new files, and remove the old ones afterwards (could probably be achieved by looking at the creation or modification time of the files in the find command).

seb4stien (commented):
Thanks for your comments, it helps understanding some of the changes.

For the main purpose of the PR, we need to revisit some of the elements to ensure we don't end up with an empty config if some operations fail. I put some ideas that we can discuss if I missed something.

src/traefik.py
swetha1654: the `_dynamic_configs` is an attribute that is created from the relation data at every hook, so even if something goes wrong at the container interactions, losing this data in dynamic_configs has no real consequence.

src/traefik.py
swetha1654: I updated the code to remove and write in the same command.

src/traefik.py
swetha1654: `_dynamic_configs` is an attribute that is constructed at every hook from the relation data. Even if something goes wrong with the container interactions, i don't see any consequence of losing the key. It will be re-added in the next hook.

seb4stien (approved):

PR #678, opened 2026-05-08
fix: Handle traefik-route

### LINKED ISSUE #676: Traefik-route relation data is not cleaned up when traefik is blocked
### Bug Description

The charm code wipes ingress relation data for `ingress` and `ingress-per-unit` when certain conditions are not met but does not handle `traefik-route`.

### Impact

High (major functionality broken, no workaround)

### Impact Rationale

_No response_

### To Reproduce

1. Deploy COS lite bundle
2. Set routing_mode to subdomain.
3. Traefik is now blocked. Ingress relation data for all apps are cleared except grafana (which uses traefik-route)

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

_No response_

swetha1654:
#### What this PR does

Fixes #676 

#### Why we need it

It doesn't make sense to wipe only ingress and ingress-per-unit when traefik is blocked, leaving only traefik-route apart simply because the cleanup method doesn't exist in the lib.

#### Checklist

- [x] **If this PR modifies charm libraries owned by this project**: I incremented the `LIBAPI` and `LIBPATCH` values

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1340,24 +1340,6 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         if not self.container.can_connect():
             event.defer()
             return
-        if self._static_config_changed:
-            # This will regenerate the static configs and reevaluate all dynamic configs,
-            # including this one.
-            self._update_ingress_configurations()
-
-        else:
-            try:
-                self._process_ingress_relation(event.relation)
-            except IngressSetupError as e:
-                err_msg = e.args[0]
-                logger.error(
-                    "failed processing the ingress relation for "
-                    f"traefik-route ready with: {err_msg!r}"
-                )
-
-                self.unit.status = ActiveStatus("traefik-route relation degraded")
-                return
-
         try:
             self.traefik.generate_static_config(_raise=True)
         except StaticConfigMergeConflictError:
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -1340,24 +1340,6 @@ def _handle_traefik_route_ready(self, event: TraefikRouteRequirerReadyEvent) ->
...
-            except IngressSetupError as e:
-                err_msg = e.args[0]
-                logger.error(
-                    "failed processing the ingress relation for "
-                    f"traefik-route ready with: {err_msg!r}"
-                )
-
-                self.unit.status = ActiveStatus("traefik-route relation degraded")
-                return
-
```
swetha1654, line 1360: The code after this is calling generate_static_config which will raise StaticConfigMergeConflictError if there is an issue with the traefik-route relation data. It doesn't make sense to have this code here since `_static_config_changed` is also internally calling `generate_static_config`, and the StaticConfigMergeConflictError is not handled here.

The logic inside the try block above is basically called inside `_process_status_and_configurations` so i added that after the try block below.

tests/scenario/test_traefik_route.py
```diff
@@ -0,0 +1,153 @@
...
+            remote_app_data={"config": ROUTE_CONFIG},
+            local_app_data={"external_host": "10.0.0.1", "scheme": "http"},
+        )
+
+        state = State(
+            leader=True,
+            config={"external_hostname": "testhostname", "routing_mode": "path"},
+            relations=[tr_rel],
+            containers=[traefik_container],
+        )
```
alithethird: Nit: I think this can be a fixture, we repeat it every test.

alithethird (approved):
1 nitpick otherwise lgtm!

Issue #684, opened 2026-05-22
traefik-k8s in error hook failed: "update-status"


hemanthnakkina:
### Bug Description

Reported in snap-openstack/sunbeam bug: https://bugs.launchpad.net/snap-openstack/+bug/2153947

Juju debug-log of traefik-rgw/0 in a failed env:

unit-traefik-rgw-0: ERROR juju.worker.uniter.operation
  hook "traefik-pebble-ready" (via hook dispatching script: dispatch) failed: exit status 1
unit-traefik-rgw-2: ERROR unit.traefik-rgw/2.juju-log Uncaught exception while in charm code:
  Traceback (most recent call last):
      raise ExecError(self._command, exit_code, None, None)
  ops.pebble.ExecError: non-zero exit code 1 executing
      ['find', '/opt/traefik/juju', '-name', '*.yaml', '-delete']

The traefik-k8s charm rev 295 (latest/candidate) appears to have a race in its pebble-ready handler when /opt/traefik/juju does not exist yet on first container start. The find ... -delete fails, the hook errors, and the unit enters the error state. Subsequent update-status hook invocations also fail because the unit is in a stuck state.

Note: the ExecError message appears in the debug log of every env in the test, including successful ones (in the success envs the charm self-recovers via retry). So the ExecError itself is benign; what kills failed envs is when the cascading state from other issues like mysql-router stuck and traefik-k8s requirer apps stuck keeps the relations in an unsteady state, leaving traefik's update-status hook unable to converge.

juju resolved is INSUFFICIENT to recover this state — re-running the failed hook does not clear the underlying pod-level state. Only kubectl rollout restart sts/<traefik-app> (delete + recreate the pods) recovers them.

### Impact

Critical (system down, data loss, affecting deployment in production)

### Impact Rationale

Manual intervention is required to make traefik-k8s back to active state. Expected traefik-k8s to converge after recovering the requirer apps. 

### To Reproduce

This is observed during 16 parallel sunbeam deployments and unfortunately i do not have a small reproducer. But as per the bug report from sunbeam, the logs should have the error messages even in successful condition.

### Environment

Sunbeam environment
traefik-k8s latest/candidate rev 295

### Relevant log output

```shell
See the logs in https://bugs.launchpad.net/snap-openstack/+bug/2153947
```

swetha1654:
[...]

PR #685, opened 2026-05-22
bug: Add a check before deleting dynamic config

swetha1654:
#### What this PR does

Addresses: #684 

#### Why we need it

Added existence guards in `delete_dynamic_configs` and `delete_dynamic_config` to prevent `ExecError` when `/opt/traefik/juju` does not exist on first container start or after a pod churn. Additionally, added a guard in the `_on_traefik_pebble_ready` handler to defer the event until the container is ready.

#### Checklist

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,6 +5,11 @@ on:
     branches:
       - main
 
+permissions:
+  actions: read
+  contents: read
+  security-events: write
+
 jobs:
   pull-request:
     name: PR
--- a/src/charm.py
+++ b/src/charm.py
@@ -1060,9 +1060,12 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
     def _configure_traefik(self) -> None:
         self.traefik.configure()
 
-    def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
+    def _on_traefik_pebble_ready(self, event: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
+        if not self.traefik.is_ready:
+            event.defer()
+            return
         self._clear_all_configs_and_restart_traefik()
         # push the (fresh new) configs.
         self._configure()
@@ -1074,6 +1077,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         # configuration files on a storage volume that survives the pod churn, before
         # we start traefik we clean up all Juju-generated config files to avoid spurious
         # routes.
+            
         self.traefik.delete_dynamic_configs()
 
         # we push the config
--- a/src/charm.py
+++ b/src/charm.py
@@ -1060,7 +1060,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
     def _configure_traefik(self) -> None:
         self.traefik.configure()
 
-    def _on_traefik_pebble_ready(self, event: PebbleReadyEvent) -> None:
+    def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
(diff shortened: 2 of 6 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -5,6 +5,11 @@ on:
     branches:
       - main
 
+permissions:
+  actions: read
+  contents: read
+  security-events: write
+
```
swetha1654, line 12: o11y team changed their workflows, this is now required for the workflow to work.

src/charm.py
```diff
@@ -1060,9 +1060,12 @@ def _udp_entrypoints(self) -> Dict[str, Any]:
     def _configure_traefik(self) -> None:
         self.traefik.configure()
 
-    def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
+    def _on_traefik_pebble_ready(self, event: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
+        if not self.traefik.is_ready:
+            event.defer()
```
Thanhphan1147: Can you elaborate a bit on why a defer is needed here?

swetha1654: This event eventually calls delete_dynamic_configs which calls `self._container.exists`. If the container isn't ready it will fail. 
And since this is a non-wholistic charm, i think we should defer the event if the container isn't ready so we don't lose the event.

src/traefik.py
```diff
@@ -727,6 +727,12 @@ def restart(self) -> None:
...
     def delete_dynamic_configs(self) -> None:
         """Delete **ALL** yamls from the dynamic config dir."""
+        if not self._container.exists(DYNAMIC_CONFIG_DIR):
+            logger.debug(
+                "Dynamic config dir %s does not exist; nothing to delete.",
+                DYNAMIC_CONFIG_DIR,
+            )
+            return
         # instead of multiple calls to self._container.remove_path(), delete all files in a swoop
         self._container.exec(
```
Thanhphan1147, line 737: Maybe instead of guarding we can delete the files in an idempotent way?

src/charm.py
Thanhphan1147: Maybe add a docstring so that the reason is clear when looking at the code?

Thanhphan1147 (approved):

src/traefik.py
swetha1654: how do you propose we do it?

src/charm.py
javierdelapuente: but... if the container is not accessible, we will get another _on_traefik_pebble_ready, are we sure we need to defer?

swetha1654: that's a good point

javierdelapuente (approved):

src/charm.py
swetha1654: @javierdelapuente Actually, the thing with defer is that "any deferred events get processed before the event (or action) that caused the current invocation of the charm". So instead of waiting for the next traefik_pebble_ready event to come after other events, i'd rather defer this and wait for pebble_ready event to run successfully before any other hooks run. Cause the other hooks can fail if this doesn't run.

src/charm.py
```diff
@@ -1060,9 +1060,14 @@ def _udp_entrypoints(self) -> Dict[str, Any]:
     def _configure_traefik(self) -> None:
         self.traefik.configure()
 
-    def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
+    def _on_traefik_pebble_ready(self, event: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
+        # We need to defer the event if the container is not reachable, as we don't want to lose
+        # the event in a non-wholistic charm.
```
seb4stien:
```suggestion
        # the event in a non-holistic charm.
```

seb4stien (approved):

src/charm.py
swetha1654: okay i will remove defer now, will add it back if they really see an issue.

PR #687, opened 2026-05-25
feat: add charm-ci integration test workflow

javierdelapuente:
## Summary

Integrates `canonical/charm-ci` reusable workflows and reworks integration test layout for charm-ci discovery.

### CI
- Add `integration-test.yaml` and `publish.yaml`
- Remove old observability `quality-gates.yaml` and `release.yaml`
- Update `pull-request.yaml` to stop handling integration gating itself

### Integration tests
- Keep jubilant tests at `tests/integration/` with a dedicated jubilant `conftest.py`
- Move ops_test-based tests to `tests/integration/ops_test/`
- Update `spread.yaml` MODULE paths for the moved tests

### Notes
- ops_test tests still build the charm if `CHARM_PATH` is not set
- jubilant tests use their own fixture set and include the Juju 4 RBAC workaround for canonical k8s


.github/workflows/quality-gates.yaml
swetha1654, line 1: we can actually remove this workflow entirely. Its not used at all

spread.yaml
```diff
@@ -0,0 +1,39 @@
...
+- .tox
+- .venv
+- .*_cache
+suites:
+  tests/integration/:
+    summary: integration tests
+    backends:
+    - integration-test
+    environment:
+      MODULE/test_basic_auth: test_basic_auth
```
swetha1654: what is the value of module here? Can we provide different paths to integration tests?

.github/workflows/pull-request.yaml
swetha1654, line 1: so the lint and unit tests will still run from this workflow right? Do we plan to migrate these also to charm-ci in a follow up PR?

.github/workflows/publish.yaml
swetha1654, line 1: have you tried testing this workflow? If yes could you link the workflow here?

swetha1654 (commented):
Could you also update the CONTRIBUTING guide with how to use this locally?

spread.yaml
javierdelapuente: The module in here is just the "test_*.py" file. 

Not yet, but it will be possible in the following release of charm-ci several paths.

.github/workflows/pull-request.yaml
javierdelapuente: Nope, initially we do not plan to include unit tests/linting in charm-ci. I know there is an initiative by charmtech about this too. 

If we want to remove the dependency on observability, I would use operator-workflows for now (not for this PR).

.github/workflows/publish.yaml
javierdelapuente: I have tested it in my own repos like https://github.com/javierdelapuente/javi-traefik-k8s-operator/actions/runs/26571537214/job/78281513159 and https://github.com/javierdelapuente/javi-indico-operator/actions/runs/26534524266/job/78159781831

This is the first try in the canonical organization...

.github/workflows/quality-gates.yaml
swetha1654: can you remove this workflow @javierdelapuente ?

swetha1654 (approved):

swetha1654:
please make sure `canonical-repo-automation` is also updated with the new required workflows @javierdelapuente

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,18 @@
+name: Integration Tests
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
```
seb4stien: I think the recommendation is to have the permissions at the job level to be specific.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,33 @@
+name: Publish
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
```
seb4stien: Should be at the job level.

seb4stien: Which is already the case for publish

spread.yaml
```diff
@@ -0,0 +1,34 @@
...
+  CONCIERGE: '$(HOST: echo "${CONCIERGE:-concierge.yaml}")'
+  OPCLI_GIT_REF: '$(HOST: echo "${OPCLI_GIT_REF:-main}")'
+exclude:
+- .git
+- .tox
+- .venv
+- .*_cache
+integration-suites:
+  tests/integration/:
+    cwd: ./
```
javierdelapuente: working-dir

spread.yaml
```diff
@@ -0,0 +1,34 @@
...
+    summary: integration tests
+    backends:
+    - integration-test
+    auto-discover: true
+    pytest-arguments-template: ""
+    pytest-environment-template: |
+      {% for build in artifacts.charms[0].builds if build.arch == arch %}
+      CHARM_PATH={{ build.path }}
```
javierdelapuente, line 28: what guardrails can we have in here?

spread.yaml
```diff
@@ -0,0 +1,34 @@
...
+path: /home/ubuntu/proj
+kill-timeout: 60m
+warn-timeout: 1m
+backends:
+  integration-test:
+    type: integration-test
+    systems:
+    - ubuntu-24.04
+environment:
+  CONCIERGE: '$(HOST: echo "${CONCIERGE:-concierge.yaml}")'
```
javierdelapuente: name it concierge-3.6 by default instead of plain concierge.yaml?

.gitignore
```diff
@@ -53,3 +53,8 @@ terraform.rc
 
 # Tests generate this file
 .charm_tracing_buffer.raw
+
+# opcli/spread generated files
+.spread-reuse.*
+.spread-run-*/
+artifacts.build.yaml
```
seb4stien, line 60: As you are know using build/, do you think it should be moved there too?
I'm not sure, just an open question.

spread.yaml
seb4stien: ~~Should we also consider the base? => "concierge-<ubuntu base>-<juju lts>" => "concierge-26.04-4.1"~~ <- forget about it, it comes from spread.

My goal is to have a naming convention that supports the versions we may have to maintain in parallel from the beginning.

seb4stien (approved):

spread.yaml
javierdelapuente: The new pytest plugin makes this an exception. I will adapt traefik to the plugin in a following PR.

spread.yaml
javierdelapuente: Adapting to 3.6-4.0 will always be manual with the current approach (but quite trivial). 

I have change the concierge file to conciege-juju3 to make it obvious what we are testing.

.gitignore
javierdelapuente: Let's discuss this in the spec (I will update it) and then we can update traefik.

Weii also had a proposal in the spec to rename artifacts.yaml to build.yaml (and the generated one to .build.yaml).

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,17 @@
+name: Integration Tests
+
+on:
+  pull_request:
+    branches:
+      - main
```
swetha1654: copilot sometimes works on my current branch so i'd drop the branch "main"

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  integration-test:
+    permissions:
+      actions: read
+      contents: read
+      packages: write
+    uses: canonical/charm-ci/.github/workflows/integration-test.yml@v0.0.1-alpha.1
```
swetha1654, line 12: is this the latest version of the workflow?

.github/workflows/publish.yaml
swetha1654, line 1: the current release.yaml is doing a lot more pre-checks before publishing the charm to edge. We need to check if these checks are already covered by the PR workflows once.

tests/integration/ops_test/__init__.py
swetha1654, line 1: can we rename `ops_test` folder to `legacy`? This is the convention we are following everywhere else.

tests/integration/conftest.py
```diff
@@ -1,441 +1,110 @@
...
-):
-    """Get relation databags for a juju relation.
+    The root cause appears to be that canonical k8s does not correctly evaluate
+    ``apiGroups: ["*"]`` + ``resourceNames: [specific-name]`` role rules for
+    the core API group secrets resource.
 
-    >>> get_relation_data('prometheus/0:ingress', 'traefik/1:ingress-per-unit')
+    As a workaround we pre-create a permissive Role + RoleBinding that grants
+    all service accounts in the namespace the ability to manage secrets.  This
+    is safe because the namespace is a throwaway test environment.
```
swetha1654, line 103: is there an existing issue that you can link here? Maybe there's a cleaner workaround

tests/integration/conftest.py
```diff
@@ -1,441 +1,110 @@
...
-    # are already related.
-    try:
-        await ops_test.model.add_relation(ep1, ep2)
-    except juju.errors.JujuAPIError as e:
-        # relation already exists? skip
-        logging.error(e)
-        pass
+@pytest.fixture(autouse=True, scope="module")
+def copy_traefik_library_into_tester_charms():
+    """No-op: jubilant tests deploy from Charmhub, not local tester charms."""
```
swetha1654: this is weird. It doesn't do anything

tests/integration/conftest.py
```diff
@@ -1,441 +1,110 @@
...
 @pytest.fixture(autouse=True, scope="module")
-async def setup_env(ops_test: OpsTest):
-    # Prevent "update-status" from interfering with the test:
-    # - if fired "too quickly", traefik will flip between active/idle and maintenance;
-    # - make sure charm code does not rely on update-status for correct operation.
-    await ops_test.model.set_config(
-        {"update-status-hook-interval": "60m", "logging-config": "<root>=WARNING; unit=DEBUG"}
-    )
+async def setup_env():
+    """No-op: jubilant tests manage their own model via jubilant."""
```
swetha1654: why do we need this?

tests/integration/test_dynamic_configs.py
```diff
@@ -29,26 +28,9 @@
...
-    charms = sorted(Path(".").glob("traefik*.charm"))
-    if charms:
-        return charms[0]
-    raise FileNotFoundError(
-        "Set CHARM_PATH to the built traefik charm, "
-        "or place a traefik*.charm file in the repo root."
-    )
+def _all_settled(status: jubilant.Status) -> bool:
+    """Return True when all apps are active and all agents are idle."""
+    return jubilant.all_active(status) and jubilant.all_agents_idle(status)
```
swetha1654: this could be inside `helpers.py`

tests/integration/test_dynamic_configs.py
```diff
@@ -61,7 +43,7 @@ def deploy_traefik(juju, traefik_charm):
         trust=True,
     )
     juju.config(TRAEFIK_APP_NAME, {"external_hostname": "traefik.test"})
-    juju.wait(jubilant.all_active, timeout=600)
+    juju.wait(_all_settled, delay=5, timeout=600)
```
swetha1654: Can we also move the traefik charm deployment to `conftest.py`?

tests/integration/test_dynamic_configs.py
```diff
@@ -74,9 +56,8 @@ def deploy_alertmanager(juju, deploy_traefik):
         channel="2/edge",
         trust=True,
     )
-    juju.wait(jubilant.all_active, timeout=600)
     juju.integrate(f"{ALERTMANAGER_APP_NAME}:ingress", TRAEFIK_APP_NAME)
-    juju.wait(jubilant.all_active, timeout=600)
+    juju.wait(_all_settled, delay=5, timeout=600)
```
swetha1654: Can we also move the alertmanager charm deployment to `conftest.py`?

tests/integration/test_dynamic_configs.py
```diff
@@ -89,9 +70,8 @@ def deploy_catalogue(juju, deploy_traefik):
         channel="1/edge",
         trust=True,
     )
-    juju.wait(jubilant.all_active, timeout=600)
     juju.integrate(f"{CATALOGUE_APP_NAME}:ingress", TRAEFIK_APP_NAME)
-    juju.wait(jubilant.all_active, timeout=600)
+    juju.wait(_all_settled, delay=5, timeout=600)
```
swetha1654: same here

tests/integration/test_receive_ca_cert_status.py
```diff
@@ -30,10 +30,10 @@ def juju():
 def traefik_charm():
     charm_path = os.environ.get("CHARM_PATH")
     if charm_path:
-        return Path(charm_path)
+        return Path(charm_path).resolve()
```
swetha1654: why this change? and why is the traefik charm fixture defined in this file again? can we remove it and use the global one?

spread.yaml
```diff
@@ -0,0 +1,34 @@
...
+    pytest-arguments-template: '-x --log-format="%(asctime)s %(levelname)s %(message)s" --keep-models --model testing'
+    pytest-environment-template: |
+      {% for build in artifacts.charms[0].builds if build.arch == arch %}
+      CHARM_PATH={{ build.path }}
+      {% endfor %}
+    environment:
+      MODULE/jubilant_test_dynamic_configs_juju4: tests/integration/test_dynamic_configs.py
+      MODULE/jubilant_test_https_multi_unit_juju4: tests/integration/test_https_multi_unit.py
+      CONCIERGE/jubilant_test_dynamic_configs_juju4: concierge-juju4.yaml
+      CONCIERGE/jubilant_test_https_multi_unit_juju4: concierge-juju4.yaml
```
swetha1654, line 35: instead of specifying the env for juju 4, can we specify it for juju 3 instead? The goal is to eventually migrate entirely from juju 3 to 4 and that time we don't want 2 env variables for each test file, even if it means an ugly env now.

swetha1654 (commented):
LGTM for the most part, just a few minor comments. Thanks!

tests/integration/conftest.py
javierdelapuente: https://github.com/juju/juju/issues/22485
https://github.com/juju/juju/pull/22580

I think once the last PR is merged, it will work. I tried other workarounds, but no luck...

.github/workflows/integration-test.yaml
javierdelapuente: I just created it :)

.github/workflows/publish.yaml
javierdelapuente: They follow a different path, they build everything again and test everything again. 

We follow the operator-workflows path. Get the last correctly executed integration test before merging.

tests/integration/conftest.py
gfouillet: The last PR is (mostly) refactoring, so I would be surprised it fix something. This is part of an epic which put all commitHook changes as a transaction instead of several transaction (which is error prone, because juju could be in an inconsistent state in case of error during a hook execution. After this EPIC (two remaining PR around update and create secrets), this should not be the case anymore.

The first one, though, might be related, since it fix a regression between 3.6 and 4.0 which prevent to use user secret on k8s, and it should fix both user and charm secret (despite the QA doesn't test it). 
I have kept a note in my task to verify this claim is still true after the current refactoring will be finished.

FYI, I've planned to finish this work soon (no more than two week), so it should be released in the next release after.

If you need more insight, don't hesitate to reach me.

spread.yaml
javierdelapuente: The auto-discover creates all the MODULE variants "magically" for all the test files (it finds them recursively).

The default CONCIERGE is concierge-juju3.yaml. 

Once all tests work with juju4, we can swap the default one,  create two integration-suites (so there is no customization at all) or even two backends. 

The alternative is to not use autodiscover, but that will create a difficult to maintain (and not clear) spread.yaml.

.github/workflows/integration-test.yaml
javierdelapuente: That's the same that is in pull-request.yaml. But no need for that (in case you open a pull request targetting another branch).

Removed.

tests/integration/test_receive_ca_cert_status.py
javierdelapuente: @javierdelapuente add this to jubilant for juju4.

javierdelapuente: removed the fixture (unneeded, exactly the same as in conftest)

.github/workflows/publish.yaml
javierdelapuente: checked, looks our approach is correct

tests/integration/conftest.py
javierdelapuente: I will create a new PR after merging this one to run the tests without removing RBAC to ck8s. We will keep you informed. 

Thanks for the info and comment!

PR #689, opened 2026-05-26
fix: Use rootCAs instead of update-ca-certificates (4 of 5)

swetha1654:
#### What this PR does

This pull request refactors how Traefik's static configuration is updated to manage CA certificates, ensuring that all trusted CAs (including those received via relations) are correctly reflected in the `serversTransport.rootCAs` field. The changes eliminate the need to manually run `update-ca-certificates` and instead regenerate the static config whenever CA certs are added or removed. The integration and unit tests are also updated to verify this new behavior.

#### Why we need it

Optimization for #612 

#### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1522,7 +1522,6 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         prefix = self._get_prefix(data)  # type: ignore
         config = self.traefik.get_per_leader_http_config(
             prefix=prefix,
-            scheme="http",  # IPL (aka ingress v1) has no https option
             port=data["port"],
             host=data["host"],
             strip_prefix=data.get("strip-prefix", False),
--- a/src/charm.py
+++ b/src/charm.py
@@ -617,6 +617,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         if not self.container.can_connect():
             return
         self._update_received_ca_certs(event)
+        self._reconcile_lb()
         # Regenerate static config so rootCAs picks up the new CA, then restart.
         self.traefik.configure()
         self._restart_traefik()
@@ -648,6 +649,7 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         # Regenerate static config so rootCAs reflects the removed CA, then restart.
         self.traefik.configure()
         self._restart_traefik()
+        self._reconcile_lb()
 
     def _is_tls_enabled(self) -> bool:
         """Return True if TLS is enabled."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -617,8 +617,8 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         if not self.container.can_connect():
             return
         self._update_received_ca_certs(event)
-        self._reconcile_lb()
-        # We need to restart Traefik now
+        # Regenerate static config so rootCAs picks up the new CA, then restart.
+        self.traefik.configure()
         self._restart_traefik()
 
     def _update_received_ca_certs(
@@ -645,9 +645,9 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
     def _on_recv_ca_cert_removed(self, event: CertificateTransferRemovedEvent) -> None:
         # Assuming only one cert per relation (this is in line with the original lib design).
         self.traefik.remove_ca(str(event.relation_id))
-        # Since remove_ca will call update_ca_certs in traefik, a restart is needed.
+        # Regenerate static config so rootCAs reflects the removed CA, then restart.
+        self.traefik.configure()
         self._restart_traefik()
-        self._reconcile_lb()
 
     def _is_tls_enabled(self) -> bool:
         """Return True if TLS is enabled."""
@@ -714,10 +714,6 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         """Update the server cert, ca, and key configuration files."""
         self._sync_certs_to_peer_databag()
         self.traefik.update_cert_configuration(self._get_certs())
-
-        # update_cert_configuration relies on traefik.update_ca_certs.
-        # Thus, we should restart traefik with the new CA certs.
-        self._restart_traefik()
 
     def _sync_certs_to_peer_databag(self) -> None:
         """Sync certificates to the peer databag (leader only).
@@ -1057,9 +1053,6 @@ class TraefikIngressCharm(CharmBase):  # pylint: disable=too-many-instance-attri
         """Return UDP entryPoints sent via traefik_route."""
         return self._traefik_route_entrypoints(protocol="udp")
 
-    def _configure_traefik(self) -> None:
-        self.traefik.configure()
-
     def _on_traefik_pebble_ready(self, _: PebbleReadyEvent) -> None:
         # If the Traefik container comes up, e.g., after a pod churn, we
         # ignore the unit status and start fresh.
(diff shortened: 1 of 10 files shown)


src/traefik.py
```diff
@@ -522,60 +526,25 @@ def _generate_config_block(  # pylint: disable=too-many-arguments,too-many-local
...
-            # Since we use the update_ca_certificates machinery, we don't need to specify the
-            # "rootCAs" entry.
-            # Keeping the serverTransports section anyway because it is informative ("endToEndTLS"
-            # vs "reverseTerminationTransport") when inspecting the config file in production.
-            transport_name = "endToEndTLS"
-            lb_def["serversTransport"] = transport_name
-            transports = {transport_name: {"insecureSkipVerify": False}}
-
-        else:
-            transports = {}
```
swetha1654, line 572: For backend TLS verification we rely on the static serversTransport.rootCAs (set in `generate_static_config` as part of this PR) which applies to all services that don't reference a named transport. Named transports do NOT inherit rootCAs from the static config, so we need to avoid assigning a named transport in the dynamic config. 
Looking at what the named serverTransport in the dynamic config is currently doing, its only setting `{"insecureSkipVerify": False}` with different transport names for troubleshooting purposes. The [default](https://doc.traefik.io/traefik/v2.11/routing/overview/#insecureskipverify) value of insecureSkipVerify is already false, so this section does not really add any value other than readability during troubleshooting.

src/traefik.py
```diff
@@ -287,19 +277,28 @@ def remove_ca(self, uid: str) -> None:
...
-        self._container.exec(["update-ca-certificates", "--fresh"]).wait()
+    def _collect_root_ca_paths(self) -> List[str]:
+        """Collect all CA certificate file paths from the CA certs directory.
+
+        This includes both pre-bundled CAs from the container image and
+        charm-managed CAs (from TLS and receive-ca-cert relations).
+        """
+        ca_paths: List[str] = []
+        try:
+            if self._container.can_connect() and self._container.isdir(str(CA_CERTS_DIR)):
```
javierdelapuente: Isn't this quite different from the system-wide Root CA certificates?

src/charm.py
```diff
@@ -1527,7 +1522,6 @@ def _get_configs_per_leader(self, relation: Relation) -> Dict[str, Any]:
         prefix = self._get_prefix(data)  # type: ignore
         config = self.traefik.get_per_leader_http_config(
             prefix=prefix,
-            scheme="http",  # IPL (aka ingress v1) has no https option
```
seb4stien, line 1506: Is it related to the PR?

src/charm.py
```diff
@@ -715,10 +717,6 @@ def _update_cert_configs(self) -> None:
         self._sync_certs_to_peer_databag()
         self.traefik.update_cert_configuration(self._get_certs())
 
-        # update_cert_configuration relies on traefik.update_ca_certs.
-        # Thus, we should restart traefik with the new CA certs.
-        self._restart_traefik()
```
seb4stien: Is it happening somewhere else in the flow now?

src/traefik.py
```diff
@@ -493,7 +498,6 @@ def _generate_config_block(  # pylint: disable=too-many-arguments,too-many-local
         unit and IPA may be more than one).
         """
         # purge any optionals:
-        scheme_: str = scheme if scheme is not None else "http"
```
seb4stien, line 500: I'm not sure to get the potential impacts of this one (and all related changes).

seb4stien (commented):
I'm a bit confused by the code removed related to is_reversation_termination, is_termination and so one. Is it related to this PR? It feels we would loose some features going this way.

src/charm.py
swetha1654: @seb4stien everything that calls this function (`update_cert_configuration`) already calls `restart_traefik` so this is actually redundant.

src/charm.py
```diff
@@ -645,7 +646,8 @@ def _update_received_ca_certs(
     def _on_recv_ca_cert_removed(self, event: CertificateTransferRemovedEvent) -> None:
         # Assuming only one cert per relation (this is in line with the original lib design).
         self.traefik.remove_ca(str(event.relation_id))
-        # Since remove_ca will call update_ca_certs in traefik, a restart is needed.
+        # Regenerate static config so rootCAs reflects the removed CA, then restart.
+        self.traefik.configure()
         self._restart_traefik()
```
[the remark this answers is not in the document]
swetha1654: unrelated to this PR but yes its a bug

src/charm.py
swetha1654: yes, pls see [this](https://github.com/canonical/traefik-k8s-operator/pull/689/changes#r3309700748) comment

src/traefik.py
swetha1654: @javierdelapuente I have updated the code now to include the system CA bundle as well, thanks for pointing it out

src/traefik.py
swetha1654: Pls see [this](https://github.com/canonical/traefik-k8s-operator/pull/689/changes#r3309700748) comment

src/traefik.py
```diff
@@ -287,19 +277,35 @@ def remove_ca(self, uid: str) -> None:
...
+        This includes the system CA bundle (so public CAs remain trusted)
+        plus charm-managed CAs (from TLS and receive-ca-cert relations).
+        """
+        ca_paths: List[str] = []
+        try:
+            if not self._container.can_connect():
+                return []
+            # Include the system CA bundle so that backends with publicly-signed
+            # certificates remain trusted when rootCAs is set.
+            system_bundle = "/etc/ssl/certs/ca-certificates.crt"
```
swetha1654, line 303: check if this is pre-bundled in the rock or not

tests/integration/test_tls_jubilant.py
```diff
@@ -0,0 +1,300 @@
...
+    session.verify = str(ca_cert_path)
+
+    response = session.get(alertmanager_url, timeout=30)
+    logger.info(
+        "End-to-end TLS result: status=%s body=%s",
+        response.status_code,
+        response.text[:200],
+    )
+    # Without the rootCAs fix, this would be a 502 because traefik can't verify
+    # alertmanager's backend certificate.
```
swetha1654, line 299: remove comment

src/charm.py
swetha1654, line 1: If user provides custom rootCA in the dynamic config file section, how does traefik behave?

seb4stien (approved):

src/traefik.py
swetha1654: yes it comes pre-bundled

swetha1654: `sudo docker run --rm --entrypoint cat docker.io/ubuntu/traefik:2-22.04 /etc/ssl/certs/ca-certificates.crt` returns a certificate

swetha1654: @javierdelapuente I have modified the code to raise an exception if the system CA is missing. This way we can be sure across diff rock versions.

src/charm.py
swetha1654: this shouldn't be a problem, Traefik will pick up the rootCAs in the dynamic config.

swetha1654:
This PR cannot be merged in its current state as it has the following problems:

1. **Scenario 1**: If user provides a custom ServerTransport in their static config:

Right now traefik tries to deep merge the existing static config with the user provided static config via `traefik-route`. The way the deep_merge logic works is this:
```
def static_config_deep_merge(dict1: dict, dict2: dict, _path: Optional[list] = None) -> dict:
    """In-place deep merge dict2 into dict1."""
    _path = _path or []

    for key, val in dict2.items():
        if key in dict1:
            val1 = dict1[key]
            if isinstance(val, dict) and isinstance(val1, dict):
                static_config_deep_merge(val1, val, _path + [str(key)])
            elif val != val1:
                raise StaticConfigMergeConflictError(f"Conflict at path {'.'.join(_path)}")
        else:
            dict1[key] = val
    return dict1
```
Traefik only merges 2 fields if they have the same values for a given key. If the user provides custom rootCA certs in their static config, this will break them.

2. **Scenario 2**: If user provides a custom ServerTransport in their dynamic config:

If uesrs provide a custom named serverTransport in their dynamic config, for e.g., something like:
```
http:
  serversTransports:
    myTransport:
      forwardingTimeouts:
        dialTimeout: 30s
```
but expect the certificates from the static config to work, this wouldn't work since the certificates are not inherited if there is a named serverTransport in the dynamic config.

cc: @seb4stien @javierdelapuente

swetha1654:
Conclusion: We will not implement this, its too complex

PR #698, opened 2026-06-04
Update Terraform module to support v2

### LINKED ISSUE #690: Add support for Juju 4 in the Terraform module
### Enhancement Proposal

The TF Juju provider [released 2.0.0](https://github.com/juju/terraform-provider-juju/releases/tag/v2.0.0) on May 26th, 2026 and we need to support this because of the Juju v4 support. This major version is said to not have any breaking changes so we are safe to bump our Juju provider version support. In our observability charms, we decided to switch the Juju provider version pin to:
- `"~> 1.0"` -> `">= 1.0"` -> so that the modules implementing Traefik can choose what to pin the providers to.
- https://github.com/canonical/alertmanager-k8s-operator/pull/415

You may decide to go with a different pinning strategy e.g., `">= 1.0, <3.0"`

Bumping the Juju provider support to v2 implies that Traefik supports Juju v4, so you should only bump the requirement once testing is complete and the module is ready for users.

### Impact

High (The feature has short-term technical value)

### Impact Rationale

This blocks COS from deploying on Juju 4. Since the Juju provider v2 was released recently, I chose to set this as "High", not "Critical", but eventually this will become a critical ask from us.

swetha1654:
#### What this PR does

Closes #690 

#### Checklist


terraform/README.md
```diff
@@ -3,33 +3,47 @@
...
+| <a name="input_units"></a> [units](#input\_units) | Unit count/scale | `number` | `1` | no |
 
-## Usage
+## Outputs
 
-### Basic usage
+| Name | Description |
+|------|-------------|
+| <a name="output_app_name"></a> [app\_name](#output\_app\_name) | n/a |
+| <a name="output_endpoints"></a> [endpoints](#output\_endpoints) | n/a |
```
f-atwi: should the description for this stay n/a?

seb4stien (approved):

seb4stien:
We need to clarify if we will still ship a terraform module for Traefik as it's not mandatory from the CC008 spec.

seb4stien (changes requested):
We have to see if we really want to support terraform provider v2 or if we discontinue the terraform module as it's not mandatory for such charms from CC008.

terraform/README.md
swetha1654: updated

swetha1654:
decision: We will keep the terraform module for the standard inputs/output.

f-atwi (approved):

terraform/outputs.tf
```diff
@@ -1,22 +1,28 @@
-output "app_name" {
```
seb4stien, line 1: You probably need to keep the old one if you don't want to introduce a breaking change.

terraform/outputs.tf
```diff
@@ -1,22 +1,28 @@
...
-  value = juju_application.traefik.name
+output "application" {
+  description = "The deployed traefik-k8s application."
+  value       = juju_application.traefik
 }
 
-output "endpoints" {
```
seb4stien, line 5: You probably need to keep the old one if you don't want to introduce a breaking change.

seb4stien (approved):

PR #728, opened 2026-07-03
Update upgrade guide for app mode

Thanhphan1147:
#### What this PR does

Add a revision specific upgrade guide due to the certificate management change from UNIT mode to APP mode

#### Checklist


docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
 juju refresh traefik-k8s
 ```
+
+## Revision-specific upgrade notes
+
+Some revisions require additional manual steps after upgrading. Check the section that applies to your current revision before refreshing.
+
+(upgrade_rev281_to_rev308)=
+
+### Upgrading from rev281–rev307 to rev308 or later
```
swetha1654: I think this should be upgrading from 280

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+Revision 308 fixes a bug where Traefik could request a new TLS certificate unnecessarily after a Juju leader change. If your deployment was running any revision between 281 and 307, follow these steps.
+
+#### Background
+
+Revisions 281–307 stored the TLS private key in a **unit-scoped** Juju secret. Because each unit held its own private key, only the leader unit's key was used to generate the Certificate Signing Request (CSR). When the leader changed, the new leader had a different private key that did not match the existing certificate, causing an unnecessary certificate re-request.
```
swetha1654: Any revision< 281 had unit-scoped secret.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+Revision 308 corrects this by storing the private key in a single **application-scoped** secret shared across all units. However, when upgrading from a previously affected revision, the old unit-scoped secrets are not automatically removed by Juju. Because the charm identifies secrets by label and `juju secret get` matches both unit-scoped and application-scoped secrets, the charm may find the stale unit-scoped secret and never create the correct application-scoped one.
```
swetha1654: Isn't revision 281 associated with https://github.com/canonical/traefik-k8s-operator/pull/620 already sharing a app scoped secret across all units?

swetha1654: > Because the charm identifies secrets by label and `juju secret get` matches both unit-scoped and application-scoped secrets

The TLS library looks for the Mode and correspondingly fetches either the unit or the app secret only, not both

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+To avoid this, you can restore the original private key immediately after the refresh, so that Traefik re-uses the same key and produces the same CSR that was already signed. This allows you to provide the original certificate without any re-signing.
+
+### Steps
+
+**1. Before upgrading**, retrieve and save the current private key and certificate from any Traefik unit:
+
+```
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.key
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.crt
+```
```
swetha1654: What happens if the user has already upgraded and lost the old private key and cert? Its better to put this step in the very beginning of the document before asking them to upgrade.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+jhack eval traefik-k8s/leader self._update_private_key()
+```
+
+**3. Fire a `config-changed` event** to trigger Traefik to re-generate the CSR with the restored key:
+
+```
+jhack fire traefik-k8s/leader config-changed
+```
+
+After this, the original CSR will reappear in the relation databag. You can then provide the previously signed certificate without any re-signing:
```
swetha1654: As mentioned in the standup, how can we be sure if the same CSR is generated? If things like datetime factor into CSR creation then this will break.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+#### Steps
+
+After running `juju refresh traefik-k8s --revision 308` (or later), check whether stale unit-scoped secrets remain:
+
+```bash
+juju secrets --format json | jq '.[] | select(.label | test("private-key"; "i"))'
+```
+
+If the output shows secrets with unit owners (e.g. `traefik-k8s/0`, `traefik-k8s/1`), the stale secrets are present and must be cleaned up. Proceed with listing and deleting each unit-scoped private-key secret:
```
seb4stien, line 49: Shouldn't we recommend to backup the content somewhere first? I know they are not supposed to be used, but we never now.

docs/how-to/upgrade.md
swetha1654: In fact, looking at the certificate library, I see a fresh unique subject identifier is added by default every time CSR generation happens. I'm surprised this worked on sunbeam.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+## Preserving TLS certificates after upgrade
+
+By default, `juju refresh` to a revision that uses APP mode will cause Traefik to generate a new private key, which triggers a new Certificate Signing Request (CSR). If you are using `manual-tls-certificates` or `notary` as your TLS provider, this means you will need to sign and provide a new certificate.
+
+To avoid this, you can restore the original private key immediately after the refresh, so that Traefik re-uses the same key and produces the same CSR that was already signed. This allows you to provide the original certificate without any re-signing.
+
+### Steps
+
+**1. Before upgrading**, retrieve and save the current private key and certificate from any Traefik unit:
```
seb4stien: As people may run the documentation "after upgrading", maybe we should add a not to mention that the key/crt may just be available on one/some units?

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+```
```
seb4stien:
```suggestion
```bash
```

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.key
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.crt
+```
+
+**2. Restore the private key on the leader unit.**
```
seb4stien: To be discussed if we turn this into an action.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,104 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+juju secrets --format json | jq '.[] | select(.label | test("private-key"; "i"))'
+```
+
+If the output shows secrets with unit owners (e.g. `traefik-k8s/0`, `traefik-k8s/1`), the stale secrets are present and must be cleaned up. Proceed with listing and deleting each unit-scoped private-key secret:
+
+```bash
+juju secrets --format json \
+  | jq -r '.[] | select(.label | test("private-key"; "i")) | select(.owner | test("/")) | .id' \
+  | xargs -I{} juju remove-secret {}
+```
```
swetha1654, line 55: We will also see the CSRs in the unit databag which must be cleaned up, we need to think about how to clean them.

docs/how-to/upgrade.md
Thanhphan1147: I think it's <308 because eventhough we were already in APP mode the private key itself was still an unit-scoped secret because the library was outdated and had a bug. We bumped that lib in rev 308 iirc

docs/how-to/upgrade.md
Thanhphan1147: I think the mode only matters when creating the secret, fetching them is done by `self.charm.model.get_secret(label=...)` and that can potentially return either the app-scoped one or the unit-scoped one, if there's still the unit-scope one leftover from a previous revision

docs/how-to/upgrade.md
Thanhphan1147: I'll probably need to update the wording on this, it's not "the original CSR" but "a CSR created by the original private key"

Thanhphan1147: @copilot update "the original CSR" to "a CSR created by the original private key"

docs/how-to/upgrade.md
Thanhphan1147: @copilot move this to the very first step before upgrade

docs/how-to/upgrade.md
Thanhphan1147: Yep, though a bug was in the lib that makes the charm still creates unit-scoped secrets. This was fixed in rev 308 iirc

docs/how-to/upgrade.md
Thanhphan1147: I mentioned below that eventhough we shipped app mode in rev 280, it's not until rev 308 that the unit-scoped secret issue was fixed, so I'd probably update this to say <308

docs/how-to/upgrade.md
swetha1654: @Thanhphan1147  But a new CSR would require resigning correct?

docs/how-to/upgrade.md
Thanhphan1147: updated

docs/how-to/upgrade.md
Thanhphan1147: added!

docs/how-to/upgrade.md
Thanhphan1147: discussed internally, it won't require a resigning

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
 juju refresh traefik-k8s
 ```
+
+## Revision-specific upgrade notes
+
+Some revisions require additional manual steps after upgrading. Check the section that applies to your current revision before refreshing.
+
+(upgrade_to_rev308)=
+
+### Upgrading to revision 308 or later
```
erinecon:
```suggestion
### Upgrade to revision 308 or later
```

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+Revision 308 officially switch the certificates management in Traefik from UNIT mode to APP mode. If your deployment was running any revision before 308, follow these steps.
```
erinecon:
```suggestion
Revision 308 officially switches the certificates management in Traefik from UNIT mode to APP mode. If your deployment uses any revision before 308, follow these steps to preserve your TLS certificates.

This guide assumes that you have [`jhack`](https://snapcraft.io/jhack) installed.
```
Nits, I think we should be consistent in the tense that we use (in this case, I opted for present tense). I also think that we should be more explicit about what the steps will achieve.

Finally I added a note about `jhack`

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+#### Important: Preserving TLS certificates
```
erinecon:
```suggestion
#### Preserve TLS certificates
```
I'm not totally opposed to including `Important:` in the section header, but I do want to avoid using gerunds so that the header is more action-oriented.

If you think that we should include `Important` in the header, let's discuss more! I would love to understand why this section would benefit from including an `Important` flag.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+By default, `juju refresh` to a revision that uses APP mode will cause Traefik to generate a new private key, which triggers a new Certificate Signing Request (CSR). If you are using `manual-tls-certificates` or `notary` as your TLS provider, this means you will need to sign and provide a new certificate.
```
erinecon:
```suggestion
By default, upgrading to a revision that uses APP mode will cause Traefik to generate a new private key, which triggers a new Certificate Signing Request (CSR). If you use `manual-tls-certificates` or `notary` as your TLS provider, this means you will need to sign and provide a new certificate.
```
nits

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+To avoid this, you can restore the original private key immediately after the refresh, so that Traefik re-uses the same key and produces the same CSR that was already signed. This allows you to provide the original certificate without any re-signing.
```
erinecon:
```suggestion
To avoid providing a new certificate, you can restore the original private key immediately after the refresh, so that Traefik re-uses the same key and produces the same CSR that was already signed. This allows you to provide the original certificate without any re-signing.
```
nit, to avoid a lone "this"

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+Before proceeding with the upgrade, save the existing private key and certificate **before** running `juju refresh`:
+
+```bash
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.key
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.crt
+```
+
+> Note: In some cases, a bug might occur during a node restart or a leader change that causes the leader unit to wipe out the certificates information in
+`/opt/traefik/juju`, you can look at other units to see if they still contain the original certificate and private key. In an HA deployment, you can also get them by looking at the `peer` relation data.
```
erinecon: Let's use an admonition for this note (if you use GitHub's auto-incorporation feature, double check that the formatting is correct):
````suggestion
```{note}
In some cases, a bug might occur during a node restart or a leader change that causes the leader unit to wipe out the certificates information in
`/opt/traefik/juju`. Look at other units to see if they still contain the original certificate and private key. In an HA deployment, you can also get them by looking at the `peer` relation data.
```
````
I also added a nit to split up the longer sentence into two.

File: docs/how-to/upgrade.md [hunk 17]
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+#### Steps
```
erinecon:
```suggestion
#### Update secrets
```
I prefer to avoid meta-documentation like "Steps" in favor of section headers that explicitly tell the user what actions they'll perform. Let me know if you think we should go with something other than `Update secrets`!

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+```bash
+juju secrets --format json \
+  | jq -r '.[] | select(.label | test("private-key"; "i")) | select(.owner | test("/")) | .id' \
+  | xargs -I{} juju remove-secret {}
+```
+
+After deletion, trigger a reconciliation on the leader unit so the charm creates a new application-scoped secret:
+
+```bash
+jhack fire traefik-k8s/leader config-changed
```
erinecon, line 60: AFAIK `jhack` isn't shipped with Juju, so including this command means we're assuming that the user has already installed `jhack`. That's fine but we should explicitly state this in the guide. I added this to my suggestion for updating the introduction

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+```
+
+#### Verification
```
erinecon:
```suggestion
#### Verify secrets
```

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+After either option, confirm that the private-key secret is now application-scoped:
```
erinecon:
```suggestion
Confirm that the private-key secret is now application-scoped:
```
nit for conciseness

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+```bash
+juju secrets --format json \
+  | jq '.[] | select(.label | test("private-key"; "i")) | {id, label, owner}'
+```
+
+The `owner` field should show the application name (e.g. `traefik-k8s`) rather than a unit name. The charm will use this shared key for all certificate signing requests going forward, and leader changes will no longer trigger unnecessary certificate renewals.
+
+## Preserving TLS certificates after upgrade
```
erinecon:
```suggestion
## Preserve TLS certificates after upgrade
```

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+Assuming that you've done the necessary steps to backup the private key and certificate mentioned above, you should be able to now restore the private key on the leader unit and "re-provide" the original signed certificates.
```
erinecon:
```suggestion
Assuming that you've backed up the private key and certificate, you should be able to now restore the private key on the leader unit and re-provide the original signed certificates.
```
nits

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+### Steps
+
+**1. Restore the private key on the leader unit.**
```
erinecon:
```suggestion
**1. Restore the private key on the leader unit.**
```
or
```suggestion
### Restore the private key on the leader unit.
```

Either way, I would like to avoid meta-documentation with the `Steps` header. If you prefer the second suggestion (promoting the `Restore the private key...` text as a header), then please do the same promotion for the other steps.

erinecon (commented):
Thank you for adding more context and information about upgrading!

seb4stien (approved):

docs/how-to/upgrade.md
swetha1654: how would they know which unit has the right private key? If they have 2 other units with different private key which one would they pick?

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+By default, `juju refresh` to a revision that uses APP mode will cause Traefik to generate a new private key, which triggers a new Certificate Signing Request (CSR). If you are using `manual-tls-certificates` or `notary` as your TLS provider, this means you will need to sign and provide a new certificate.
+
+To avoid this, you can restore the original private key immediately after the refresh, so that Traefik re-uses the same key and produces the same CSR that was already signed. This allows you to provide the original certificate without any re-signing.
+
+Before proceeding with the upgrade, save the existing private key and certificate **before** running `juju refresh`:
+
+```bash
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.key
+juju ssh --container traefik traefik-k8s/0 cat /opt/traefik/juju/<hostname>.crt
+```
```
swetha1654, line 34: instead of 0 i think we should specify the leader. The 0th unit might not be the right private key.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,105 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+
+> Note: In some cases, a bug might occur during a node restart or a leader change that causes the leader unit to wipe out the certificates information in
+`/opt/traefik/juju`, you can look at other units to see if they still contain the original certificate and private key. In an HA deployment, you can also get them by looking at the `peer` relation data.
+
+#### Steps
+
+After running `juju refresh traefik-k8s --revision 308` (or later), check whether stale unit-scoped secrets remain:
```
swetha1654, line 43:
```
swetha.swaminathan@canonical.com@zbook:~/Canonical/hrms-operator$(feat/frappe-hrms-charm)juju refresh traefik-k8s --revision 308
ERROR cannot upgrade from single base "ubuntu@20.04" charm to a charm supporting ["ubuntu@26.04"]. Use --force-series to override.
```
This fails and sunbeam is using amd64 so maybe we should lose the revision number

docs/how-to/upgrade.md
swetha1654, line 60: with my changes i expect the charm to go into ErrorState, if thats the case they can just run juju resolve and a new private key and CSR will be created.

docs/how-to/upgrade.md
swetha1654: it doesn't require resigning. As long as the private key and cert are consistent, manual tls charm can just feed the old certificate back to traefik with the new CSR.

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,107 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
...
+```bash
+juju secrets --format json | jq '.[] | select(.label | test("private-key"; "i"))'
+```
+
+If the output shows secrets with unit owners (e.g. `traefik-k8s/0`, `traefik-k8s/1`), the stale secrets are present and must be cleaned up. Proceed with listing and deleting each unit-scoped private-key secret:
+
+```bash
+juju secrets --format json \
+  | jq -r '.[] | select(.label | test("private-key"; "i")) | select(.owner | test("/")) | .id' \
+  | xargs -I{} juju remove-secret {}
```
swetha1654, line 54: btw this doesn't work. ID is not a field in the json, but rather they key itself:
```
swetha.swaminathan@canonical.com@zbook:~$juju secrets --format json \
  | jq -r '.[] | select(.label | test("private-key"; "i")) | select(.owner | test("/")) | .id' \
  | xargs -I{} juju remove-secret {}
ERROR secret "null" not found
ERROR secret "null" not found
ERROR secret "null" not found
swetha.swaminathan@canonical.com@zbook:~$juju secrets --format json | jq
{
  "122sdmbq59kfqvn2b750": {
    "revision": 1,
    "owner": "traefik/2",
    "label": "afd8c2bccf834997afce12c2706d2ede-private-key-certificates",
    "created": "2026-07-06T07:47:28Z",
    "updated": "2026-07-06T07:47:28Z"
  },
  "2j6il33nfla5ppnjaic0": {
    "revision": 1,
    "owner": "traefik/0",
    "label": "afd8c2bccf834997afce12c2706d2ede-private-key-certificates",
    "created": "2026-07-06T07:47:23Z",
    "updated": "2026-07-06T07:47:23Z"
  },
  "9ele4veiueitmufq2t9g": {
    "revision": 1,
    "expires": "2027-05-30T19:47:59Z",
    "owner": "traefik/0",
    "label": "afd8c2bccf834997afce12c2706d2ede-certificate-eaf987838610f9ee6358fafe07da7f799ae48ee4653ca48cd268dab32338b2c6",
    "created": "2026-07-06T07:48:11Z",
    "updated": "2026-07-06T07:48:11Z"
  },
  "9if672umgk0qdcb4m7og": {
    "revision": 1,
    "owner": "traefik",
    "label": "tls-key",
    "created": "2026-07-06T07:48:11Z",
    "updated": "2026-07-06T07:48:11Z"
  },
  "f7uoilbpb55duvn4c8fg": {
    "revision": 1,
    "owner": "traefik/1",
    "label": "afd8c2bccf834997afce12c2706d2ede-private-key-certificates",
    "created": "2026-07-06T07:47:29Z",
    "updated": "2026-07-06T07:47:29Z"
  }
}
swetha.swaminathan@canonical.com@zbook:~$juju secrets --format json \
| jq -r '
    to_entries[]
    | select(.value.label | test("private-key"; "i"))
    | select(.value.owner | test("/"))
    | .key
'
122sdmbq59kfqvn2b750
2j6il33nfla5ppnjaic0
f7uoilbpb55duvn4c8fg
```

Thanhphan1147: @copilot update the command to use `.key` instead if `.id`

docs/how-to/upgrade.md
```diff
@@ -7,3 +7,107 @@ The Traefik charm has a stateless workload. It can safely be upgraded through th
 ```
 juju refresh traefik-k8s
 ```
+
+## Revision-specific upgrade notes
```
erinecon:
```suggestion
## Revision-specific upgrades
```
nit

Thanhphan1147:
closing as no longer relevant because the newest revision already address this automatically

PR #729, opened 2026-07-03
Handle TLS private key and CSR migration

swetha1654:
#### What this PR does

 Reworked the TLS private key handling so that an existing private key, stored in a Juju secret by an earlier revision, is reused on upgrade instead of generating a new one. This prevents an unnecessary new CSR and certificate re-issuance, supporting both the app-owned `tls-key` secret (rev >= 281) and the legacy per-unit secret (rev <= 280).

Things to note:
1. I intentionally don't clear the unit scoped secrets and the unit databag CSRs, they are harmless and we can leave it there. 
 
#### Checklist


src/charm.py
```diff
@@ -369,6 +377,131 @@ def cleanup_tls_configuration(self, _: EventBase) -> None:
...
+
+    def _migrate_unit_csrs_to_app_databag(self) -> None:
+        """Migrate CSRs from unit relation data (Mode.UNIT) to app relation data (Mode.APP)."""
+        if not self.unit.is_leader():
+            return
+        relation = self.model.get_relation(CERTIFICATES_RELATION_NAME)
+        if not relation:
+            return
+        app_data = relation.data[self.app]
+        if app_data.get(CSR_DATABAG_KEY):
```
javierdelapuente, line 490: I wonder if there are unseen things here. What happens if the relation for example is removed and reintegrated? This will function retrigger forever in every hook (before any other thing)

src/charm.py
```diff
@@ -369,6 +377,131 @@ def cleanup_tls_configuration(self, _: EventBase) -> None:
...
+        try:
+            private_key = PrivateKey.from_string(raw_key)
+        except (ValueError, TypeError) as exc:
+            raise RuntimeError(
+                "Found an existing TLS private key in a Juju secret but it is invalid. "
+                "The secret is likely corrupt and requires manual intervention."
+            ) from exc
+
+        if not private_key.is_valid():
+            raise RuntimeError(
```
f-atwi, line 409: should we have more specific errors here?

src/charm.py
```diff
@@ -1227,6 +1374,10 @@ def _process_status_and_configurations(self) -> None:
             self.unit.status = WaitingStatus(f"waiting for service: '{self.traefik.service_name}'")
             return
 
+        if self._certificates_pending():
+            self.unit.status = BlockedStatus("Certificate not available yet")
+            return
```
javierdelapuente: Returning here is a functionality change?

Will other ingress be "not updated" because of one of the certs?

tests/integration/helpers.py
```diff
@@ -3,10 +3,314 @@
...
+ca_cert: Optional[x509.Certificate] = None
+ca_cert_path: Optional[Path] = None
 
 
 def all_settled(status: jubilant.Status) -> bool:
     """Return True when all apps are active and all agents are idle."""
     return jubilant.all_active(status) and jubilant.all_agents_idle(status)
 
+
+def generate_ca(tmp_path: Path) -> None:
```
javierdelapuente, line 63: you can use the tls_certificates.py library for all these kind of functions, just as the library does.

src/charm.py
seb4stien: I would at least have a slightly different message from the previous one to be sure to know where the error is coming from.

tests/integration/helpers.py
```diff
@@ -3,10 +3,314 @@
...
+        return len(leaders) == 1 and leaders[0] != old_leader
+
+    try:
+        juju.wait(_reelected, timeout=120, delay=5)
+    except TimeoutError as exc:
+        raise AssertionError(
+            f"leadership did not move away from {old_leader} within 2 minutes"
+        ) from exc
+    new_leader = leader_unit_name(juju, app)
+    logger.info("Leadership moved from %s to %s", old_leader, new_leader)
```
javierdelapuente, line 303: shouldn't we want to restart the checks and the original container here? Otherwise that unit is not receiving events, and is in a very unnatural state.

src/charm.py
```diff
@@ -661,6 +794,20 @@ def _is_tls_enabled(self) -> bool:
             return True
         return False
 
+    def _certificates_pending(self) -> bool:
+        """Return True when a CSR has been sent but no certificate is available yet."""
+        if not self.unit.is_leader():
+            return False
```
Thanhphan1147, line 800: This means that non-leader unit will always block? should we check the peer relation data?

Thanhphan1147: my bad it's the other way around

src/charm.py
```diff
@@ -369,6 +377,131 @@ def cleanup_tls_configuration(self, _: EventBase) -> None:
...
+            ) from exc
+
+        if not private_keys:
+            raise RuntimeError(
+                "The 'tls-key' secret has an empty 'private-keys' map; "
+                "it is likely corrupt and requires manual intervention."
+            )
+
+        # Exclude the user-provided config key stored under 'local-config'; every
+        # remaining entry is the (identical) library-managed private key.
```
seb4stien, line 451: Just to be sure: the private key is the same one for all hosts? Is this behavior the same for "old" versions we have to handle?

Thanhphan1147 (approved):
First pass done, looks good but I didn't look at the tests

src/charm.py
swetha1654: @f-atwi by specific errors if you mean custom exceptions, i don't see the value of custom exceptions here. @seb4stien I have updated the error messages to be different.

src/charm.py
swetha1654: The `tls-key` secret was introduced after rev 280. So for <= 280 we have the fallback function `_read_key_from_unit_secret`. By "old" if you mean the very old version of traefik that uses CertHandler, I'm not sure how things are handled there, i need to see.

src/charm.py
swetha1654: If we remove the relation this function will just return without doing anything. If we reintegrate, then this function will return nothing again since the new CSR has not been created yet.
This function will indeed retrigger in every hook forever but the other alternative is to put it in the upgrade hook, and there's no guarantee the upgrade hook runs immediately after refresh. (if we have deferred events and so on)

tests/integration/helpers.py
swetha1654: I initially thought we can still test what we want since we have 2 units left, but this makes sense, I'll update it.

tests/integration/test_upgrade_ssc_single_unit_from_298.py
```diff
@@ -0,0 +1,39 @@
...
+        revision=SOURCE_REVISION,
+        trust=True,
+    )
+    url = bring_up_self_signed_traefik(juju, tmp_path)
+    unit_name = next(iter(juju.status().apps[TRAEFIK_APP_NAME].units))
+
+    juju.refresh(TRAEFIK_APP_NAME, path=traefik_charm, resources=TRAEFIK_RESOURCES)
+    juju.wait(all_settled, delay=5, timeout=900)
+
+    verify_https_on_unit(juju, unit_name, url)
```
seb4stien, line 46: Should we ensure that Traefik is now with the right version?
(comment most likely apply to all tests)

seb4stien: Said differently: ensure the refresh was successful.

swetha1654: @seb4stien these are the tests for self signed certificates, they auto heal, I'm not sure what you mean by the right version. And the test checks if all the charms are active and idle.

tests/integration/test_upgrade_self_signed_leader_change_from_298.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    for unit_name in surviving_units:
+        verify_https_on_unit(juju, unit_name, alertmanager_url)
+
+    juju.refresh(TRAEFIK_APP_NAME, path=traefik_charm, resources=TRAEFIK_RESOURCES)
+    juju.wait(jubilant.all_agents_idle, timeout=900, delay=5, successes=5)
+
+    for unit_name in surviving_units:
+        verify_https_on_unit(juju, unit_name, alertmanager_url)
```
seb4stien: Aren't we missing a test on "old-leader"?

tests/integration/test_upgrade_mtls_from_280.py
```diff
@@ -0,0 +1,67 @@
...
+        trust=True,
+    )
+    url = bring_up_certified_traefik(juju, tmp_path)
+
+    # Upgrade to the charm under test.
+    juju.refresh(TRAEFIK_APP_NAME, path=traefik_charm, resources=TRAEFIK_RESOURCES)
+    juju.wait(all_settled, delay=5, timeout=900)
+
+    # The migrated key must still match the certificate on every unit ...
+    verify_https_on_all_units(juju, expected_url=url)
```
javierdelapuente, line 65: assert_?

tests/integration/test_upgrade_ssc_from_280_via_298.py
```diff
@@ -0,0 +1,55 @@
...
+        revision=SOURCE_REVISION,
+        num_units=NUM_TRAEFIK_UNITS,
+        trust=True,
+    )
+    url = bring_up_self_signed_traefik(juju, tmp_path)
+
+    juju.refresh(TRAEFIK_APP_NAME, channel=SOURCE_CHANNEL, revision=INTERMEDIATE_REVISION)
+    juju.wait(all_settled, delay=5, timeout=900)
+
+    verify_https_on_all_units(juju, expected_url=url)
```
seb4stien, line 57: + verify we are on the right version

tests/integration/test_upgrade_ssc_from_280_via_298.py
```diff
@@ -0,0 +1,55 @@
...
+
+    juju.refresh(TRAEFIK_APP_NAME, path=traefik_charm, resources=TRAEFIK_RESOURCES)
+    juju.wait(all_settled, delay=5, timeout=900)
+
```
seb4stien, line 62: + verify we are on the right version

tests/integration/test_upgrade_no_tls_leader_change_from_298.py
```diff
@@ -0,0 +1,83 @@
...
+        ),
+        timeout=900,
+        delay=5,
+    )
+    for unit_name in surviving_units:
+        verify_http_on_unit(juju, unit_name, alertmanager_url)
+
+    juju.refresh(TRAEFIK_APP_NAME, path=traefik_charm, resources=TRAEFIK_RESOURCES)
+    juju.wait(all_settled, timeout=900, delay=5)
+    for unit_name in surviving_units:
```
seb4stien: Aren't we missing a check on old-leader?

tests/integration/helpers.py
```diff
@@ -3,10 +3,390 @@
...
+
+def verify_https_on_unit(juju: jubilant.Juju, unit_name: str, alertmanager_url: str) -> None:
+    """Assert HTTPS returns 200 with the CA cert on a specific traefik unit."""
+    unit_ip = _unit_address(juju, unit_name)
+    logger.info("Verifying HTTPS on %s (%s) -> %s", unit_name, unit_ip, alertmanager_url)
+    session = requests.Session()
+    session.mount("https://", DNSResolverHTTPSAdapter(MOCK_HOSTNAME, unit_ip))
+    session.verify = str(ca_cert_path)
+    response = session.get(alertmanager_url, timeout=30)
+    response.raise_for_status()
```
seb4stien, line 323: The whole testing strategy rely on this one, but I'm not to have seen a test that confirm that it raises when migrating to a known buggy version. Do we we have one?

seb4stien (commented):
Unless I missed something, we're not checking that the refresh actually worked in the test (= the revision number is the one we are expecting). Apart from that it looks quite complete. Thanks for the extensive work.

src/charm.py
swetha1654: that's a very good point, I will check this tomorrow.

tests/integration/helpers.py
swetha1654: I have another function `verify_https_broken_on_unit` below for those cases, but now that i think about it I can just collapse both the functions into 1.

tests/integration/test_upgrade_no_tls_leader_change_from_298.py
swetha1654: initially my code didn't bother bringing up the unit that was forced to become down during leader change. But i added that logic now, so this is in place.

tests/integration/test_upgrade_self_signed_leader_change_from_298.py
swetha1654: initially my code didn't bother bringing up the unit that was forced to become down during leader change. But i added that logic now, so this is in place.

tests/integration/test_upgrade_ssc_from_280_via_298.py
swetha1654: @seb4stien these are the tests for self signed certificates, they auto heal, I'm not sure what you mean by the right version. And the test checks if all the charms are active and idle.

tests/integration/test_upgrade_ssc_from_280_via_298.py
swetha1654: @seb4stien these are the tests for self signed certificates, they auto heal, I'm not sure what you mean by the right version. And the test checks if all the charms are active and idle.

tests/integration/helpers.py
swetha1654: done

tests/integration/test_upgrade_mtls_from_280.py
swetha1654: technically there's no assert statement inside the function

src/charm.py
swetha1654: I moved it to the end of this function. So there will be no return statement, no functionality change. Just new Blocked status in case of missing certificates. Traefik will still function.

tests/integration/test_upgrade_ssc_from_280_via_298.py
swetha1654: Added a version check.

tests/integration/test_upgrade_ssc_from_280_via_298.py
swetha1654: Added a version check. Although i think these might be a bit redundant. We can remove them later once we are comfortable enough with the upgrades.

tests/integration/test_upgrade_ssc_single_unit_from_298.py
swetha1654: Added a version check. Although i think these might be a bit redundant. We can remove them later once we are comfortable enough with the upgrades.

src/charm.py
swetha1654: I'm not sure what the exact "legacy" revision used by Sunbeam is. For now I am omitting that test.

seb4stien (approved):

src/charm.py
```diff
@@ -369,6 +377,131 @@ def cleanup_tls_configuration(self, _: EventBase) -> None:
...
+        )
+        try:
+            secret = self.model.get_secret(label=unit_key_label)
+            content = secret.get_content(refresh=True)
+        except SecretNotFoundError:
+            return None
+
+        key = content.get(PRIVATE_KEY_FIELD)
+        if not key:
+            raise RuntimeError(
```
javierdelapuente, line 475: Do we really want to handle these kind of errors? 

I mean, if this really happens (which I doubt), and it is self signed certificates, no problem. And if it is manual tls, it goes to blocked (and we can print some debug logs with this info).

src/charm.py
```diff
@@ -1239,6 +1386,9 @@ def _process_status_and_configurations(self) -> None:
         # After processing all ingress relations, check if cert hostnames changed
         self._refresh_certs_if_needed()
 
+        if self._certificates_pending() and isinstance(self.unit.status, ActiveStatus):
+            self.unit.status = BlockedStatus("Certificate not available yet")
```
javierdelapuente, line 1390: Although we like this, we will probably break some integration tests or automations with this.

javierdelapuente (approved):
A couple of comments, but I see no issue in the code, so I approve

src/charm.py
swetha1654: Yeah especially if the hook ordering bug i noticed today is still there in the latest revision i expect to see some broken CIs.

src/charm.py
swetha1654: @javierdelapuente I thought about it. If its self-signed certificates, they can just delete the unit secret and run `juju resolve`. For manual-tls, i'd rather the code go into error state instead of blocked so that we avoid losing any hooks. Traefik is not holistic so its risky to have some hooks not processed properly.

src/charm.py
swetha1654: I updated the `test_https_multi_unit` to test the hook ordering bug that i noticed today and it doesn't seem to happen with the latest revision, so i think we're good here.

PR #735, opened 2026-07-08
Migrate all tests to jubilant

javierdelapuente:
#### What this PR does

Migrate all integration tests to Jubilant.

Bump opcli.

Update juju fixture (so it really uses the testing model created by concierge). 

#### Checklist

- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I added a [change artifact](https://github.com/canonical/traefik-k8s-operator/blob/main/docs/release-notes/template/_change-artifact-template.yaml) for user-relevant changes in `docs/release-notes/artifacts` according to the [contributing guidelines](https://github.com/canonical/traefik-k8s-operator/blob/main/CONTRIBUTING.md#release-notes). If no change artifact is necessary, I tagged the PR with the label `no-release-note`.


tests/integration/test_charm_health.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_relate(juju: jubilant.Juju):
+    juju.integrate(f"{HEALTH_TESTER_APP}:require-ingress", f"{TRAEFIK_APP}:ingress")
+    juju.wait(all_settled, timeout=600)
+
+
+def test_health(juju: jubilant.Juju):
+    traefik_address = get_k8s_service_address(juju.model, f"{TRAEFIK_APP}-lb")
+    assert traefik_address, "Expected a traefik load balancer address"
+    health_address = f"http://{traefik_address}/{juju.model}-{HEALTH_TESTER_APP}/health"
```
swetha1654, line 56: can we get the proxied endpoint from the juju action instead of constructing it ourselves? That way we don't need the `get_k8s_service_address` helper.

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+
+def test_disintegrate(juju: jubilant.Juju):
+    if ROOT_CA_APP not in juju.status().apps:
+        return
+    juju.remove_relation(f"{ROOT_CA_APP}:certificates", f"{TRAEFIK_APP}:certificates")
+    juju.wait(all_settled, timeout=600)
+
+
+def test_cleanup(juju: jubilant.Juju):
+    delete_k8s_service(juju.model, f"{TRAEFIK_APP}-lb")
```
swetha1654: deleting the app should automatically delete the LB. We don't need to explicitly delete that seperately. We can also remove the helper function as its not used anywhere else. We might have to call `remove_application` with force set to False so the LB is properly deleted.

tests/integration/helpers.py
```diff
@@ -48,6 +51,100 @@ def all_settled(status: jubilant.Status) -> bool:
...
+    """Fetch a URL with retries until the expected status is returned."""
+    last_exc: Optional[Exception] = None
+    for _ in range(retries):
+        try:
+            response = requests.get(url, verify=False, allow_redirects=True, timeout=10)
+            if response.status_code == expected_status:
+                return response
+        except Exception as exc:
+            last_exc = exc
+        time.sleep(delay)
```
swetha1654: we could also use tenacity for retries instead of implementing our own retry logic.

tests/integration/test_basic_auth.py
```diff
@@ -0,0 +1,124 @@
...
+        remote_unit=f"{IPA_TESTER_APP}/0",
+        remote_endpoint="require-ingress",
+        local_unit=f"{TRAEFIK_APP}/0",
+        local_endpoint="ingress",
+    )
+    app_data = yaml.safe_load(relation["application-data"]["ingress"])
+    return app_data["url"]
+
+
+def _assert_status(
```
swetha1654, line 87: same here, we can do something like:
```
@retry(
    stop=stop_after_delay(60),
    wait=wait_fixed(2),
    retry=retry_if_exception_type((requests.exceptions.RequestException)),
    reraise=True,
)
```

tests/integration/test_charm_health.py
```diff
@@ -0,0 +1,94 @@
...
+HEALTH_TESTER_APP = "health-tester"
+
+_METADATA = yaml.safe_load(Path("./metadata.yaml").read_text(encoding="utf-8"))
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+
+def test_deployment(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
```
swetha1654, line 33: can we use the traefik_app fixture instead?

tests/integration/test_basic_auth.py
```diff
@@ -0,0 +1,124 @@
...
+TEST_AUTH_USER = r"admin:$2a$13$XOHdzKdVS4mPKT0LvOfXru4LqyLbwcEvFlssXGS3laC6d/i6cKrLS"
+
+_METADATA = yaml.safe_load(Path("./metadata.yaml").read_text(encoding="utf-8"))
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+
+def test_deployment(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
```
swetha1654, line 36: can we use the traefik_app fixture here?

tests/integration/test_charm_ipa.py
```diff
@@ -0,0 +1,159 @@
...
+# any-charm injects this code into its src/ at deploy time, replicating the
+# legacy ipa-tester charm: request ingress with known host/port and expose
+# all relation data in a single RPC call.
+#
+# We intentionally implement the v2 ingress protocol directly (rather than
+# importing our library) to avoid the pydantic dependency inside any-charm:
+#   - host in unit databag (json-encoded)
+#   - model + name + port in app databag (json-encoded)
+#   - provider writes back the URL via the "ingress" key in its app databag
+_ANY_CHARM_SRC_OVERWRITE = {
```
swetha1654: can we use the `testers/any_charm_src/ipa/any_charm.py` here? If not, can we move the contents of this into `any_charm_src`?

tests/integration/test_charm_ipa.py
```diff
@@ -0,0 +1,159 @@
...
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+# any-charm injects this code into its src/ at deploy time, replicating the
+# legacy ipa-tester charm: request ingress with known host/port and expose
+# all relation data in a single RPC call.
+#
+# We intentionally implement the v2 ingress protocol directly (rather than
+# importing our library) to avoid the pydantic dependency inside any-charm:
```
swetha1654: why not just add pydantic as a dependency when installing the any_charm instead of reimplementing things?

tests/integration/test_charm_ipa.py
```diff
@@ -0,0 +1,159 @@
...
+                    "app_data": _decode(dict(rel.data[self.app])),
+                    "unit_data": _decode(dict(rel.data[self.unit])),
+                }
+        """
+    ),
+}
+
+
+def test_deployment(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
```
swetha1654, line 38: same here

tests/integration/test_charm_ipa.py
```diff
@@ -0,0 +1,159 @@
...
+    assert_can_connect(parsed.hostname, parsed.port or 80)
+
+
+def test_relation_data_shape(juju: jubilant.Juju):
+    data = _rpc(juju, "get_relation_data")
+
+    # Provider gave back a well-formed URL pointing at the actual LB IP
+    traefik_address = get_k8s_service_address(juju.model, f"{TRAEFIK_APP}-lb")
+    assert traefik_address, "Expected a traefik load balancer address"
+    assert data["url"] == f"http://{traefik_address}/{juju.model}-{IPA_TESTER_APP}"
```
swetha1654, line 70: do we need to test this? integration tests are usually testing only functionality. Imo i think checking if the ingress URL is reachable is sufficient

tests/integration/test_charm_ipu.py
```diff
@@ -0,0 +1,148 @@
...
+IPU_TESTER_APP = "ipu-tester"
+
+_METADATA = yaml.safe_load(Path("./metadata.yaml").read_text(encoding="utf-8"))
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+
+def test_deployment(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
```
swetha1654, line 39: same here

tests/integration/test_charm_ipu.py
```diff
@@ -0,0 +1,148 @@
...
+        if (
+            relation.get("endpoint") == remote_endpoint
+            and relation.get("related-endpoint") == local_endpoint
+            and local_unit in relation.get("related-units", {})
+        ):
+            return relation
+    raise AssertionError(
+        f"No relation data for {remote_unit}:{remote_endpoint} and "
+        f"{local_unit}:{local_endpoint}"
+    )
```
swetha1654: this function is repeated in every module. Maybe we can move it to helper.py instead?

tests/integration/test_charm_route.py
```diff
@@ -0,0 +1,151 @@
...
+def test_cleanup(juju: jubilant.Juju):
+    remove_application(juju, TRAEFIK_APP, timeout=60)
+
+
+def _rpc(juju: jubilant.Juju, unit: str, method: str) -> Any:
+    raw = juju.run(unit, "rpc", params={"method": method}).results["return"]
+    return json.loads(raw)
+
+
+def _route_config_path(juju: jubilant.Juju) -> str:
```
swetha1654:
```suggestion
def _get_route_config_path(juju: jubilant.Juju) -> str:
```
nit

tests/integration/test_charm_ipu_tcp.py
```diff
@@ -0,0 +1,154 @@
...
+TCP_TESTER_APP = "tcp-tester"
+
+_METADATA = yaml.safe_load(Path("./metadata.yaml").read_text(encoding="utf-8"))
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+
+def test_deployment(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
```
swetha1654, line 38: same here

tests/integration/test_charm_ipu_tcp.py
swetha1654, line 1: can you rename this module to `test_charm_ipu_tcp`? `test_charm_tcp` feels like its testing traefik-route but its actually testing ipu

tests/integration/test_charm_ipa.py
```diff
@@ -0,0 +1,159 @@
...
+    juju.integrate(f"{IPA_TESTER_APP}:require-ingress", f"{TRAEFIK_APP}:ingress")
+    juju.wait(all_settled, timeout=600)
+
+
+def test_ipa_charm_has_ingress(juju: jubilant.Juju):
+    data = _rpc(juju, "get_relation_data")
+    url = data["url"]
+    assert url, "Expected a non-empty ingress URL"
+    parsed = urlparse(url)
+    assert_can_connect(parsed.hostname, parsed.port or 80)
```
swetha1654, line 61: assert_can_connect only checks if the TCP connection can be established. I feel we should also check if the backend is reachable via the URL for a complete check. Same of test_charm_ipu and other modules

tests/integration/test_pebble_restart_after_cert_relation_joined.py
```diff
@@ -0,0 +1,50 @@
...
+
+_METADATA = yaml.safe_load(Path("./metadata.yaml").read_text(encoding="utf-8"))
+_TRAEFIK_RESOURCES = {
+    name: val["upstream-source"] for name, val in _METADATA["resources"].items()
+}
+
+
+def test_build_and_deploy(juju: jubilant.Juju, traefik_charm):
+    juju.deploy(traefik_charm, TRAEFIK_APP, resources=_TRAEFIK_RESOURCES, trust=True)
+    juju.deploy("ch:alertmanager-k8s", ALERTMANAGER_APP, channel="2/edge", trust=True)
```
swetha1654, line 27: can we use anycharm here instead?

tests/integration/test_pebble_restart_after_cert_relation_joined.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_can_route_ingress_using_tls(juju: jubilant.Juju):
+    juju.integrate(f"{SSC_APP}:certificates", TRAEFIK_APP)
+    juju.wait(all_settled, timeout=600)
+
+    traefik_url = _external_url(juju, TRAEFIK_APP)
+    alertmanager_url = f"{traefik_url}/{juju.model}-{ALERTMANAGER_APP}"
+
+    fetch_with_retry(alertmanager_url)
+    fetch_with_retry(alertmanager_url.replace("https://", "http://"))
```
swetha1654, line 44: fetch_with_retry currently disabled verify. Shouldn't we also check if the URL is reachable with verify set to True after cert relation joins?

tests/integration/test_tcp_ipa_compatibility.py
```diff
@@ -0,0 +1,137 @@
...
+    while time.monotonic() < deadline:
+        try:
+            with socket.create_connection((host, port), timeout=10) as sock:
+                sock.sendall(payload)
+                response = sock.recv(1024)
+            assert response == payload
+            return
+        except OSError:
+            time.sleep(5)
+    raise AssertionError(f"Timed out waiting for TCP echo on {host}:{port}")
```
swetha1654: these are repeated in other modules as well

tests/integration/test_tcp_ipu_compatibility.py
swetha1654, line 1: i think we should just have one test module that does test_tcp_ipa_ipu compatibility instead 2 seperate modules

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+    assert traefik_ip, "Expected a traefik load balancer address"
+    for endpoint in _endpoints(juju.model, "http", traefik_ip):
+        response = requests.get(endpoint, timeout=30)
+        response.raise_for_status()
+
+
+@pytest.mark.xfail(
+    reason=(
+        "Failing due to #491. Remove this after #509 is implemented. "
+        "See also: https://github.com/canonical/traefik-k8s-operator/issues/521"
```
swetha1654, line 60: can we disable the xfail and close 521? seems irrelevant now

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+    juju.integrate(f"{ROOT_CA_APP}:certificates", TRAEFIK_APP)
+    juju.wait(all_settled, timeout=300)
+
+    cert_path = _pull_server_cert(juju)
+    traefik_ip = get_k8s_service_address(juju.model, f"{TRAEFIK_APP}-lb")
+    assert traefik_ip, "Expected a traefik load balancer address"
+    _assert_https_endpoints(juju, cert_path, traefik_ip)
+
+
+@pytest.mark.xfail(reason="Flaky test; cannot reproduce failure locally.")
```
swetha1654, line 76: this as well

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+    return _CERT_PATH
+
+
+def _assert_https_endpoints(juju: jubilant.Juju, cert_path: Path, traefik_ip: str) -> None:
+    session = requests.Session()
+    session.mount("https://", DNSResolverHTTPSAdapter(MOCK_HOSTNAME, traefik_ip))
+    session.verify = str(cert_path)
+    for endpoint in _endpoints(juju.model, "https", MOCK_HOSTNAME):
+        response = session.get(endpoint, timeout=30)
+        response.raise_for_status()
```
swetha1654, line 123: can we remove this and use verify_https_on_all_units instead?

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+
+def _pull_server_cert(juju: jubilant.Juju) -> Path:
+    _ARTIFACT_DIR.mkdir(exist_ok=True)
+    cert = juju.ssh(
+        f"{TRAEFIK_APP}/0",
+        "cat /opt/traefik/juju/server.cert",
+        container="traefik",
+    )
+    _CERT_PATH.write_text(cert, encoding="utf-8")
+    return _CERT_PATH
```
swetha1654, line 114: we can use pull_ssc_ca_certificate instead

tests/integration/test_tls.py
```diff
@@ -0,0 +1,125 @@
...
+    delete_k8s_service(juju.model, f"{TRAEFIK_APP}-lb")
+    remove_application(juju, TRAEFIK_APP, timeout=60)
+
+
+def _endpoints(model: str, scheme: str, netloc: str) -> list[str]:
+    return [
+        f"{scheme}://{netloc}/{model}-{PROMETHEUS_APP}-0",
+        f"{scheme}://{netloc}/{model}-{ALERTMANAGER_APP}",
+        f"{scheme}://{netloc}/{model}-{GRAFANA_APP}",
+    ]
```
swetha1654, line 103: we can run get_proxied_endpoints instead of manually constructing URLs

tests/integration/test_workload_tracing.py
```diff
@@ -0,0 +1,106 @@
...
+    tempo_host = juju.status().apps[TEMPO_APP].address
+    assert _get_traces_patiently(tempo_host)
+
+
+def _deploy_tempo_cluster(juju: jubilant.Juju) -> None:
+    juju.deploy("ch:tempo-worker-k8s", TEMPO_WORKER_APP, channel="2/edge", trust=True)
+    juju.deploy("ch:tempo-coordinator-k8s", TEMPO_APP, channel="2/edge", trust=True)
+    juju.deploy("ch:s3-integrator", S3_INTEGRATOR_APP, channel="edge")
+    juju.deploy(
+        "ch:minio",
```
swetha1654, line 52: can we use microceph instead

javierdelapuente: I would prefer to do that in a following PR (together with all the juju 4 changes, as otherwise this will become very big and different from the original tests)

tests/integration/test_tls.py
javierdelapuente: AFAIK, get_proxied_endpoints returns the hostname, not the IP of the load balancer, and so I imagine this is why it was done. If you agree, I will review this in a following PR, but there is not much we can do probably...

tests/integration/test_charm_health.py
javierdelapuente: I think this is in many other places, and it is related to having a hostname. We also need the IP in many cases, so the action is not enough. I will review the case in a future PR, but we need the LB ip someway in many of these cases.

tests/integration/test_charm_ipa.py
javierdelapuente: I agree with you, better to test the functionality than the implementation (unless it is relevant). 

However, these tests were in the original ops tests. As I am going to do another PR, I would better do these kind of changes there, so this PR does not diverge too much from the previous tests.

tests/integration/test_basic_auth.py
javierdelapuente: This is the same test as was before, and the deployment is a bit different (different juju config) than the fixture. I agree that using a fixture is a better option, but if you do not mind, I would prefer to change this test in a future PR so I do not diverge more from the original tests.

tests/integration/test_charm_ipa.py
javierdelapuente: This is the same test as was before, and the deployment is a bit different (different juju config) than the fixture. I agree that using a fixture is a better option, but if you do not mind, I would prefer to change this test in a future PR so I do not diverge more from the original tests.

tests/integration/test_charm_ipu.py
javierdelapuente: This is the same test as was before, and the deployment is a bit different (different juju config) than the fixture. I agree that using a fixture is a better option, but if you do not mind, I would prefer to change this test in a future PR so I do not diverge more from the original tests.

tests/integration/test_charm_ipu_tcp.py
javierdelapuente: This is the same test as was before, and the deployment is a bit different (different juju config) than the fixture. I agree that using a fixture is a better option, but if you do not mind, I would prefer to change this test in a future PR so I do not diverge more from the original tests.

tests/integration/test_pebble_restart_after_cert_relation_joined.py
javierdelapuente: I would rather leave that for a future PR as it would add complexity and I will not know if the failures are related to original flakiness, the "translation" or other improvements. We may need to do it anyway to tests for new archs or maybe for juju 4.

tests/integration/test_tls.py
javierdelapuente: Is it the same? Is the server.cert just the server instead of the CA?

tests/integration/test_tls.py
javierdelapuente: `verify_https_on_all_units` currently checks for _alertmanager_url and does some other checks. Although having a single function sounds more reasonable, I would rather do this in another PR to not diverge the current tests for the original ones with ops test.

tests/integration/test_tcp_ipu_compatibility.py
javierdelapuente: could we leave this for a future PR? This is a copy from the previous ops test.

tests/integration/test_tls.py
javierdelapuente: reviewing some of the last executions (before and after the migration to Jubilant):
```
============= 4 passed, 2 xfailed, 3 warnings in 269.32s (0:04:29) =============
```

they are `reliably` failing. I think it would be better to add another story to tackle all these issues with "integration tests" when appropriate instead of doing everything in this PR.

tests/integration/test_tls.py
javierdelapuente: Same as before.

tests/integration/test_pebble_restart_after_cert_relation_joined.py
javierdelapuente: I think it would be better to add another story to tackle all these issues with "integration tests" when appropriate instead of doing everything in this PR.

tests/integration/test_charm_ipa.py
javierdelapuente: I would rather leave that for a future PR if possible as it is an extension of what we had in the tests.

tests/integration/test_tls.py
swetha1654: yeah they're not, my bad

swetha1654 (approved):

PR #742, opened 2026-07-14
Improvements to integration tests

javierdelapuente:
#### What this PR does

Remove minio (minio repository is archived: https://github.com/minio/minio and license changed) and use microceph instead.
Remove usage of `juju show-unit` in integration tests. Instead use the rpc in `any-charm` to get any data.
Remove rbac and pin juju 4 to 4.0/edge, that is currently the most "reliable" version.
Adjust waits and timeouts to improve flakiness. 
Add AGENTS.md based on the /init command in copilot.

All tests work now in juju 4 (see https://github.com/canonical/traefik-k8s-operator/actions/runs/29488305728/job/87599003870?pr=742) but they are not added to not overload the number of integration tests. Once juju 4 is the default, juju 4 should be swapped with juju 3, so only a few tests run in juju 3.

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I added a [change artifact](https://github.com/canonical/traefik-k8s-operator/blob/main/docs/release-notes/template/_change-artifact-template.yaml) for user-relevant changes in `docs/release-notes/artifacts` according to the [contributing guidelines](https://github.com/canonical/traefik-k8s-operator/blob/main/CONTRIBUTING.md#release-notes). If no change artifact is necessary, I tagged the PR with the label `no-release-note`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,118 @@
+# AGENTS.md
+
+Juju Kubernetes charm (Python 3.12, ops framework) deploying/managing Traefik as an ingress
+controller. Uses `tox`/`tox-uv`/`uv`. Line length 99, Google-style docstrings.
+
+## Commands
+
+```shell
+tox -e fmt                                                  # auto-format
+tox -e lint                                                 # ruff + mypy + pylint
+tox -e unit                                                  # unit + scenario tests
+tox -e unit -- tests/unit/test_charm.py                     # single file
+tox -e unit -- tests/unit/test_charm.py::TestClass::test_x  # single test
+tox -e static                                                # pyright + lib LIBPATCH check
+tox -e interface                                             # interface tests
+tox -e integration                                           # full integration suite (needs live Juju/k8s)
+tox -e integration -- tests/integration/test_charm_route.py  # single integration test file
+charmcraft pack                                              # build the .charm artifact
+```
+
+## Architecture
+
+- `src/charm.py` — `TraefikIngressCharm(CharmBase)`: all Juju event handling, relations, TLS;
+  delegates workload ops to `Traefik`.
+- `src/traefik.py` — `Traefik`: owns filesystem paths, generates static/dynamic Traefik config,
+  manages the Pebble service.
+- `src/utils.py` — small helpers (e.g. `is_hostname`).
+- `lib/charms/traefik_k8s/` — charm libraries consumed by other charms via `charmcraft fetch-lib`:
+  `v2/ingress.py` (per-app, current), `v1/ingress.py` (per-app, legacy), `v1/ingress_per_unit.py`,
+  `v0/traefik_route.py` (raw passthrough).
+
+### Design decisions
+
+- **File provider config**: Traefik is configured entirely via YAML files on disk (its
+  [File provider](https://doc.traefik.io/traefik/providers/file/)). Dynamic per-relation configs go
+  to `/opt/traefik/juju/juju_ingress_*.yaml` (a Juju-managed mounted volume, not the container FS)
+  because `inotify` doesn't work in the upstream busybox-based Traefik image — hence the charm uses
+  a custom Ubuntu-based image (`docker.io/ubuntu/traefik`). Static config is at
+  `/etc/traefik/traefik.yaml`.
+- **`flush_dynamic_configs`**: tar-archive push in production; monkey-patched to individual
+  `container.push()` calls in tests.
+- **Routing modes**: `path` (default) or `subdomain` (`external_hostname` must be a DNS name, not
+  an IP, in subdomain mode).
+- **Leader-only write-back**: only the leader writes the externally-reachable URL to relation data
+  (followers can't write the app databag).
+- **Lightkube** does direct k8s API calls (e.g. LoadBalancer service management); always mocked in
+  tests.
+
+## Unit/Scenario Tests
+
+- `tests/unit/` — mix of `ops.testing.Harness` (legacy) and `ops-scenario`.
+- `tests/scenario/` — exclusively `ops-scenario` (`ops-scenario~=6.0`).
+- Both run under `tox -e unit`.
+- Pattern:
+  ```python
+  from scenario import Context, State, Container, Relation
+
+  def test_something(traefik_ctx, traefik_container):
+      state = State(leader=True, containers=[traefik_container])
+      state_out = traefik_ctx.run(traefik_container.pebble_ready_event, state)
+      assert state_out.unit_status.name == "active"
+  ```
+- Shared fixtures (`tests/scenario/conftest.py`, `tests/unit/conftest.py`): `traefik_charm` (patches
+  lightkube + `_get_loadbalancer_status`), `traefik_ctx`, `traefik_container` (mounts at `/opt/` and
+  `/etc/traefik/`), `mock_lightkube_client` (autouse), `_mock_flush_dynamic_configs` (autouse).
+- Harness gotcha: `cached_property` on the charm is per-event in production but persists across
+  events in `Harness`. Call `_clear_cached_properties(harness.charm)` after config changes.
+
+## Integration Tests
+
+Use `jubilant` (not `pytest-operator`) against a real Juju/k8s model. CI orchestration uses `opcli`
+(from [`canonical/charm-ci`](https://github.com/canonical/charm-ci)) and `spread`.
+
+- **Run locally without spread** (see charm-ci's
+  [local-testing guide](https://github.com/canonical/charm-ci#local-testing-without-spread)):
+  ```shell
+  opcli env provision -c concierge-juju3.yaml   # or concierge-juju4.yaml for Juju 4
+  opcli pytest run --suite tests/integration/ -- tests/integration/test_charm_route.py \
+    -x --no-juju-teardown --juju-model testing
+  ```
+  `--no-juju-teardown` keeps the model up for follow-up runs. Without a matching
+  `artifacts.build.yaml`, the `traefik_charm` fixture needs a locally-built `.charm`
+  (`charmcraft pack`).
+- **`spread.yaml`** defines two parallel CI suites over the same auto-discovered modules:
+  `tests/integration/` (Juju 3, `concierge-juju3.yaml`) and `tests/integration/juju4/` (Juju 4,
+  `concierge-juju4.yaml`). Per-module `BASE/<module>`/`MODULE/<job>`/`CONCIERGE/<job>` overrides pin
+  specific tests to a different base or Juju channel (e.g. mTLS/SSC upgrade-from-Charmhub tests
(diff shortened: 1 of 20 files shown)


tests/integration/testers/any_charm_src/tcp_ipu/any_charm.py
```diff
@@ -64,3 +65,28 @@ def _ipu_created(self, _event):
     def get_tcp_ingress_data(self):
         """Return ingress URL info for this unit. Callable via rpc action."""
         return {"url": self.ipu.url, "urls": self.ipu.urls or {}}
+
+    def get_relation_data(self):
+        """Return the raw relation databag contents. Callable via rpc action.
```
swetha1654, line 70: it seems like we already have get_tcp_ingress_data above? Can't we just use that?

tests/integration/testers/any_charm_src/ipu/any_charm.py
```diff
@@ -22,3 +24,28 @@ def __init__(self, *args, **kwargs):
     def get_ingress_data(self):
         """Return ingress URL info for this unit. Callable via rpc action."""
         return {"url": self.ipu.url, "urls": self.ipu.urls or {}}
+
+    def get_relation_data(self):
+        """Return the raw relation databag contents. Callable via rpc action.
```
swetha1654, line 29: it seems like we already have get_ingress_data above? Can't we just use that?

s3-installation.sh
```diff
@@ -0,0 +1,47 @@
...
+log "RGW enabled."
+
+log "Waiting for RGW endpoint to respond..."
+curl --connect-timeout 2 --max-time 5 --retry 30 --retry-delay 4 --retry-all-errors -s -o /dev/null \
+    http://127.0.0.1:7480
+log "RGW endpoint is responding."
+
+log "Creating ci-user..."
+sudo microceph.radosgw-admin user create --uid ci-user --display-name "CI User" --access-key "${S3_ACCESS_KEY}" --secret-key "${S3_SECRET_KEY}"
+log "ci-user created."
```
swetha1654, line 42: it would be nice to be able to do this via concierge but sadly idt it has support for microceph

spread.yaml
```diff
@@ -73,3 +68,17 @@ integration-suites:
...
+    - integration-test
+    systems:
+    - ubuntu-24.04
+    auto-discover: false
+    pytest-arguments-template: '-x --log-format="%(asctime)s %(levelname)s %(message)s" --no-juju-teardown --juju-model testing --base {{ env.get("BASE", "ubuntu@26.04") }}'
+    environment:
+      CONCIERGE: concierge-juju4.yaml
+      MODULE/jubilant_test_dynamic_configs_juju4: tests/integration/test_dynamic_configs.py
+      MODULE/jubilant_test_https_multi_unit_juju4: tests/integration/test_https_multi_unit.py
+      MODULE/jubilant_test_receive_ca_cert_status_juju4: tests/integration/test_receive_ca_cert_status.py
```
swetha1654, line 87: why can't we just make juju4 the default one for all the tests and just have a few on juju 3 instead?

AGENTS.md
```diff
@@ -0,0 +1,118 @@
...
+- **Routing modes**: `path` (default) or `subdomain` (`external_hostname` must be a DNS name, not
+  an IP, in subdomain mode).
+- **Leader-only write-back**: only the leader writes the externally-reachable URL to relation data
+  (followers can't write the app databag).
+- **Lightkube** does direct k8s API calls (e.g. LoadBalancer service management); always mocked in
+  tests.
+
+## Unit/Scenario Tests
+
+- `tests/unit/` — mix of `ops.testing.Harness` (legacy) and `ops-scenario`.
```
swetha1654: I would add a note here. Any new unit tests must be written only using `ops-scenario`.

AGENTS.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Integration Tests
+
+Use `jubilant` (not `pytest-operator`) against a real Juju/k8s model. CI orchestration uses `opcli`
+(from [`canonical/charm-ci`](https://github.com/canonical/charm-ci)) and `spread`.
+
+- **Run locally without spread** (see charm-ci's
+  [local-testing guide](https://github.com/canonical/charm-ci#local-testing-without-spread)):
+  ```shell
+  opcli env provision -c concierge-juju3.yaml   # or concierge-juju4.yaml for Juju 4
```
swetha1654, line 78: i personally don't use this most of the time, and i don't want the agent to reinstall juju in different versions on my local machine.

s3-installation.sh
swetha1654, line 1: we need to specify in the contributing guide / agent.md that this script needs to be run manually if integration tests are run locally without spread

spread.yaml
```diff
@@ -9,6 +9,8 @@ backends:
     - ubuntu-24.04:
     - ubuntu-24.04-arm64:
         runner: [ubuntu-24.04-arm]
+    prepare: |
+      ${SPREAD_PATH}/s3-installation.sh
```
swetha1654: is there a way we can run the s3-installation only for specific tests? AFAIK, only one test uses it

swetha1654 (commented):
LGTM, some small comments. However I'm not sure if we want to merge this. Although  these are only changes to integration tests, Varshi mentioned yesterday that we don't want new PRs on Traefik until mid year break. Lets discuss once if we should wait before merging

tests/integration/testers/any_charm_src/tcp_ipu/any_charm.py
javierdelapuente: That would be a different test. Currently, there are several tests that "test" the onwire data, that is, the real data that the relationship contains, not the library. 

Although it is discussible whether to include those tests as integration tests, they have a value, if the library looks to work correctly, but the onwire data changes, then we are backward incompatible.

tests/integration/testers/any_charm_src/ipu/any_charm.py
javierdelapuente: Same comment as above.

s3-installation.sh
javierdelapuente: There is an open issue in concierge: https://github.com/canonical/concierge/issues/60

If it gets done, we can change to that.

spread.yaml
javierdelapuente: The reason I haven't done it is flakiness. Juju 4 only works "well" in the edge version, and is flakier than juju 3(even the juju bootstrap fails randomly). Then change to do is a 3 lines change, I would wait for a juju 4.1 stable working before having our pipeline flakier.

AGENTS.md
javierdelapuente: Sure! Updated to make clearer that new tests should be written with "ops-scenario"

AGENTS.md
javierdelapuente: This is the way to provision the machine as in the pipeline. I mean, you can use concierge prepare -c <file> but it is the same... 

I think it is reasonable, and if a developer installs it manually, it is ok, but it will be a bit less than "like the pipeline".

About reinstalling it... well if you have a juju 4 test failing in the pipeline... and you want to test it locally, you will have to install juju 4...

s3-installation.sh
javierdelapuente: sure!

spread.yaml
javierdelapuente: yeah, I was considering a few options:
- The first one putting it in the prepare only in the suite. Same as here but not for 100% of the tests
- Putting some bash code to filter with the test name. This would do it, but it was ugly and coupled with the test name in some bash code.
- Create another suite, but that breaks the autodiscover unless we move the tests using s3 to another directory.

At the end I put it in the prepare of the backend. The reason is that that runs while the charm is building (probably no extra time) and as I was reliable I left it there.

javierdelapuente:
> LGTM, some small comments. However I'm not sure if we want to merge this. Although these are only changes to integration tests, Varshi mentioned yesterday that we don't want new PRs on Traefik until mid year break. Lets discuss once if we should wait before merging

I do not think that applies to this PR. There is no change to the production code and it does not really change the tests coverage either. There is even not a reason to create a stable release because of this change either.

AGENTS.md
swetha1654: Most of the times, the the integration tests failing in the CI are usually issues with the test itself instead of a juju 3/4 specific issue. I agree that running `concierge prepare` would make it more closer to the pipeline, but I'm just wondering if i can avoid the agent constantly reinstalling different juju and k8s versions on my local machine, especially since I personally have a manual step that I need to do on K8s which is currently not supported by conceirge (providing a fixed IP during `k8s bootstrap` instead of the local IP to avoid k8s not working when i change networks)

spread.yaml
swetha1654: What about something like:
1. An env variable called `PRE_TEST_SCRIPT` that is passed only to the test that requires the microceph installation in spread.yaml:
```
PRE_TEST_SCRIPT/legacy_test_workload_tracing: tests/integration/scripts/install_microceph.sh
```
Then a fixture in the integration tests that runs the bash script only if the env variable exists:
```
@pytest.fixture(scope="module", autouse=True)
def run_pre_test_script() -> None:
    """Run an optional pre-test script for the current spread variant.

    The spread runner injects variant-specific environment variables from
    spread.yaml entries in the form VARIANT_KEY/task-name.
    """
    script = os.environ.get("PRE_TEST_SCRIPT")
    if not script:
        return

    script_path = Path(script)
    if not script_path.is_absolute():
        script_path = trfk_root / script_path

    if not script_path.exists():
        raise FileNotFoundError(f"PRE_TEST_SCRIPT does not exist: {script_path}")

    logger.info("Running pre-test script: %s", script_path)
    subprocess.run(["bash", str(script_path)], check=True, cwd=trfk_root)
```

swetha1654 (approved):
Approved, except for the microceph comment

spread.yaml
javierdelapuente: I would not go that way, the reason is sudo. If you try to run that in your local machine and you do not have passwordless sudo, you will get stuck in a test asking for a password (and probably with issues because of tty). 

But I like the idea of making obvious for a test to put the pre test script in an env var. Let me try to do something in between.

AGENTS.md
javierdelapuente: Not sure I follow, this is the local testing. You just need to do it once. After that you may need to "clean up" the model if the test did not do it, but that's all.

In your case, IINM, the issue is related to certificates and k8s and your machine changing the IP of the control plane... but that's a different issue...

spread.yaml
javierdelapuente: Updated to only run on one test with the env "trick" instead of hard-coding the test name.

PR #754, opened 2026-07-22
docs: Add basic tutorial

erinecon:
#### What this PR does

Adds a basic tutorial with step-by-step instructions to achieve a basic Traefik deployment.

#### Why we need it

The current tutorial covers more advanced topics. The documentation lacks a proper entry point for users to start using the charm. 

Related ticket: ISD-5105

#### Checklist

- [X] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [X] I added or updated the documentation (if applicable)
- [X] I updated `docs/changelog.md` with user-relevant changes
- [X] I added a [change artifact](https://github.com/canonical/traefik-k8s-operator/blob/main/docs/release-notes/template/_change-artifact-template.yaml) for user-relevant changes in `docs/release-notes/artifacts` according to the [contributing guidelines](https://github.com/canonical/traefik-k8s-operator/blob/main/CONTRIBUTING.md#release-notes). If no change artifact is necessary, I tagged the PR with the label `no-release-note`.
- [N/A] I used AI to assist with preparing this PR
- [N/A] I added or updated tests as needed (unit and integration)
- [N/A] **If this PR involves a Grafana dashboard:** I added a screenshot of the dashboard
- [N/A] **If this PR involves Terraform:** `terraform fmt` passes and `tflint` reports no errors
- [N/A] **If this PR modifies charm libraries owned by this project**: I incremented the `LIBAPI` and `LIBPATCH` values

### DIFF
--- a/docs/tutorial/basic-deployment.rst
+++ b/docs/tutorial/basic-deployment.rst
@@ -76,8 +76,7 @@ Set up the Juju model
 ---------------------
 
 To manage resources effectively and to separate this tutorial's workload from
-your usual work, create a new model in the controller
-using the following command:
+your usual work, create a new model in the controller with:
 
 .. code-block:: bash
 
@@ -129,10 +128,10 @@ as our application. Let's deploy it now:
 
     juju deploy flask-k8s 
 
-Integrate Traefik and the Flask app
------------------------------------
+Integrate Traefik and the Flask application
+-------------------------------------------
 
-Let's now provide the communication pathway between the Flask app
+We'll provide a communication pathway between the Flask application
 and Traefik by integrating them:
 
 .. code-block::
@@ -140,7 +139,7 @@ and Traefik by integrating them:
     juju integrate traefik-k8s flask-k8s
 
 
-We'll check what's going on with our deployment using ``juju status --relations``:
+Let's check what's going on with our deployment using ``juju status --relations``:
 
 .. terminal::
     :user: ubuntu
@@ -242,7 +241,7 @@ Visit in a browser
 ~~~~~~~~~~~~~~~~~~
 
 The HTML output from cURL can be difficult to read in a terminal.
-As a final step, let's visit the Flask application using a browser.
+As a final step, let's visit the Flask application in a browser.
 
 .. note::
 
@@ -254,10 +253,11 @@ As a final step, let's visit the Flask application using a browser.
 
         multipass info charm-tutorial-vm
 
-    Several IPs may be listed; use the first listed IP.
+    Several IPs may be listed in the output; use the first listed IP.
 
-    We also need the IP for the Traefik application listed in ``show-proxied-endpoints``.
-    In the output above, this IP is ``10.43.45.0``.
+    We also need the IP for the Traefik application listed in the output from the
+    ``show-proxied-endpoints`` action.
+    In the output shown previously, this IP is ``10.43.45.0``.
 
     Then route:
 
--- a/docs/tutorial/basic-deployment.rst
+++ b/docs/tutorial/basic-deployment.rst
@@ -204,8 +204,9 @@ Test the deployment using cURL:
 If the deployment is successful, the output should show HTML containing
 ``<title>Welcome to flask-k8s Charm</title>``.
 
-Now we can check the external URL set up by Traefik. To determine that URL,
-we need to use the charm's ``show-proxied-endpoints`` action:
+Now we can check the external URL set up by Traefik to verify that traffic
+is routed through Traefik. To determine the external URL,
+let's use the charm's ``show-proxied-endpoints`` action:
 
 .. code-block:: bash
 
@@ -277,6 +278,14 @@ application, and verified that the routing works by accessing the external URL.
 
 You can clean up your environment by following this guide:
 :ref:`Tear down your deployment <juju:tear-things-down>`
+
+.. note::
+
+    If you used Multipass for this tutorial, remove the route with:
+
+    .. code-block::
+
+        sudo ip route del 10.43.45.0
 
 Next steps
 ----------
--- a/docs/tutorial/basic-deployment.rst
+++ b/docs/tutorial/basic-deployment.rst
@@ -1,0 +1,274 @@
+.. meta::
+    :description: Follow step-by-step instructions to achieve a basic deployment of the Traefik charm.
+
+.. _tutorial_basic_deployment:
+
+Deploy the Traefik charm
+========================
+
+As part of a Kubernetes deployment, Traefik provides reverse
+proxy capabilities to ingress-requiring applications. In this tutorial,
+we'll go step-by-step through the process of deploying and integrating
+the Traefik charm to set up a reverse proxy for a Mattermost application.
+
+What you'll do
+--------------
+
+#. Deploy the Traefik charm
+#. Deploy the Mattermost charm with a database
+#. Integrate Mattermost and Traefik
+#. Inspect the routing
+
+What you'll need
+----------------
+
+You will need a working station, e.g., a laptop, with AMD64 architecture.
+Your working station should have at least 4 CPU cores, 8 GB of RAM, and 30 GB of disk space.
+
+.. tip::
+
+    You can use Multipass to create an isolated environment by running:
+
+    .. code-block::
+
+        multipass launch 24.04 --name charm-tutorial-vm --cpus 4 --memory 8G --disk 30G
+
+To be able to work inside the Multipass VM, log in with the following command:
+
+.. code-block:: bash
+
+    multipass shell charm-tutorial-vm 
+
+.. note::
+
+    If you're working locally, you don't need to do this step.
+
+This tutorial requires the following software to be installed on your working station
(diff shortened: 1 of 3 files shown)


docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,274 @@
...
+
+.. _tutorial_basic_deployment:
+
+Deploy the Traefik charm
+========================
+
+As part of a Kubernetes deployment, Traefik provides reverse
+proxy capabilities to ingress-requiring applications. In this tutorial,
+we'll go step-by-step through the process of deploying and integrating
+the Traefik charm to set up a reverse proxy for a Mattermost application.
```
swetha1654: i think we can simplify the tutorial even further by using alertmanager / flask-k8s or other ingress-requiring charms that don't have dependencies. Mattermost requires postgres.

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,274 @@
...
+* Juju 3.6+
+* Canonical Kubernetes 1.32+
+
+Use `Concierge <https://github.com/canonical/concierge>`_ to set up Juju
+and Canonical Kubernetes:
+
+.. code-block::
+
+    sudo snap install --classic concierge
+    sudo concierge prepare -p k8s
```
swetha1654: This is missing the load-balancer addon configuration.

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,274 @@
...
+
+.. code-block:: bash
+
+    MATTERMOST_IP=$(juju status --format json | jq -r '.applications."mattermost-k8s".address')
+
+Test the deployment using cURL:
+
+.. code-block:: bash
+
+    curl $MATTERMOST_IP:8080
```
swetha1654: I would recommend asking the users to get the mattermost ingress URL from either the get-proxied-endpoints action on traefik or reading the ingress URL from mattermost's relation data. Sometimes for other apps there would be a path / subdomain added.

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,274 @@
...
+.. terminal::
+    :user: ubuntu
+    :host: charm-tutorial-vm
+
+    curl $MATTERMOST_IP:8080
+
+    <a href="/traefik-tutorial-mattermost-k8s">Found</a>.
+
+Now we can check the external URL set up by Traefik. To determine that URL,
+we need to use the charm's ``show-proxied-endpoints`` action:
```
swetha1654: does this actually work without the `traefik-tutorial-mattermost-k8s` path suffix? I'm surprised. And also do we need this part? we can just ask them to curl the ingress endpoint directly.

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,274 @@
...
+Notice that Traefik has set up the URL ``http://10.43.45.0/traefik-tutorial-mattermost-k8s``
+for our Mattermost application. Once again we'll test with cURL:
+
+.. code-block::
+
+    curl http://10.43.45.0/traefik-tutorial-mattermost-k8s
+
+The terminal should show ``<a href="/traefik-tutorial-mattermost-k8s">Found</a>.``,
+just like it did when we directly tested the Mattermost application's IP address.
+
```
swetha1654, line 284: maybe we can take it one step forward and ask them to open the URL in the browser to see the expected UI?

docs/tutorial/basic-deployment.rst
erinecon: I tried both setups (with explicit load-balancer addon configuration in a concierge.yaml file, and the bare setup shown in the tutorial), and both setups yielded the exact same results. From my perspective it seems like the load-balancer configuration is included in the bare setup

docs/tutorial/basic-deployment.rst
erinecon: My intention in the tutorial was to show both URLs as a way to demonstrate that the traffic routing was working. Later in the tutorial the user uses get-proxied-endpoints

docs/tutorial/basic-deployment.rst
erinecon: Both URLs worked when I tested. I'm curious why this surprises you -- should the `MATTERMOST_IP` no longer work after integrating with Traefik?

My intention in the tutorial was to show both URLs as a way to demonstrate that the traffic routing was working

docs/tutorial/basic-deployment.rst
erinecon: Based on https://charmhub.io/flask-k8s/docs/tutorials-getting-started , it seems like moving to flask-k8s would be complex in a different way, as users would need to provide their own container

docs/tutorial/basic-deployment.rst
erinecon: I intentionally avoided opening in a browser because I wasn't convinced that's necessary in a Traefik tutorial

For a Mattermost tutorial, opening the URL in a browser does have value since Mattermost is not meant to be used in the CLI. But for a Traefik tutorial, I figured that a CLI approach would be sufficient to demonstrate the routing 

WDYT?

docs/tutorial/basic-deployment.rst
erinecon: If alertmanager is https://charmhub.io/alertmanager-k8s, then it doesn't seem to be a PFE-owned charm. Part of my motivation for using Mattermost was to use a charm that PFE owns 

I had not considered flask-k8s before, so it's good to know about that option. In addition to Mattermost, I also explored these options:
* Indico
* Hockeypuck
* NetBox

erinecon: I tested flask-k8s and it works without adding a container. So that's simpler than Mattermost -- thanks for the suggestion! I'll update the tutorial to use flask-k8s instead

docs/tutorial/basic-deployment.rst
erinecon: Did another test just now while switching to flask-k8s, and the concierge setup does seem to configure load balancing out of the box:

```
ubuntu@charm-tutorial-vm:~$ sudo k8s status
cluster status:           ready
control plane nodes:      10.114.45.70:6400 (voter)
high availability:        no
datastore:                etcd
network:                  enabled
dns:                      enabled at 10.152.183.246
ingress:                  disabled
load-balancer:            enabled, L2 mode
local-storage:            enabled at /var/snap/k8s/common/rawfile-storage
gateway                   enabled
```

erinecon: I will explicitly mention the requirement for load-balancing in the setup section

docs/tutorial/basic-deployment.rst
erinecon: After switching to flask-k8s, I agree with you that opening the URL in a browser is a better experience since the HTML is hard to read in the terminal. I'll keep in the "check both URLs" step because I still believe that's a good way of demonstrating that the traffic is being routed, but I'll add an additional section for visiting the Flask app

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,298 @@
...
+    We also need the IP for the Traefik application listed in the output from the
+    ``show-proxied-endpoints`` action.
+    In the output shown previously, this IP is ``10.43.45.0``.
+
+    Then route:
+
+    .. code-block::
+
+        sudo ip route add 10.43.45.0 via <Multipass VM IP>
```
swetha1654: I think we can simplify this by setting the load-balancer IP during concierge setup to be the multipass VM so we don't need to add any routes (similar to HRMS [here](https://github.com/canonical/hrms-operator/blob/main/docs/tutorial/basic-deployment.rst#what-youll-need)). If we provide the concierge file to the users directly then we don't even have to explicitly mention that load-balancer has to be enabled as the concierge file has it.

docs/tutorial/basic-deployment.rst
```diff
@@ -0,0 +1,298 @@
...
+You can clean up your environment by following this guide:
+:ref:`Tear down your deployment <juju:tear-things-down>`
+
+.. note::
+
+    If you used Multipass for this tutorial, remove the route with:
+
+    .. code-block::
+
+        sudo ip route del 10.43.45.0
```
swetha1654: this can also be avoided if we go the concierge way

swetha1654 (approved):
LGTM, just some minor comments, thanks!

docs/tutorial/basic-deployment.rst
erinecon: Thanks -- I tested the custom setup and indeed we can avoid adding more routes. I'll update the tutorial to use the custom setup! 

However, I think we should still explicitly mention the load-balancer in the setup because it's a strict requirement for using the Traefik charm

docs/tutorial/basic-deployment.rst
erinecon: Thanks for explicitly pointing it out -- I'll remove :)

PR #765, opened 2026-08-25
docs: fix ingress/traefik_route documentation issues (#600, #564, #640, #511, #512)

### LINKED ISSUE #511: Update the traefik_route documentation
The traefik_route relation allows another application to send static and dynamic configuration for traefik to use.  There's some basic guidance on how this works in the traefik_route library docstring, but this is minimal and not easy for a user to understand.  We should improve this documentation.

Some ideas for improvement:
* if the dynamic config sent over traefik_route is for http/https traffic, it should use the traefik charm's existing `web` and `websecure` entrypoints - we should document this
* if users may have multiple instances of the same application related to traefik, we should instruct them on how to avoid route name and route path collisions
* as explained in #510, the `tls` section of a traefik route should always be either 1) left blank, for HTTP, or 2) exactly `tls: {}` for HTTPS.  It should not have `certResolver` or `domains` keys, as those are for certificate generation.  We should clearly state this, as many applications like [grafana-k8s](https://github.com/canonical/grafana-k8s-operator/blob/a2770608526f2c82ddc3d225870f1a48eaa87179/src/charm.py#L298-L304) (and [even the traefik charm](charmed traefik does not allow for automatic certificate generation (because we don't connect to a certificate provider) - instead, we expect users to add a single TLS cert for the entire traefik gateway via relation or charm config.  This means that)...) have this wrong
* the traefik charm does some manipulation of the dynamic/static config provided, but we don't clearly state that anywhere.  We should define in user docs what traefik will do (eg: does it add a tls route?  etc)
* the metadata.yaml relation description might currently be correct, but it doesn't reflect how people actually use traefik_route.  we should update it

### LINKED ISSUE #512: Document how TLS works in the Traefik charm
### Enhancement Proposal

We need docs that qualitatitvely explain the different Traefik TLS configuration options.  Primarily, we need to explain how TLS is broken into two separate domains:
* upstream (external) traffic (from Client to Traefik): this is toggled based on whether the administrator provides the Traefik charm with certs.  It is set **on an all or nothing basis**, eg: either all routes get it or no routes get it.  In practice when using this, it means all clients communicating with traefik need to have a ca-cert available to them to validate the cert used by Traefik
* downstream (cluster-internal) traffic (from Traefik to the ingressed apps): this is configured **on a per app basis**.  Each app individually sets a `scheme` in its `ingress` relation: if `scheme=https`, then Traefik validates the application's cert and then **encrypts and sends** the user request to the app with TLS.  For this, Traefik must be provided with a ca-cert that lets it validate the app's certificate

So if TLS is enabled on both segments, what happens is:
* Client calls to `https://traefik.gateway/some-route`
* Traefik replies with a cert
* Client verifies the cert, encrypts the data with it, and sends to `https://traefik.gateway/some-route`
* Traefik decrypts the data
* Traefik calls to `https://some-internal-svc-serving-the-route`
* The service replies with a cert
* Traefik verifies the cert, encrypts the data with it, and sends to `https://some-internal...`

### LINKED ISSUE #564: Documentation is a little confusing
The docs contain this snippet:

```py
... # your charm's __init__(self, ...):
        self.ingress = IngressPerAppRequirer(self, host="foo.bar", port=80)
...
    def _foo(self):
        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
```

Why are there two different ports?
Why are the host parameters different?

Specifically, are the charmers expected to use only one API or the other?
Or are they expected to use both?
Or do these two lines do something completely different?

### LINKED ISSUE #600: docs: Error in Example Requirer
The docs provide the following example

```
# src/charm.py
from charms.traefik_k8s.v2.ingress import IngressPerAppRequirer, IngressReadyEvent

... # your charm's __init__(self, ...):
        self.ingress = IngressPerAppRequirer(self, host="foo.bar", port=80)
        self.framework.observe(self.ingress.ready, self._on_ingress_ready)
        self.framework.observe(self.ingress.revoked, self._on_ingress_revoked)

    def _on_ingress_ready(self, event: IngressPerAppReadyEvent):
        self.unit.status = ops.ActiveStatus(f"I have ingress at {event.url}!")

    def _on_ingress_revoked(self, _):
        self.unit.status = ops.WaitingStatus(f"I have lost my ingress URL!")

    def _foo(self):
        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
```

However, this results in the following error:

```
AttributeError: 'IngressPerAppRequirer' object has no attribute 'ready'
```

The example on the [Charmhub page](https://charmhub.io/traefik-k8s/libraries/ingress) appears correct, and the example above should be tweaked, such that the two `self.framework.observe` lines become the following (that is, `ingress` --> `ingress.on`)

```
        self.framework.observe(self.ingress.on.ready, self._on_ingress_ready)
        self.framework.observe(self.ingress.on.revoked, self._on_ingress_revoked)
```

---
*Reported from: https://documentation.ubuntu.com/traefik-k8s-charm/latest/how-to/integrate/*

### LINKED ISSUE #640: [charm-lib] Inconsistent and missing docstrings for constructor parameters in v2.ingress
### Description
There is currently a lack of documentation for the constructor parameters in the charms.traefik_k8s.v2.ingress library (e.g., IngressPerAppRequirer / IngressPerUnitRequirer). Several arguments are either missing from the docstrings entirely or lack sufficient explanation, forcing charm authors to read the library's internal implementation to understand how to use them.

#### Specific Concerns

- **Inconsistent Docstrings:** The constructor methods do not document all accepted parameters.
- **Unexplained Arguments:** Key arguments are undocumented. For example, `strip_prefix` is available as a parameter but has no explanation regarding what it does, how it interacts with Traefik's routing modes, or its default behavior.
- `traefik` expects to be deployed with `--trust`. App enters `error` state(showing its not handled well) when deployed without `--trust`.  This is not documented. 

#### Expected Behavior
The constructor docstrings should explicitly list:
- Every accepted parameter.
- The expected type.
- A brief description of what the parameter does and its impact on the Ingress relation.
- Default values (if applicable).

### Actual Behavior
Developers must guess the purpose of parameters like `strip_prefix` or dig into the underlying library code/Traefik documentation to deduce their behavior.

swetha1654:
## Summary

This PR fixes all currently open documentation-related issues in a single, cohesive change. It corrects broken/misleading code examples, fixes inconsistent library docstrings, corrects an inaccurate relation description, and adds a new explanation doc for TLS.

Fixes #600
Fixes #564
Fixes #640
Fixes #511
Fixes #512

## What changed

- **`docs/how-to/integrate.md`** (#600, #564): Fixed the `IngressPerAppRequirer` example which referenced non-existent `self.ingress.ready` / `self.ingress.revoked` attributes (and a wrong event/import name) instead of `self.ingress.on.ready` / `self.ingress.on.revoked`. Replaced the confusing inline port-42 example with a new "Updating the ingress request later on" section clarifying that `provide_ingress_requirements` **overwrites** (not adds to) the host/port passed to the constructor.
- **`lib/charms/traefik_k8s/v2/ingress.py`** and **`lib/charms/traefik_k8s/v1/ingress_per_unit.py`** (#640): Fixed inconsistent/missing docstrings, including a `listen_to="both"` vs `"all"` mismatch in `IngressPerUnitRequirer`, an undocumented `port` constructor argument in `IngressPerAppRequirer`, and missing defaults for several arguments. `LIBPATCH` bumped for both.
- **`lib/charms/traefik_k8s/v0/traefik_route.py`** and **`metadata.yaml`** (#511): Rewrote the library's module docstring examples — the previous raw-mode example incorrectly used `certResolver`/`domains` (Traefik never performs its own ACME resolution for charm-managed certs) and an incorrect `Host()` rule for what is actually a TCP router (`HostSNI()` is correct). Documented the auto-generated `<router>-tls` twin router and route/service name collision risk across relations. Corrected the `traefik-route` relation description in `metadata.yaml`, which incorrectly described it as "per-unit". `LIBPATCH` bumped.
- **`docs/explanation/tls.md`** (new) (#512): Added an explanation doc describing how upstream (client-to-Traefik) and downstream (Traefik-to-application) TLS are configured independently, linked from `docs/explanation/index.md`, `docs/index.md`, and `docs/reference/ingress-integration.md`.
- **`docs/changelog.md`**: Added a changelog entry summarizing all of the above.
- **`docs/release-notes/artifacts/`**: Added the required change-artifact YAML.

## Validation

- `tox -e static` — passes (pyright clean, LIBPATCH bumps verified for all 3 modified lib files).
- `tox -e lint` — passes (ruff, mypy, pylint all clean).
- Sphinx HTML docs build (`docs/.sphinx/venv/bin/sphinx-build -b html . _build/html`) — succeeds with no new warnings; new page and cross-references resolve correctly.

## Disclosure

Portions of this documentation review and fix were prepared with AI assistance (GitHub Copilot CLI), with all technical claims verified directly against the current source code before being written.


docs/how-to/integrate.md
```diff
@@ -31,21 +31,23 @@ You can immediately pass to it the host and port of the server you want ingress
...
 
     def _on_ingress_revoked(self, _):
         self.unit.status = ops.WaitingStatus(f"I have lost my ingress URL!")
+```
 
-    def _foo(self):
-        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
+```{note}
+The events are namespaced under `IngressPerAppRequirer.on` (i.e. `self.ingress.on.ready` /
+`self.ingress.on.revoked`), not directly on the requirer object.
```
swetha1654: this note is unnecessary, remove it.

docs/how-to/integrate.md
```diff
@@ -54,6 +56,27 @@ Once you have added the `ingress` library the charm would need to be re-packed w
 `traefik-k8s` and notifying the charm whenever Traefik replies with an ingress URL or
 that URL is revoked for some reason (e.g. the cloud admin removed the relation).
 
+### Updating the ingress request later on
+
+The `host` and `port` passed to the `IngressPerAppRequirer` constructor are only the *initial*
+request, used to publish ingress data as soon as the relation is available. If your charm needs
+to change what is being ingressed after init (for example, because the workload's port is only
+known once it's configured), call `provide_ingress_requirements` again with the new values at any
```
swetha1654:
```suggestion
known once it's configured), call `provide_ingress_requirements` with the values at any
```

docs/how-to/integrate.md
```diff
@@ -54,6 +56,27 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+later point (e.g. from a config-changed or pebble-ready handler):
```
swetha1654:
```suggestion
later point:
```

docs/how-to/integrate.md
```diff
@@ -54,6 +56,27 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+
+```python
+    def _on_config_changed(self, _):
+        # This replaces the (host, port) originally passed to the constructor, it does not
+        # add a second, independent ingress request.
+        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
```
swetha1654:
```suggestion
        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
```

docs/how-to/integrate.md
```diff
@@ -54,6 +56,27 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+```
+
+Each call to `provide_ingress_requirements` overwrites the previous request in the relation
+databag; there is only ever a single, current `(host, port)` pair per unit/application, not one
+per call. Use whichever of the two APIs (constructor args, or `provide_ingress_requirements`)
+matches when the host/port become known in your charm - most charms only need one or the other,
+not both.
```
swetha1654: Make this a note so users are cautious of not using both. They should use either constructor or the method, not both.

docs/index.md
```diff
@@ -36,7 +36,7 @@ Ingress is a purely in-model concern. Traefik will happily cross-model relate wi
 * - **Get started**
   - {ref}`Guided tutorial <tutorial_basic_deployment>`
 * - **Deployment**
-  - {ref}`Configure TLS termination using a local CA <how_to_tls_termination_using_a_local_ca>`
+  - {ref}`Configure TLS termination using a local CA <how_to_tls_termination_using_a_local_ca>` | {ref}`How TLS works in the Traefik charm <explanation_tls>`
```
swetha1654: I don't think a TLS explanation guide comes under the deployment section. It might be better fit under design.

metadata.yaml
```diff
@@ -59,8 +59,9 @@ provides:
   traefik-route:
     interface: traefik_route
     description: |
-      Provides endpoint for a traefik-route charm to sit between Traefik and a
-      charm in need of ingress, configuring the relation on a per-unit basis.
+      Accepts raw Traefik dynamic/static configuration from a requiring charm (such as
+      traefik-route-k8s, or any charm implementing the TraefikRouteRequirer API), one
+      configuration per relation, and merges it into this Traefik instance's own config.
```
swetha1654:
```suggestion
      Accepts raw Traefik dynamic/static configuration from an ingress requiring charm, allowing one configuration per relation, and merges it into this Traefik instance's own config after preprocessing.
```

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -748,13 +753,19 @@ def __init__(
...
+                URL, or a string if the scheme is known and stable at charm initialization
+                (defaults to always returning "http").
+
+        Note:
+            Calling `provide_ingress_requirements(host=..., ip=..., port=..., scheme=...)` at any
+            later point (e.g. from a config-changed handler) overwrites the values initially
+            passed to this constructor - it is not a second, independent request. Use whichever
+            mechanism suits your charm: pass `host`/`ip`/`port` here if they are known and static
+            at init time, or call `provide_ingress_requirements` later if they are only known
+            once the workload is configured.
```
swetha1654: this is not required, remove it. Mentioning it in the RTD documentation is enough.
```suggestion
```

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+  set together; the charm blocks if only some of them are provided.
+
+Once upstream TLS is enabled, clients that call `https://<external_hostname>/...` need the CA
+certificate available to them in order to validate the certificate Traefik presents.
+
+## Downstream TLS: Traefik to the ingressed application
+
+Downstream TLS is configured **per application**, independently of upstream TLS. Each requirer
+charm sets the `scheme` field on its `ingress`/`ingress-per-app` relation data (the
+`IngressPerAppRequirer(..., scheme="https")` constructor argument, or the `scheme` argument to
```
swetha1654:
```suggestion
charm sets the `scheme` field on its `ingress`/`ingress-per-app` relation data argument, or the `scheme` argument to
```

docs/how-to/troubleshoot-gateway-address-unavailable.md
```diff
@@ -28,6 +28,33 @@ tailor the exact steps and commands to your setup.
...
+
+Fix it by granting trust and resolving the error:
+
+```bash
+juju trust <traefik> --scope=cluster
+juju resolve <traefik>/0
+```
+
+If you deployed Traefik as part of a bundle, redeploy with `--trust` passed to the bundle deploy,
+or run `juju trust` immediately after deployment before the first `error` hook fires.
```
swetha1654: remove this
```suggestion
```

README.md
```diff
@@ -25,6 +25,11 @@ microk8s enable metallb 192.168.0.10-192.168.0.100 # You likely want change thes
 juju deploy ./traefik-k8s_ubuntu-20.04-amd64.charm traefik-ingress --trust --resource traefik-image=ghcr.io/canonical/traefik:2.10.4
 ```
 
+The `--trust` flag is required: Traefik manages its own `LoadBalancer` Kubernetes service, which
+needs cluster-scoped permissions. Deploying without `--trust` will make the unit go into `error`
+status; run `juju trust traefik-ingress --scope=cluster` and `juju resolve traefik-ingress/0` to
+recover.
+
```
swetha1654: remove this entirely. Mentioning it in the troubleshooting guide is enough.

docs/reference/ingress-integration.md
swetha1654, line 1: delete this documentation entirely. The first part (ingress related integrations is already covered in the how-to integrate guide. 
Currently the how-to integrate guide mentions only the 2 relations (ingress , ingress-per-unit)
Update that documentation to reflect traefik-route as well and move whatever content you added below over there.

docs/reference/ingress-integration.md
```diff
@@ -36,3 +36,24 @@ This charm is ideal for use cases where you need the full expressive power of
...
+- Traefik never connects to a certificate authority capable of automatically generating
+  certificates (e.g. via ACME), so a `tls` block - whether yours (in `raw` mode) or
+  auto-generated - should never contain `certResolver` or `domains` keys; those only apply when
+  Traefik is doing automatic certificate issuance for you.
+- If more than one instance of your charm (or more than one application) can relate to the same
+  Traefik over `traefik-route`, ensure your router/service names are unique across relations to
+  avoid collisions when Traefik merges everyone's configuration together.
+
+See {ref}`How TLS works in the Traefik charm <explanation_tls>` for more on how upstream and
+downstream TLS are configured in general.
```
swetha1654: this is not required. remove it.

docs/tutorial/basic-deployment.rst
```diff
@@ -127,7 +127,23 @@ Let's begin by deploying the Traefik charm:
...
+    ``Forbidden`` Kubernetes API error in ``juju debug-log``). To recover, grant trust after the
+    fact and resolve the error:
+
+    .. code-block:: bash
+
+        juju trust traefik-k8s --scope=cluster
+        juju resolve traefik-k8s/0
+
+    See :ref:`Troubleshoot "Gateway Address Unavailable" <how_to_troubleshoot_gateway_address_unavailable>`
+    for more details.
```
swetha1654: remove this entirely. The doc already mentions that we must use the trust flag.

docs/changelog.md
```diff
@@ -10,6 +10,30 @@ Each revision is versioned by the date of the revision.
...
+    (`Host` instead of `HostSNI`), and documented the `web`/`websecure` entrypoint convention,
+    route name collisions across relations, and why `certResolver`/`domains` must never be set
+    (fixes [#511](https://github.com/canonical/traefik-k8s-operator/issues/511)).
+  - Corrected the `traefik-route` relation description in `metadata.yaml`, which incorrectly
+    described the relation as being configured "on a per-unit basis"
+    (fixes [#511](https://github.com/canonical/traefik-k8s-operator/issues/511)).
+  - Added a new explanation page, "How TLS works in the Traefik charm", documenting the
+    independent upstream (client-to-Traefik) and downstream (Traefik-to-application) TLS
+    configuration domains, and linked it from the ingress integration reference and the docs
+    home page (fixes [#512](https://github.com/canonical/traefik-k8s-operator/issues/512)).
```
swetha1654:
```suggestion
- Fixed several documentation issues.
```
this is too verbose. Shorten it to above.

docs/how-to/integrate.md
```diff
@@ -54,6 +62,30 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+```python
+    def _on_config_changed(self, _):
+        # This replaces the (host, port) originally passed to the constructor, it does not
+        # add a second, independent ingress request.
+        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
+```
+
+Each call to `provide_ingress_requirements` overwrites the previous request in the relation
+databag; there is only ever a single, current `(host, port)` pair per unit/application, not one
+per call.
```
swetha1654:
```suggestion
databag; there is only ever a single, current `(host, port)` pair per unit/application.
```

docs/how-to/integrate.md
```diff
@@ -54,6 +62,30 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+
+```{note}
+Use whichever of the two APIs (constructor args, or `provide_ingress_requirements`) matches when
+the host/port become known in your charm. Most charms only need one or the other, not both.
```
swetha1654:
```suggestion
Use whichever of the two APIs: constructor arguments, or `provide_ingress_requirements`. 
Most charms only need one or the other, not both.
```

docs/how-to/integrate.md
```diff
@@ -68,3 +98,37 @@ juju run traefik/0 show-proxied-endpoints
...
+
+The [Traefik route charm](https://charmhub.io/traefik-route-k8s) is a proxy charm that sits
+between Traefik and a charm in need of ingress, and is used to provide low-level access to
+Traefik configuration, as well as to allow configuration for each relation.
+
+This charm is ideal for use cases where you need the full expressive power of
+[Traefik's routing configuration](https://doc.traefik.io/traefik/routing/overview/),
+or if you want to use a single Traefik instance to provide domain-name-based
+URL routing to some charms, but path-based URL routing to others.
+
```
swetha1654:
```suggestion
```
remove this entirely. The traefik-route-k8s charm is archived.

docs/how-to/integrate.md
```diff
@@ -68,3 +98,37 @@ juju run traefik/0 show-proxied-endpoints
...
+- Traefik never connects to a certificate authority capable of automatically generating
+  certificates (e.g. via ACME). A `tls` block - whether yours (in `raw` mode) or
+  auto-generated - should never contain a `certResolver` key; that only applies when Traefik is
+  doing automatic certificate issuance for you.
+- If more than one instance of your charm (or more than one application) can relate to the same
+  Traefik over `traefik-route`, ensure your router/service names are unique across relations to
+  avoid collisions when Traefik merges everyone's configuration together.
+
+See {ref}`How TLS works in the Traefik charm <explanation_tls>` for more on how upstream and
+downstream TLS are configured in general.
```
swetha1654:
```suggestion
```
remove this block

docs/how-to/integrate.md
```diff
@@ -68,3 +98,37 @@ juju run traefik/0 show-proxied-endpoints
...
+full usage examples):
+
+- By default, Traefik automatically generates a TLS-enabled twin of every router you declare and
+  attaches it to its own `websecure` entrypoint; attach your own router to the `web` entrypoint.
+  You should not declare your own `-tls`-suffixed routers or `tls` blocks unless you opt out of
+  this via the library's `raw=True` mode.
+- Traefik never connects to a certificate authority capable of automatically generating
+  certificates (e.g. via ACME). A `tls` block - whether yours (in `raw` mode) or
+  auto-generated - should never contain a `certResolver` key; that only applies when Traefik is
+  doing automatic certificate issuance for you.
```
swetha1654:
```suggestion
- Traefik in this charm does not request certificates from an ACME server by itself. It only uses certificates that are given to it by the charm. Because of that, your router `tls` settings should not include certResolver, since that option is only for setups where Traefik is directly obtaining certificates for you.
```

docs/.custom_wordlist.txt
```diff
@@ -1,9 +1,15 @@
 addon
 Alertmanager
+args
 BasicAuth
 bcrypt
+databag
+docstring
+docstrings
```
swetha1654: do we need to specify both? docstring is sufficient. docstrings is not required.

docs/.custom_wordlist.txt
```diff
@@ -1,9 +1,15 @@
...
 hardcoded
+ingressed
```
swetha1654: ingressed is not a word. Substitute all its usages with "ingress requirer" or "app requiring ingress"

docs/.custom_wordlist.txt
```diff
@@ -1,9 +1,15 @@
...
 middleware
+plaintext
```
swetha1654: substitute plaintext with plain text wherever its used.

docs/.custom_wordlist.txt
```diff
@@ -1,9 +1,15 @@
 addon
 Alertmanager
+args
```
swetha1654: substitute args usage with arguments

docs/redirects.txt
```diff
@@ -2,3 +2,5 @@
 # Format: source destination
 # Example: old-page.html new-page.html
 how-to/force-https-redirect.html index.html
+reference/ingress-integration.html how-to/integrate/index.html
+reference/ingress-integration.md how-to/integrate.md
```
swetha1654, line 6: i don't think the markdown redirection is required. Will it even work?

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -74,6 +74,25 @@ def __init__(self, *args):
...
+`websecure` (HTTPS) entrypoints already exist; attach your router to `web` (or omit
+`entryPoints` for HTTP, since `web` is Traefik's default). Traefik will not rewrite your
+router's `entryPoints` for you - the automatically generated `-tls` twin defaults to
+`websecure` independently of what entrypoint your own router uses.
+
+If your charm may be related to Traefik more than once, or from more than one unit/application,
+make sure your router/service names are unique across relations (e.g. by including the app name
+or relation id in the name) to avoid collisions when Traefik merges everyone's dynamic
+configuration together.
+
```
swetha1654: this is already specified in the how-to integrate guide. Remove this entirely.
```suggestion
```

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -94,15 +113,25 @@ def __init__(self, *args):
...
+In `raw=True` mode, Traefik will not add or modify any TLS router/config for you: you are
+responsible for declaring your own `web`/`websecure` (or raw TCP/UDP) entrypoints and TLS
+sections. Because the Traefik charm never connects to a certificate provider capable of
+automated certificate generation (e.g. ACME), the `tls` section of a router should be either
+omitted entirely (for plain HTTP), or exactly `tls: {}` (for HTTPS termination using the
+certificate Traefik itself was configured with, via the `certificates` relation or the
+`tls-cert`/`tls-key`/`tls-ca` config options). Do **not** set `certResolver` or `domains` under
+`tls`: those keys are only meaningful when Traefik is responsible for automatically requesting
+certificates for you, which is not supported here.
```
swetha1654:
```suggestion
```
this is already specified in the how-to integrate guide. Remove this entirely.

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -732,6 +732,11 @@ def __init__(
                 application; if unspecified, the default Kubernetes service name will be used.
             ip: Alternative addressing method other than host to be used by the ingress provider;
                 if unspecified, the binding address from the Juju network API will be used.
+            port: The port of the service to be ingressed. Required (and must be a keyword arg)
+                if `host` or `ip` is set, since the constructor publishes the ingress request
+                as soon as the relation is available; if none of `host`/`ip`/`port` are provided
+                at init time, call `provide_ingress_requirements(port=..., ...)` later instead
+                once the values are known (see the "Request Args" note below).
```
swetha1654:
```suggestion
            port: The port of the service requiring ingress.
```

docs/how-to/integrate.md
```diff
@@ -54,6 +62,48 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+the prefix only from the request Traefik forwards on to your workload, so your application sees
+the request as if it had been made against `/` instead of `/model_name-app_name/...`.
+
+- If your workload is already configured to serve under the `/model_name-app_name` prefix (for
+  example via a `--root-path`/base-path style option), leave `strip_prefix` at its default of
+  `False`.
+- If your workload only knows how to serve requests at its own root path, set
+  `strip_prefix=True` so Traefik strips the prefix before forwarding.
+- `strip_prefix` only has an effect in `path` routing mode. In `subdomain` routing mode there is
+  no path prefix to strip, so the option is a no-op.
```
swetha1654:
```suggestion
`strip_prefix` only has an effect in `path` routing mode. In `subdomain` routing mode there is
  no path prefix to strip, so the option is a no-op.
```

docs/reference/ingress-integration.md
```diff
@@ -36,3 +36,24 @@ This charm is ideal for use cases where you need the full expressive power of
...
+Over the `traefik-route` relation, the requiring charm submits raw Traefik dynamic (and
+optionally static) configuration, which Traefik merges into its own configuration. A few things
+to keep in mind when using this interface (see the
+[`traefik_route` library docstring](https://charmhub.io/traefik-k8s/libraries/traefik_route) for
+full usage examples):
+
+- By default, Traefik automatically generates a TLS-enabled twin of every router you declare and
+  attaches it to its own `websecure` entrypoint (your router is attached to `web`); you should
+  not declare your own `-tls`-suffixed routers or `tls` blocks unless you opt out of this via the
+  library's `raw=True` mode.
```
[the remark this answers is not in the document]
swetha1654: Fixed - this whole page (`docs/reference/ingress-integration.md`) was deleted per a later review comment and its content was merged into `docs/how-to/integrate.md`'s new "Use `traefik-route` for raw Traefik configuration" section, with the corrected wording (attach your own router to `web` explicitly, no automatic attachment).

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -74,6 +74,23 @@ def __init__(self, *args):
...
+automatically generate a matching `<router-name>-tls` router that serves the same rule/service
+over its `websecure` entrypoint, with `tls: {}` (or a `tls.domains` block derived from Traefik's
+own `external_hostname`, if one is configured). **Do not declare your own `-tls`-suffixed router
+or set a `tls` key on your router** - Traefik will add it for you and your own TLS
+configuration/middlewares would either conflict or be silently dropped when Traefik regenerates
+the router (see [issue #335](https://github.com/canonical/traefik-k8s-operator/issues/335)).
+
+For HTTP/HTTPS traffic, do not declare custom entrypoints. Traefik's own `web` (HTTP) and
+`websecure` (HTTPS) entrypoints are used automatically: the router you submit is attached to
+`web`, and the automatically generated `-tls` twin is attached to `websecure`.
```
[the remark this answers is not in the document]
swetha1654: Fixed in 97b252e - reworded to: "attach your router to `web` (or omit `entryPoints` for HTTP, since `web` is Traefik's default). Traefik will not rewrite your router's `entryPoints` for you...", removing the inaccurate automatic-attachment claim.

docs/how-to/integrate.md
swetha1654: Applied your suggestion.

docs/how-to/integrate.md
swetha1654: Applied your suggestion.

docs/how-to/integrate.md
swetha1654: Removed - agreed the traefik-route-k8s charm being archived makes this misleading. Kept the rest of the section since the `traefik-route` relation/interface itself is still valid (any charm can implement the requirer side, not just traefik-route-k8s).

docs/how-to/integrate.md
swetha1654: Removed this cross-reference block.

docs/how-to/integrate.md
swetha1654: Applied your suggested wording for the certResolver bullet.

docs/.custom_wordlist.txt
swetha1654: Agreed, removed the `docstrings` (plural) entry - only the singular `docstring` is actually used in the docs.

docs/.custom_wordlist.txt
swetha1654: Agreed - replaced all 3 usages (docs/explanation/tls.md x2, docs/how-to/integrate.md x1) with "app requiring ingress"/"application requiring ingress" and "what ingress is being requested for", and removed `ingressed` from the wordlist since it's no longer used anywhere.

docs/.custom_wordlist.txt
swetha1654: Agreed - replaced both usages in docs/explanation/tls.md with "plain text" and removed `plaintext` from the wordlist.

docs/.custom_wordlist.txt
swetha1654: Agreed - the one prose usage of "args" (in the note about the two APIs) is gone now that the note was reworded to "constructor arguments" per your other suggestion, so removed `args` from the wordlist. The remaining `*args`/`self.args` occurrences are all in .py code, which isn't spell-checked (vale only scans `*.{md,rst}`).

docs/redirects.txt
swetha1654: Good question - I checked this directly. It is required, and it does work, but only in a specific format.

There are two separate mechanisms both driven by this file:
1. `sphinx-rerediraffe` (via `rediraffe_redirects = "redirects.txt"` in conf.py) generates the actual HTML redirect page users hit. It needs the destination in the built `.html` path form, which is the `reference/ingress-integration.html how-to/integrate/index.html` line (confirmed locally: `sphinx-build -b dirhtml` produces `(good) reference/ingress-integration.html/index.html --> how-to/integrate/index.html`, 0 warnings).
2. The separate "Check removed URLs" CI job (`sphinx-stack/docs/_dev/check_removed_urls.py`) *also* parses `redirects.txt`, but its matcher (`source_candidates_for_url`) only accepts the source in `.md`/`/index.md`/trailing-`/` form, not `.html` form - so it does not recognize the first line as covering the removed page. I confirmed this by running the actual script's logic locally: with only the `.html` line present, `disjoint=True` (job fails); adding the `reference/ingress-integration.md how-to/integrate.md` line makes it `disjoint=False` (job passes). This matched exactly what happened in CI (job failed until I added it, then passed).

So both lines are needed for different reasons: one for the real redirect, one to satisfy this specific checker's (arguably quirky) matching logic. Happy to simplify further if you spot a cleaner way, but for now both are required for CI to pass.

lib/charms/traefik_k8s/v0/traefik_route.py
swetha1654: Removed - this is now only covered in the how-to integrate guide's traefik-route section.

lib/charms/traefik_k8s/v0/traefik_route.py
swetha1654: Removed - the certResolver/omitted-vs-tls:{} guidance is now only in the how-to integrate guide.

lib/charms/traefik_k8s/v2/ingress.py
swetha1654: Applied your suggestion.

docs/how-to/integrate.md
swetha1654: This wording was already applied in a previous commit (matches your suggestion verbatim) - marking resolved.

docs/how-to/integrate.md
```diff
@@ -68,3 +117,29 @@ juju run traefik/0 show-proxied-endpoints
 
 These are the URLs at which your workloads are externally accessible.
 
+(how_to_integrate_traefik_route)=
+
+## Use `traefik-route` for raw Traefik configuration
+
+The [Traefik route charm](https://charmhub.io/traefik-route-k8s) is a proxy charm that sits
+between Traefik and a charm in need of ingress, and is used to provide low-level access to
+Traefik configuration, as well as to allow configuration for each relation.
```
swetha1654: I see this is still here. It should be removed. Traefik-route-k8s charm is archived.
```suggestion
```

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -74,6 +74,20 @@ def __init__(self, *args):
...
+or set a `tls` key on your router** - Traefik will add it for you and your own TLS
+configuration/middlewares would either conflict or be silently dropped when Traefik regenerates
+the router (see [issue #335](https://github.com/canonical/traefik-k8s-operator/issues/335)).
+
+For HTTP/HTTPS traffic, do not declare custom entrypoints. Traefik's own `web` (HTTP) and
+`websecure` (HTTPS) entrypoints already exist; attach your router to `web` (or omit
+`entryPoints` for HTTP, since `web` is Traefik's default). Traefik will not rewrite your
+router's `entryPoints` for you - the automatically generated `-tls` twin defaults to
+`websecure` independently of what entrypoint your own router uses.
+
```
swetha1654: i see this is still there. I asked you to remove this entirely.
```suggestion
```

lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -732,6 +732,11 @@ def __init__(
                 application; if unspecified, the default Kubernetes service name will be used.
             ip: Alternative addressing method other than host to be used by the ingress provider;
                 if unspecified, the binding address from the Juju network API will be used.
+            port: The port of the service requiring ingress. Required (and must be a keyword arg)
+                if `host` or `ip` is set, since the constructor publishes the ingress request
+                as soon as the relation is available; if none of `host`/`ip`/`port` are provided
+                at init time, call `provide_ingress_requirements(port=..., ...)` later instead
+                once the values are known (see the "Request Args" note below).
```
swetha1654:
```suggestion
            port: The port of the service requiring ingress. 
```

docs/how-to/integrate.md
```diff
@@ -4,7 +4,18 @@
 
 Traefik provides ingress to other charmed applications. If a charm integrates with Traefik, it can delegate the responsibility of providing ingress to Traefik.
 
-See {ref}`Ingress-related relations <reference_ingress_integrations>` for more details on the Traefik charm ingress-related relations.
+The Traefik charm supports three ways for a charm to obtain ingress:
+
+- [`ingress`](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/ingress/v2#readme):
+  each charmed application requests a single, cluster-unique URL for ingress. You can choose
+  between a domain-name-based URL (`your.parameters.domain.com`) and a path-based URL
+  (`domain.com\your\parameters`).
```
javierdelapuente: are slashes reversed?

docs/how-to/troubleshoot-gateway-address-unavailable.md
```diff
@@ -28,6 +28,30 @@ tailor the exact steps and commands to your setup.
 - Traefik's service type is `LoadBalancer`.
 - An external IP address is assigned to Traefik.
 
+## Traefik was deployed without `--trust`
+
+Traefik manages its own `LoadBalancer` Kubernetes service, which requires elevated
+(cluster-scoped) Kubernetes permissions. If Traefik is deployed without the `--trust` flag (or
+`juju trust` is not run afterwards), the charm cannot create or update this service. Juju's
```
javierdelapuente, line 34: I have always had the doubt if it works without trust when RBAC is not enabled.

javierdelapuente (approved):

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+## Upstream TLS: client to Traefik
+
+Upstream TLS termination is configured on Traefik as a whole, on an **all-or-nothing** basis:
+either every route Traefik serves gets HTTPS, or none of them do. There is no per-route toggle
+for upstream TLS.
+
+You can provide Traefik with a certificate in one of two ways:
+
+- Relate a certificate provider charm (such as `self-signed-certificates`) over the
+  `certificates` relation. See {ref}`how_to_tls_termination_using_a_local_ca` for a worked
```
seb4stien:
```suggestion
  `certificates` relation. See {ref}`how_to_tls_termination_using_a_local_ca` for a working
```

seb4stien (approved):

docs/explanation/index.md
```diff
@@ -12,5 +12,6 @@ The pages in this section provide additional context and deeper understanding of
 
 ```{toctree}
 :maxdepth: 1
+tls.md
```
erinecon:
```suggestion
How TLS works <tls.md>
```
nit, to be more concise in the left-hand navigation menu

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Upstream TLS: client to Traefik
+
+Upstream TLS termination is configured on Traefik as a whole, on an **all-or-nothing** basis:
+either every route Traefik serves gets HTTPS, or none of them do. There is no per-route toggle
+for upstream TLS.
+
+You can provide Traefik with a certificate in one of two ways:
+
+- Relate a certificate provider charm (such as `self-signed-certificates`) over the
```
erinecon:
```suggestion
- Integrate a certificate provider charm (such as `self-signed-certificates`) over the
```

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+  `certificates` relation. See {ref}`how_to_tls_termination_using_a_local_ca` for a working
+  example.
+- Set the `tls-cert`, `tls-key`, and `tls-ca` charm config options directly. All three must be
```
erinecon:
```suggestion
- Set the `tls-cert`, `tls-key`, and `tls-ca` charm configuration options directly. All three must be
```
nit

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+## Downstream TLS: Traefik to the app requiring ingress
+
+Downstream TLS is configured **per application**, independently of upstream TLS. Each requirer
+charm sets the `scheme` field on its `ingress`/`ingress-per-app` relation data argument, or the
+`scheme` argument to `provide_ingress_requirements`. Allowed values are `http`, `https`, and `h2c`.
+
+- If `scheme=http` (the default), Traefik forwards the request to the application in plain
+  text.
+- If `scheme=https`, Traefik validates the application's own certificate and encrypts traffic
+  between itself and the application. For this to work, Traefik needs a CA certificate that lets
```
erinecon:
```suggestion
  between itself and the application. For this validation to work, Traefik needs a CA certificate that lets
```
nit, to avoid a lone "this"

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+2. Traefik replies with its own certificate (issued via the `certificates` relation, or via the
+   `tls-*` config options).
+3. The client verifies the certificate, encrypts the request, and sends it to Traefik.
+4. Traefik decrypts the request.
+5. Traefik calls `https://some-internal-svc-serving-the-route`.
+6. The application replies with its own certificate.
+7. Traefik verifies that certificate (using a CA cert it received separately), encrypts the
+   request, and sends it to the application.
+
+Because the two segments are configured independently, it is entirely possible (and common) to
```
erinecon:
```suggestion
Because the two segments are configured independently, it is entirely possible to
```
nit, for conciseness

docs/how-to/integrate.md
```diff
@@ -54,6 +62,42 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+model/app-unique path prefix for your application (e.g. `/model_name-app_name`), and the
+ingress URL handed back to clients always includes that prefix - `strip_prefix` does **not**
+change the externally-visible URL.
+
+What `strip_prefix=True` actually does is add a Traefik `stripPrefix` middleware that removes
+the prefix only from the request Traefik forwards on to your workload, so your application sees
+the request as if it had been made against `/` instead of `/model_name-app_name/...`.
+
+`strip_prefix` only has an effect in `path` routing mode. In `subdomain` routing mode there is
+  no path prefix to strip, so the option is a no-op.
```
erinecon: IMO this information doesn't belong in a how-to guide -- there are no instructions. It feels much more like reference material to me. I recommend moving this section into a separate reference page (or maybe the "Charm architecture" page if you feel like it's relevant).

docs/how-to/integrate.md
```diff
@@ -54,6 +62,42 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+
+### Updating the ingress request later on
```
erinecon:
```suggestion
### Update the ingress request later on
```
nit, to be consistent with the other headers

docs/how-to/integrate.md
```diff
@@ -54,6 +62,42 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+    def _on_config_changed(self, _):
+        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
+```
+
+Each call to `provide_ingress_requirements` overwrites the previous request in the relation
+databag; there is only ever a single, current `(host, port)` pair per unit/application.
+
+```{note}
+Use whichever of the two APIs: constructor arguments, or `provide_ingress_requirements`.
+Most charms only need one or the other, not both.
```
erinecon:
```suggestion
Use one of the two APIs: constructor arguments, or `provide_ingress_requirements`.
Most charms only need one or the other, not both.
```
What happens if you use both APIs?

docs/how-to/integrate.md
```diff
@@ -68,3 +112,25 @@ juju run traefik/0 show-proxied-endpoints
...
+  attaches it to its own `websecure` entrypoint; attach your own router to the `web` entrypoint.
+  You should not declare your own `-tls`-suffixed routers or `tls` blocks unless you opt out of
+  this via the library's `raw=True` mode.
+- Traefik in this charm does not request certificates from an ACME server by itself. It only
+  uses certificates that are given to it by the charm. Because of that, your router `tls`
+  settings should not include `certResolver`, since that option is only for setups where Traefik
+  is directly obtaining certificates for you.
+- If more than one instance of your charm (or more than one application) can relate to the same
+  Traefik over `traefik-route`, ensure your router/service names are unique across relations to
+  avoid collisions when Traefik merges everyone's configuration together.
```
erinecon: A similar comment about how this doesn't feel like a how-to guide to me. This section feels more like explanation material about the charm design. I think the information here is useful, but the majority of this section should go into an Explanation page.

But I see more instruction-type text in this section compared to the `strip_prefix` section. I support keeping a similar section in this page, but the text would need some refactoring. Some ideas:
* If you don't want to duplicate the information in the Charmhub page you linked, then highlight that page as the place where users will find further instructions. Right now the page is relegated to a side note in parentheses, but that page provides instructions on how to fetch and use the library. 
* First bullet point: The section in this guide should focus on the second sentence. 
* Second and third bullet points could be moved entirely to Explanation. The note in the third bullet point about unique router/service names could be included here as a warning.

docs/how-to/troubleshoot-gateway-address-unavailable.md
erinecon: Since you've added a dedicated section about the `trust` point, I think we should remove it from the "Requirements" section

erinecon (changes requested):
Thanks so much for resolving so many documentation issues! I have some concerns about two of the sections added to the integrate how-to guide

docs/how-to/integrate.md
swetha1654: I think a seperate reference page makes sense. And i will link it here because people need to understand how to use the attributes of the interface.

docs/how-to/integrate.md
swetha1654: Using both APIs will lead to inconsistent ingress requirements across hooks. (if some hooks don't call the `provide_ingress_requirements`), then the constructor would set new ingress requirements. I'll update the note to add this.

docs/how-to/integrate.md
swetha1654: I think it makes sense to move this content into the traefik-route library docstring and explicitly mention here that the charmhub page contains more information on what to keep in mind when using this library.
I will also refactor the points like you mentioned.

swetha1654:
@copilot address all the open review comments in this PR.

docs/how-to/troubleshoot-gateway-address-unavailable.md
swetha1654: I will test this outside of this PR. If its true then its interesting.

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,30 @@
...
+
+(reference_ingress_libraries)=
+
+# Ingress library reference
+
+This page covers selected parameters and behaviour of the ingress charm libraries provided by
+Traefik. For full API documentation, visit the Charmhub library pages:
+
+- [`ingress`](https://charmhub.io/traefik-k8s/libraries/ingress)
+- [`ingress_per_unit`](https://charmhub.io/traefik-k8s/libraries/ingress_per_unit)
```
swetha1654, line 15: add traefik-route also here

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -94,15 +94,69 @@ def __init__(self, *args):
...
+are supplied to it via the `certificates` relation or the `tls-*` config options. Do not set
+`certResolver` in your router's `tls` block; that option is only valid when Traefik itself is
+obtaining certificates via ACME.
+
+**Unique router and service names**
+
+If more than one application can relate to the same Traefik instance over `traefik-route`,
+ensure your router and service names are unique across relations. Traefik merges all submitted
+configurations together, so a name collision will cause one relation's config to silently
+overwrite another's.
```
swetha1654: move this to ingress-libraries.md and add a note here to refer to the charm documentation on further details on how the traefik-route library works.

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -94,15 +94,69 @@ def __init__(self, *args):
...
+                        'secure-route': {
+                            'rule': 'HostSNI(`secure.example.com`)',
+                            'service': 'my-service',
+                            'tls': {}
+                        }
+                    }
+                }
+            }
+        )
+```
```
swetha1654: why is this added?

docs/how-to/integrate.md
swetha1654: on the other hand it makes sense to keep it in the reference section instead.

docs/how-to/integrate.md
```diff
@@ -68,3 +104,24 @@ juju run traefik/0 show-proxied-endpoints
...
 These are the URLs at which your workloads are externally accessible.
 
+(how_to_integrate_traefik_route)=
+
+## Use `traefik-route` for raw Traefik configuration
+
+Over the `traefik-route` relation, the requiring charm submits raw Traefik dynamic (and
+optionally static) configuration, which Traefik merges into its own configuration. The
+[`traefik_route` library page on Charmhub](https://charmhub.io/traefik-k8s/libraries/traefik_route)
+contains full usage examples and further details on the library's behavior.
```
swetha1654: the library contains only examples now that we are moving the further details on library's behaviour to the reference page `ingress-libraries.md`. Update this to link the reference as well.

swetha1654:
@copilot address open review comments

docs/how-to/integrate.md
```diff
@@ -68,3 +104,25 @@ juju run traefik/0 show-proxied-endpoints
...
+
+- Attach your HTTP router to the `web` entrypoint. Traefik will not rewrite your router's
+  `entryPoints` for you.
+
+```{warning}
+If more than one application can relate to the same Traefik instance over `traefik-route`,
+ensure your router and service names are unique across relations. Traefik merges all submitted
+configurations together, so a name collision will cause one relation's config to silently
+overwrite another's.
+```
```
swetha1654: remove this, if this is already covered in the reference document

lib/charms/traefik_k8s/v0/traefik_route.py
```diff
@@ -94,15 +94,20 @@ def __init__(self, *args):
...
                     }
                 }
             }
         )
 ```
+
+For further details on how the library works (TLS twin routers, certificate resolvers, unique
+name requirements), see the
+[`traefik-route` section of the ingress library reference](https://charmhub.io/traefik-k8s/docs/ingress-libraries)
+on the Traefik charm documentation.
```
swetha1654:
```suggestion
For further details on how the library works, see the `traefik-route` section of the ingress library reference
on the Traefik RTD documentation.
```

docs/how-to/integrate.md
swetha1654:
```suggestion
```

docs/explanation/tls.md
```diff
@@ -0,0 +1,77 @@
...
+  provider used by the application, or from the `certificates` relation if the same authority
+  issues Traefik's own certificate.
+
+## Putting it together
+
+If both segments have TLS enabled, a request flows like this:
+
+1. A client calls `https://traefik.gateway/some-route`.
+2. Traefik replies with its own certificate (issued via the `certificates` relation, or via the
+   `tls-*` config options).
```
erinecon:
```suggestion
   `tls-*` configuration options).
```
nit

docs/how-to/integrate.md
```diff
@@ -54,6 +62,34 @@ Once you have added the `ingress` library the charm would need to be re-packed w
 `traefik-k8s` and notifying the charm whenever Traefik replies with an ingress URL or
 that URL is revoked for some reason (e.g. the cloud admin removed the relation).
 
+### The `strip_prefix` parameter
+
+For details on the `strip_prefix` parameter and when to use it, see
+{ref}`reference_ingress_libraries`.
```
erinecon, line 71: Nit: Since `strip_prefix` doesn't appear in any of the code blocks on this page, it would be nice to have a sentence that describes the purpose of the parameter and where a user would expect to find it. The sentence would need to be authoritative and instructional, something like:
```suggestion
### The `strip_prefix` parameter

Add the `strip_prefix` constructor argument to the `IngressPerAppRequirer` /  `IngressPerUnitRequirer`
object to strip the model path prefix from the request Traefik forwards on to your workload.

For details on the `strip_prefix` parameter and when to use it, see
{ref}`reference_ingress_libraries`.
```
Please update or edit as you see fit -- my suggestion comes from my understanding of
the parameter as written in the docs, and I didn't mention `IngressPerUnitRequirer` at all.

docs/how-to/integrate.md
```diff
@@ -54,6 +62,34 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+```python
+    def _on_config_changed(self, _):
+        self.ingress.provide_ingress_requirements(host="foo.com", port=42)
+```
+
+Each call to `provide_ingress_requirements` overwrites the previous request in the relation
+databag; there is only ever a single, current `(host, port)` pair per unit/application.
+
+```{note}
+Use one of the two APIs: constructor arguments, or `provide_ingress_requirements`.
```
erinecon:
```suggestion
Use only one of the two APIs: either constructor arguments, or `provide_ingress_requirements`.
```
nit

docs/how-to/integrate.md
```diff
@@ -54,6 +62,34 @@ Once you have added the `ingress` library the charm would need to be re-packed w
...
+Using both can lead to inconsistency: if some hooks call `provide_ingress_requirements` and
+others do not, hooks that skip the call will leave the previous values in the relation databag,
+which may not reflect the charm's current state. Most charms only need one or the other.
```
erinecon:
```suggestion
which may not reflect the charm's current state.
```
Removing the final sentence because it introduces uncertainty and confusion. We tell users "Use only one of the two APIs" and then imply that there's a use case where a charm would need both.

docs/how-to/integrate.md
```diff
@@ -68,3 +104,13 @@ juju run traefik/0 show-proxied-endpoints
...
 These are the URLs at which your workloads are externally accessible.
 
+(how_to_integrate_traefik_route)=
+
+## Use `traefik-route` for raw Traefik configuration
+
+Over the `traefik-route` relation, the requiring charm submits raw Traefik dynamic (and
+optionally static) configuration, which Traefik merges into its own configuration. The
+[`traefik_route` library page on Charmhub](https://charmhub.io/traefik-k8s/libraries/traefik_route)
+contains full usage examples, and {ref}`reference_ingress_libraries` covers the key behaviours
```
erinecon:
```suggestion
contains full usage examples, and {ref}`reference_ingress_libraries` covers the key behaviors
```
nit, consistency with US English in the rest of the docs

docs/how-to/troubleshoot-gateway-address-unavailable.md
```diff
@@ -23,11 +23,34 @@ tailor the exact steps and commands to your setup.
...
+into `error` status rather than a clean blocked/waiting status.
+
+Check with:
+
+```bash
+juju status
+```
+
+If the Traefik unit shows `error` and `juju debug-log` shows a `lightkube.core.exceptions.ApiError`
+mentioning `services is forbidden` or `cannot [...] resource "services"`, this is the cause.
```
erinecon:
```suggestion
Confirm that Traefik was deployed without `trust` by checking for an `error` message in
the Traefik unit and checking if `juju debug-log` shows a `lightkube.core.exceptions.ApiError`
mentioning `services is forbidden` or `cannot [...] resource "services"`.
```
nit, an attempt to make the text more authoritative and instructional

docs/reference/index.md
```diff
@@ -13,6 +13,6 @@ The pages in this section contain technical information for topics related to th
 ```{toctree}
 charm-architecture
```
erinecon, line 15:
```suggestion
:maxdepth: 1
charm-architecture
```
a bit out of scope, but while we're here, I'd love to clean up the rendered version of this page

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
+---
+myst:
+  html_meta:
+    "description lang=en": "Reference material for the ingress charm libraries provided by Traefik."
```
erinecon:
```suggestion
    "description lang=en": "Reference material for the ingress libraries provided by the Traefik charm."
```
nit

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+(reference_ingress_libraries)=
+
+# Ingress library reference
+
+This page covers selected parameters and behaviour of the ingress charm libraries provided by
```
erinecon:
```suggestion
This page covers selected parameters and behavior of the ingress charm libraries provided by
```
nit, US English

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+model/app-unique path prefix for your application (e.g. `/model_name-app_name`), and the
+ingress URL handed back to clients always includes that prefix — `strip_prefix` does **not**
+change the externally-visible URL.
+
+What `strip_prefix=True` actually does is add a Traefik `stripPrefix` middleware that removes
+the prefix only from the request Traefik forwards on to your workload, so your application sees
+the request as if it had been made against `/` instead of `/model_name-app_name/...`.
+
+`strip_prefix` only has an effect in `path` routing mode. In `subdomain` routing mode there is
+no path prefix to strip, so the option is a no-op.
```
erinecon, line 31: Question: This parameter is only relevant for the `ingress` and `ingress_per_unit` libraries, correct? Might be good to clarify if that's true

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+- [`ingress_per_unit`](https://charmhub.io/traefik-k8s/libraries/ingress_per_unit)
+- [`traefik_route`](https://charmhub.io/traefik-k8s/libraries/traefik_route)
+
+## The `strip_prefix` parameter
+
+`IngressPerAppRequirer` (and `IngressPerUnitRequirer`) accept a `strip_prefix: bool = False`
+constructor argument. In `path` routing mode (the default), Traefik always generates a
+model/app-unique path prefix for your application (e.g. `/model_name-app_name`), and the
+ingress URL handed back to clients always includes that prefix — `strip_prefix` does **not**
+change the externally-visible URL.
```
erinecon, line 24:
```suggestion
`IngressPerAppRequirer` (and `IngressPerUnitRequirer`) accept a `strip_prefix: bool = False`
constructor argument. In `path` routing mode (the default), Traefik always generates a
unique model/app path prefix for your application (e.g. `/model_name-app_name`), and the
ingress URL handed back to clients always includes that prefix. Setting `strip_prefix` does **not**
change the externally-visible URL.
```
nits

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+
+What `strip_prefix=True` actually does is add a Traefik `stripPrefix` middleware that removes
+the prefix only from the request Traefik forwards on to your workload, so your application sees
+the request as if it had been made against `/` instead of `/model_name-app_name/...`.
```
erinecon: Why is there a `...` in `/model_name-app_name/...`  , but not for `/`? Basically I'm wondering whether this line should say:
```suggestion
the request as if it had been made against `/...` instead of `/model_name-app_name/...`.
```
or

```suggestion
the request as if it had been made against `/` instead of `/model_name-app_name/`.
```

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+
+`strip_prefix` only has an effect in `path` routing mode. In `subdomain` routing mode there is
+no path prefix to strip, so the option is a no-op.
+
+## The `traefik-route` interface: things to keep in mind
+
+### Automatic TLS twin router (default `raw=False` mode)
```
erinecon: The [style guide discourages adding two headers with no text between them](https://documentation.ubuntu.com/style-guide/#other-considerations), and I'm not a huge fan of the `: things to keep in mind` part of the H2 header.  I'm wondering whether we can move that intention into the text itself, not really sure what the text would be but this is my vision:
```suggestion
## The `traefik-route` interface

The `traefik-route` interface allows users to provide Traefik configuration directly. However, when `raw = False` is set, 
the Traefik charm does some preprocessing of the Traefik configuration provided by the user before applying the configuration. 
Here are some things to keep in mind if you are using the `traefik-route` library:

### Automatic TLS twin router (default `raw=False` mode)
```

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+By default, Traefik automatically generates a TLS-enabled twin of every HTTP router you declare
+(named `<router>-tls`) and attaches it to the `websecure` entrypoint. Do not declare your own
+`-tls`-suffixed routers or `tls` blocks in your submitted config unless you opt out of this
+behaviour by setting `raw=True`. Attach your own HTTP router to the `web` entrypoint; Traefik
+will not rewrite your router's `entryPoints` for you.
+
+### Certificate resolvers
+
+This charm does not request certificates from an ACME server. It only uses certificates that
+are supplied to it via the `certificates` relation or the `tls-*` config options. Do not set
```
erinecon:
```suggestion
are supplied to it via the `certificates` relation or the `tls-*` configuration options. Do not set
```
nit

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+`certResolver` in your router's `tls` block; that option is only valid when Traefik itself is
+obtaining certificates via ACME.
+
+### Unique router and service names
+
+If more than one application can relate to the same Traefik instance over `traefik-route`,
```
erinecon:
```suggestion
If more than one application can integrate to the same Traefik instance over `traefik-route`,
```

docs/reference/ingress-libraries.md
```diff
@@ -0,0 +1,55 @@
...
+ensure your router and service names are unique across relations. Traefik merges all submitted
+configurations together, so a name collision will cause one relation's config to silently
```
erinecon:
```suggestion
configurations together, so a name collision will cause one relation's configuration to silently
```
nit

erinecon (approved):
Thank you for refactoring the how-to guide -- my concerns have been addressed :)

I spotted some nits and had some additional questions, but nothing that blocks this PR

docs/how-to/integrate.md
swetha1654: i updated your suggestion and adding it

docs/reference/ingress-libraries.md
swetha1654: yes, we start this section by mentioning this parameter is available in `IngressPerAppRequirer` and `IngressPerUnitRequirer`

docs/reference/ingress-libraries.md
swetha1654: updated your suggestion with an intro

PR #773, opened 2026-09-04
feat: Add custom CSR attributes

### LINKED ISSUE #667: Config option to specify subjects for CSR
### Enhancement Proposal

Currently the CSR attributes like CN, sans dns, country name etc are autogenerated or hardcoded in the charm. Provide config options to expose optional attributes of CSR (Country, Organization, State, Location and Organization Unit) so that user can set those configurations if required. (Similar to how vault-k8s exposes config option https://github.com/canonical/vault-k8s-operator/blob/main/k8s/charmcraft.yaml#L264)

If the config options are provided, sunbeam project can use the existing commands `tls ca` which allows managing certificates via manual-tls-certificates operator.

Workaround is to set the TLS certificates via config options tls-ca, tls-cert, tls-key and distribute CA certs to all necessary applications external to sunbeam.

Related bug: https://bugs.launchpad.net/sunbeam-charms/+bug/2148394

### Impact

High (The feature has short-term technical value)

### Impact Rationale

The workaround bypasses completely sunbeam tls ca feature and which requires running some actions of certificate-distributor component (keystone-k8s) in sunbeam outside sunbeam commands.

swetha1654:
#### What this PR does

Adds custom CSR subject attributes 

Closes #667 

#### Checklist

### DIFF
--- a/docs/release-notes/artifacts/pr0773.yaml
+++ b/docs/release-notes/artifacts/pr0773.yaml
@@ -8,7 +8,8 @@ changes:
       The charm now accepts `custom-csr-subject-attributes` to populate
       additional CSR subject fields in certificate requests.
     urls:
-      pr: []
+      pr:
+        - https://github.com/canonical/traefik-k8s-operator/pull/773
       related_issue:
         - https://github.com/canonical/traefik-k8s-operator/issues/667
     visibility: public
--- a/docs/release-notes/artifacts/pr0773.yaml
+++ b/docs/release-notes/artifacts/pr0773.yaml
@@ -1,0 +1,15 @@
+# Version of the artifact schema
+version_schema: 2
+changes:
+  - title: Added configurable CSR subject attributes
+    author: swetha1654
+    type: minor
+    description: |
+      The charm now accepts `custom-csr-subject-attributes` to populate
+      additional CSR subject fields in certificate requests.
+    urls:
+      pr: []
+      related_issue:
+        - https://github.com/canonical/traefik-k8s-operator/issues/667
+    visibility: public
+    highlight: false
--- a/docs/release-notes/artifacts/pr0773.yaml
+++ b/docs/release-notes/artifacts/pr0773.yaml
@@ -5,7 +5,7 @@ changes:
     author: swetha1654
     type: minor
     description: |
-      The charm now accepts `custom-csr-subject-attributes` to populate
+      The charm now accepts `csr-subject-atttributes` to populate
       additional CSR subject fields in certificate requests.
     urls:
       pr:
--- a/src/charm.py
+++ b/src/charm.py
@@ -2159,13 +2159,15 @@ def parse_custom_csr_subject_attributes(attributes: Optional[str]) -> Optional[D
 
     parsed_attributes: Dict[str, str] = {}
     supported_keys = ", ".join(sorted(CUSTOM_CSR_SUBJECT_ATTRIBUTE_ALIASES))
+    error = False
 
     for pair in normalized_attributes.split(","):
         if not pair.strip():
             logger.error(
                 "Invalid format for 'custom-csr-subject-attributes'. Empty attribute found."
             )
-            return None
+            error = True
+            continue
 
         key_value = pair.split("=")
         if len(key_value) != 2:
@@ -2173,7 +2175,8 @@ def parse_custom_csr_subject_attributes(attributes: Optional[str]) -> Optional[D
                 "Invalid format for 'custom-csr-subject-attributes'. "
                 "Expected format: key1=value1,key2=value2."
             )
-            return None
+            error = True
+            continue
 
         key = key_value[0].strip()
         value = key_value[1].strip()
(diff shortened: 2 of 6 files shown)


src/charm.py
```diff
@@ -526,6 +535,13 @@ def _get_cert_requests(self) -> List[CertificateRequestAttributes]:
             for endpoint in self._get_proxied_endpoints(use_gateway_address=True).values()
             if "url" in endpoint and urlparse(endpoint["url"]).scheme
         }
+        custom_subject_attributes = self._custom_csr_subject_attributes
+        if custom_subject_attributes is None:
+            logger.warning(
+                "CSR subject config is invalid; proceeding with default CSR attributes."
+            )
```
f-atwi, line 542: could this be misleading since it could be none if they user did not want to configure it?

src/charm.py
```diff
@@ -566,11 +582,21 @@ def _get_cert_requests(self) -> List[CertificateRequestAttributes]:
                     "Skipping certificate request for address %s - no valid common name", addr
                 )
                 continue
+            configured_common_name = custom_subject_attributes.get("common_name")
+            common_name = configured_common_name or common_name
```
f-atwi: nit 
```suggestion
            common_name = custom_subject_attributes.get("common_name", common_name)
```

src/charm.py
```diff
@@ -827,29 +861,38 @@ def _is_workload_tracing_ready(self) -> bool:
...
-        old_valid = sorted(self.csrs, key=lambda c: c.common_name)
+        new_keys = {self._csr_key(csr) for csr in new_csrs}
+        old_keys = {self._csr_key(csr) for csr in self.csrs}
 
-        # Compare by common_name + sans sets
-        def _csr_key(c: CertificateRequestAttributes) -> tuple:
-            return (c.common_name, frozenset(c.sans_dns or ()), frozenset(c.sans_ip or ()))
+        if new_keys == old_keys:
+            logger.debug("Certificate requests unchanged (%d); skipping cert refresh.", len(old_keys))
+            return
```
f-atwi, line 870: why not do `if new_csrs == self.csrs` directly?
`CertificateRequestAttributes` already defines `__eq__`

src/charm.py
```diff
@@ -827,29 +861,38 @@ def _is_workload_tracing_ready(self) -> bool:
...
 
-        new_keys = {_csr_key(c) for c in new_valid}
-        old_keys = {_csr_key(c) for c in old_valid}
+        logger.info(
+            "Certificate requests changed (old=%d, new=%d). Refreshing certs.",
+            len(old_keys),
+            len(new_keys),
+        )
```
f-atwi, line 876: not sure if only using the lengths here is enough
but this is just a logging message so idk

src/charm.py
```diff
@@ -1377,6 +1420,12 @@ def _process_status_and_configurations(self) -> None:
             )
             return
 
+        if self._custom_csr_subject_attributes is None:
+            self.unit.status = BlockedStatus(
+                'invalid "custom-csr-subject-attributes" value; see logs.'
+            )
+            return
```
f-atwi, line 1426: in a previous comment i mentioned the case where `custom-csr-subject-attributes`
apparently it is not supposed to be none
however I am confused as to why we cant not have this config

docs/release-notes/artifacts/pr0773.yaml
```diff
@@ -0,0 +1,16 @@
+# Version of the artifact schema
+version_schema: 2
+changes:
+  - title: Added configurable CSR subject attributes
+    author: swetha1654
+    type: minor
+    description: |
+      The charm now accepts `custom-csr-subject-attributes` to populate
```
seb4stien: I'm not sure if we should name it "custom", isn't a config option a customization by definition. WDYT of just "csr-subject-atttributes"?

swetha1654: will change it to "csr-subject-atttributes"

src/charm.py
swetha1654: no if the user did not configure it then custom_subject_attributes is an empty dictionary.

src/charm.py
swetha1654: `new_csrs == self.csrs` would be a list comparision, which is order sensitive. And the `self.csrs` attribute is built from a set of `addrs`, which is not order sensitive so i don't expect it to be the same all the time.

With that being said, i would like to also mention that i don't usually expect more than 1 CSR in the list because all the relations will use the same hostname, but I'd rather not make assumptions and keep the code as is.

I went through a huge rabbit hole with this function; i think there's a lot of scope for improvement here, but in the end I'd like my changes to be as minimal as possible and not break any existing code.

src/charm.py
swetha1654: you're right, let me change it to print the old and new keys.

src/charm.py
swetha1654: the code will take care of making this an empty dict if the user hasn't provided anything. The function only returns None when there are issues parsing the config.

f-atwi (approved):

